PR #53, opened 2024-06-21
Docker image

wuputah:
* Usage follows https://hub.docker.com/_/postgres/
* Images are pushed to [`pgduckdb/pgduckdb`](https://hub.docker.com/r/pgduckdb/pgduckdb)
* Anytime an image is built, it's pushed to `17-$sha`, `16-$sha`, `15-$sha`
* Nightly images pushed to `17-main`, `16-main`, `15-main`
* Version tags (tags beginning with `v`) will be built automatically and pushed to `17-$tag`, `16-$tag`, `15-$tag`
* Only built on PR when the Docker-related files are changed

To test, [grab a tag](https://hub.docker.com/r/pgduckdb/pgduckdb/tags), for instance:

```
TAG=17-b203b63b052df8a7fece37f79face6b7522a498f
PORT=6543
```

Then:

```
docker run --name pgduckdb -e POSTGRES_PASSWORD=secret -p $PORT:5432 \
  pgduckdb/pgduckdb:$TAG -c shared_preload_libraries=pg_duckdb

psql postgres://postgres:secret@127.0.0.1:$PORT/postgres
```

Cleanup:

```
docker stop pgduckdb
docker rm pgduckdb
```

This can be made easier with a `docker-compose.yml`, which will be forthcoming.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+*.o
+*.so
+*.dylib
+.depend
+Dockerfile
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,8 @@
+.git
 *.o
 *.so
 *.dylib
-.depend
+*.bc
+.deps
 Dockerfile
+third_party/duckdb/build
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -3,12 +3,14 @@ name: Build Docker
 on:
   push:
     tags: ["v*"]
-    branches: ["main"]
   pull_request:
     paths:
       - Dockerfile
       - docker-bake.hcl
       - ".github/workflows/docker.yaml"
+  schedule:
+    - cron: "44 4 * * *"
+  workflow_dispatch:
 
 jobs:
   docker:
@@ -20,12 +22,6 @@ jobs:
         platform: ["linux/amd64", "linux/arm64"]
 
     steps:
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
@@ -37,7 +33,11 @@ jobs:
           submodules: "recursive"
       - name: Set env
         run: |
-          echo "POSTGRES_IMAGE=pgduckdb/pgduckdb:${{matrix.postgres}}-${{github.sha}}" >> $GITHUB_ENV
+          # find the first tag for this commit starting with 'v'
+          tag=$(git tag --points-at HEAD | grep ^v | head -n 1)
+          # if no tag is found, use the sha
+          if [ "$tag" = "" ]; then tag="${{ github.sha }}"; fi
+          echo "POSTGRES_TAGS=pgduckdb/pgduckdb:${{ matrix.postgres }}-$tag" >> $GITHUB_ENV
           echo "POSTGRES_VERSION=${{ matrix.postgres }}" >> $GITHUB_ENV
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
@@ -54,4 +54,4 @@ jobs:
             *.platform=${{ matrix.platform }}
             *.cache-to=type=gha,mode=max
             *.cache-from=type=gha
-            postgres.tags=${{ env.POSTGRES_IMAGE }}
+            postgres.tags=${{ env.POSTGRES_TAGS }}
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -49,7 +49,7 @@ jobs:
         uses: docker/bake-action@v5
         with:
           targets: pg_duckdb_${{ matrix.postgres }}
-          push: true
+          push: ${{ github.event_name != 'pull_request' }}
           set: |
             *.platform=${{ matrix.platform }}
             *.cache-to=type=gha,mode=max
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,48 @@
+name: Build Docker
+
+on:
+  push:
+     tags: ["v*"]
+     branches: ['main']
+  pull_request:
+    paths:
+      - Dockerfile
+      - docker-bake.hcl
+      - '.github/workflows/docker.yaml'
+
+jobs:
+  docker:
+    name: Build Docker
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        postgres: ["16", "17"]
+
+    steps:
+      -
+        name: setup POSTGRES_IMAGE env
+        run: echo "POSTGRES_IMAGE=${{ format('wuputah/pg_duckdb:{0}-{1}', matrix.postgres, github.sha) }}" >> $GITHUB_ENV
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Docker buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
+      # -
+      #   name: Login to DockerHub
+      #   uses: docker/login-action@v3
+      #   with:
+      #     username: ${{ secrets.DOCKERHUB_USERNAME }}
+      #     password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: docker bake
+        uses: docker/bake-action@v5
+        with:
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -45,7 +45,8 @@ jobs:
         uses: docker/bake-action@v5
         with:
           targets: pg_duckdb_${{ matrix.postgres }}
-          push: ${{ github.event_name != 'pull_request' }}
+          # push: ${{ github.event_name != 'pull_request' }}
+          push: true
           set: |
             *.platform=${{ matrix.platform }}
             *.cache-to=type=gha,mode=max
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -2,38 +2,36 @@ name: Build Docker
 
 on:
   push:
-     tags: ["v*"]
-     branches: ['main']
+    tags: ["v*"]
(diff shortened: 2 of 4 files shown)


mike-luabase:
this is what ultimately worked for me:

```
FROM postgres:16-bookworm as base

###
### BUILDER
###
FROM base as builder

RUN --mount=type=cache,target=/var/cache/apt \
  apt-get update -qq && \
  apt-get install -y build-essential libreadline-dev zlib1g-dev flex bison libxml2-dev libxslt-dev \
  libssl-dev libxml2-utils xsltproc pkg-config libc++-dev libc++abi-dev libglib2.0-dev libtinfo5 cmake \
  libstdc++-12-dev postgresql-server-dev-16 liblz4-dev ccache git

WORKDIR /build

ENV PATH=/usr/lib/ccache:$PATH
ENV CCACHE_DIR=/ccache

# Clone the pg_duckdb repository and initialize submodules
RUN git clone --branch main https://github.com/duckdb/pg_duckdb.git . && \
    git submodule update --init --recursive

# permissions so we can run as `postgres` (uid=999,gid=999)
RUN chown -R postgres:postgres .
RUN chown -R postgres:postgres /usr/lib/postgresql /usr/share/postgresql
RUN mkdir /out && chown postgres:postgres /out
RUN rm -f .depend

USER postgres

# Build and install
RUN --mount=type=cache,target=/ccache/,uid=999,gid=999 make install
RUN --mount=type=cache,target=/ccache/,uid=999,gid=999 DESTDIR=/out make install

###
### CHECKER
###
FROM builder as checker

USER postgres
RUN --mount=type=cache,target=/ccache/,uid=999,gid=999 make installcheck

###
### OUTPUT
###
# This creates a usable postgres image but without the packages needed to build
FROM base as output
COPY --from=builder /out /
```

jorinvo:
I would love to install pg_duckdb in our Postgres Docker image. I tried the Dockerfile from above, but it was stuck at 100% at `make install` for 30 minutes.

mike-luabase:
> I would love to install pg_duckdb in our Postgres Docker image. I tried the Dockerfile from above, but it was stuck at 100% at `make install` for 30 minutes.

Took a very long time to run for me too. Might have been more than 30 minutes before it was complete.

jorinvo:
Thanks @mike-luabase, that's good to know. 
I am afraid that's not usable for us for now. But I am looking forward to having a prebuilt image available some day. pg_duckdb is definitely an exciting project :star_struck:

JelteF:
To make the build faster it would help a lot if you changed the `make install` commands to be parallel, by using e.g. `make -j10 install`. Or maybe `make -j$(nproc) install`

wuputah:
> To make the build faster it would help a lot if you changed the `make install` commands to be parallel, by using e.g. `make -j10 install`. Or maybe `make -j$(nproc) install`

Even after #211, there remains an issue with `-j` being passed to the duckdb build that I haven't managed to solve. You'll see this printed in the log when running `make duckdb`:

```
make[2]: warning: jobserver unavailable: using -j1.  Add `+' to parent make rule.
```

This is despite the fact we are using `$(MAKE)` [as suggested](https://stackoverflow.com/questions/60702726/warning-jobserver-unavailable-using-j1-add-to-parent-make-rule) for this issue. This probably has [something to do with DuckDB's use of cmake](https://stackoverflow.com/a/33171336), but I don't know anything about cmake.

.dockerignore
```diff
@@ -0,0 +1,8 @@
+.git
+*.o
+*.so
+*.dylib
+*.bc
+.depend
```
JelteF, line 6:
```suggestion
.deps
```

wuputah:
In order to test this end-to-end, I've continued work at [hydradatabase/pg_duckdb-docker](https://github.com/hydradatabase/pg_duckdb-docker), complete with pushes to [ghcr.io/hydradabase/pg_duckdb](https://github.com/hydradatabase/pg_duckdb-docker/pkgs/container/pg_duckdb), which required a few secrets and such (notably, an extra login to Dockerhub to get around rate limits).

This isn't an endorsement of this opinion, but a separate repo might be a preferable approach overall just to keep any iteration of the docker builds, caches, and overall user experience separated rather than maintaining that here. The downside would be a lack of visibility if the Docker image stops building for some reason, so this would have to monitored independently. 

So unless there's an objection, we can close this PR. I don't have a strong opinion either way, and we can certainly change our minds later.

Y--:
> In order to test this end-to-end, I've continued work at [hydradatabase/pg_duckdb-docker](https://github.com/hydradatabase/pg_duckdb-docker), complete with pushes to [ghcr.io/hydradabase/pg_duckdb](https://github.com/hydradatabase/pg_duckdb-docker/pkgs/container/pg_duckdb), which required a few secrets and such (notably, an extra login to Dockerhub to get around rate limits).
> 
> This isn't an endorsement of this opinion, but a separate repo might be a preferable approach overall just to keep any iteration of the docker builds, caches, and overall user experience separated rather than maintaining that here. The downside would be a lack of visibility if the Docker image stops building for some reason, so this would have to monitored independently.
> 
> So unless there's an objection, we can close this PR. I don't have a strong opinion either way, and we can certainly change our minds later.

Ah I just saw this comment - I've created a Docker Hub account for PG and added a login step. I don't have strong opinion on single vs separated repo though :-)

wuputah:
Ah, you're pushing to dockerhub, which does get around any github/ghcr permission issues (in fact we no longer need that ghcr login step). I couldn't set secrets here so wanted a place to test without jumping through those hopes.

The separate repo would let me continue iterating on a README with Docker instructions, docker-compose, etc without doing PRs here, but either way is fine. The only thing I can't do in a separate repo is trigger off new commits or tags. 

As for prior art, duckdb itself does not do Docker, and it seems popular docker image projects are [davidgasquez/awesome-duckdb](https://github.com/davidgasquez/awesome-duckdb) and [duckerlabs/ducker](https://github.com/duckerlabs/ducker).

Y--:
++ agreed. 

I think to be able to ship this PR we need to:
* implement build cache
* parallelize build
* have the build take O(minute) (not 3h like it is right now)

wuputah:
I'm not aware a way to do any better on speed, at least with Github runners - the main issue is the arm64 emulation. After that, it's the cache layer, which doesn't work particularly well with GHA. There's paid solutions that work a lot better (e.g. Depot). Basically, free = slow.

JelteF:
We can have the job run nightly on main, and allow manually triggering it. Instead of running it constantly on every PR. I think I'd prefer to keep Dockerfile in the same repo, especially since this is pretty much the main way we'll be shipping binaries to users. Having stuff together in the same repo makes that a bit easier to manage imho.

wuputah:
Nothing left on my to-do list here, though a docker-compose file would be nice to add, that can be done in a separate PR. As well as a separate doc + quick mention in the README.

JelteF:
> quick mention in the README.

I think a quick mention in the readme, contributing.md or PR description would be very helpful. Mainly just so I can copy paste commands that should work when trying this out for the review.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          targets: pg_duckdb_${{ matrix.postgres }}
+          # push: ${{ github.event_name != 'pull_request' }}
+          push: true
+          set: |
+            *.platform=${{ matrix.platform }}
+            *.cache-to=type=gha,mode=max
+            *.cache-from=type=gha
+            postgres.tags=pgduckdb/pgduckdb:${{ matrix.postgres }}-${{ github.sha }}
+            ${{ !contains(github.ref_name, '/') && format('postgres.tags=pgduckdb/pgduckdb:{0}-{1}', matrix.postgres, github.ref_name) || '' }}
```
Y--, line 55: Will this override the one above or add it? Also instead of skipping when branch name have a `/` could we replace with `-`? (seems arbitrary to just exclude these no?)

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          platforms: ${{ matrix.platform }}
+      - name: docker bake
+        uses: docker/bake-action@v5
+        with:
+          targets: pg_duckdb_${{ matrix.postgres }}
+          # push: ${{ github.event_name != 'pull_request' }}
```
Y--, line 48: Let's remove?

Y-- (commented):
Looks good. We will need a green build to merge though no?

Y-- (commented):
Looks good. We will need a green build to merge though no?

.github/workflows/docker.yaml
wuputah: I was going to put this back in, but, yeah, it's fine.

.github/workflows/docker.yaml
wuputah: It adds to the one above it, so we will push with the sha anytime it gets built, including PRs.

Mm yeah we could replace `/` with `-`, but this was sorta an accidental "feature" -- by doing it this way we only get tags and `main` pushed as "named tags" rather than PRs.

wuputah:
> Looks good. We will need a green build to merge though no?

Yeah, looks like the azure ubuntu mirror is broken atm.

.github/workflows/docker.yaml
wuputah: That won't work for tags though (`v0.1.0`)

wuputah:
> I think a quick mention in the readme, contributing.md or PR description would be very helpful.

PR description updated.

I'd like to get this merged here then continue on in separate PRs with docker-compose, docs changes, etc. Especially since there won't be an image tagged with `main` until it's merged. 😁

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+  rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build
+
+ENV PATH=/usr/lib/ccache:$PATH
+ENV CCACHE_DIR=/ccache
+
+# A more selective copy might be nice, but the git submodules are not cooperative.
+# Instead, use .dockerignore to not copy files here.
+COPY . .
```
JelteF, line 24: I think it would be good to automatically run `CREATE EXTENSION pg_duckdb` and configure `shared_preload_libraries=pg_duckdb`. That'll make it easier to run.

Using an initialization script should make automating that possible I think: https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts

JelteF (approved):
Overall this looks good. I left one comment, but that can also be done later. Feel free to merge as you see fit.

Dockerfile
wuputah: yeah we can handle that at the docker compose level

PR #57, opened 2024-06-24
SHLIB_LINK: prefer new libs to old libs

wuputah:
This fixes a linker error when an older version of libduckdb is already installed in PG_LIB. We should prefer the version we just finished building instead.


JelteF:
Ah yeah, I definitely ran into this a few times I think. I worked around it then by manually removing the libduckdb.so file from the installed location.

JelteF (approved):

PR #63, opened 2024-07-02
Parallel Scan with multiple processes

mkaruza:
* We can now start multiple worker process that can read relation blocks and write buffer to shared memory to be consumed by duckdb reader threads. Problem can be viewed as variant of multi producer / multi consumer where producer are responsible for releasing buffer after they are read.


JelteF:
@mkaruza what's the deal with this? I guess it's not critical for 0.1.0. Given the amount of merge conflicts it has now, I'm think it probably makes the most sense to simply close this and maybe create an issue for it instead.

mkaruza:
@JelteF agree not critical, but could be approach to get better performances of heap table scans (specifically to fetch table pages in parallel; currently there is bottleneck there as only one thread in process can request page fetching).

src/scan/heap_reader_worker.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+			if (!thread_running) {
+				break;
+			}
+
+			Buffer buffer = ReadBufferExtended(relation, MAIN_FORKNUM, i, RBM_NORMAL, strategy);
+			LockBuffer(buffer, BUFFER_LOCK_SHARE);
+
+			/* is previous buffer done */
+			while (!pg_atomic_unlocked_test_flag(&thread_worker_shared_state->buffer_ready)) {
```
Y--, line 70: Did you try adding a very small "sleep" in the loop rather than busy-spinning? I wonder if it would help performances.
Also could we use signals instead?
Should we have a timeout to handle situation where thread dies without ever resetting its `thread_running` or `buffer_ready` flags?

src/scan/heap_reader_worker.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+			if (BufferIsValid(previous_buffer)) {
+				UnlockReleaseBuffer(previous_buffer);
+			}
+		}
+	}
+
+	if (thread_running) {
+		/* We are out of blocks fo reading so wait for last buffer to be done */
+		while (!pg_atomic_unlocked_test_flag(&thread_worker_shared_state->buffer_ready)) {
```
Y--, line 95: I guess same apply here - and we could factorize the two loops in a "wait" function?

src/scan/heap_reader.cpp
```diff
@@ -11,55 +11,167 @@ extern "C" {
...
+		*buffer = InvalidBuffer;
+		return;
+	}
+
+	if (BufferIsValid(*buffer)) {
+		pg_atomic_clear_flag(&m_thread_worker_shared_state->buffer_ready);
+	}
+
+	/* Is buffer ready for reading */
+	while (pg_atomic_unlocked_test_flag(&m_thread_worker_shared_state->buffer_ready)) {
```
Y--, line 123: I wonder if we could use signals here rather than busy-spinning?

Y-- (commented):
High level looks reasonable (and perf improvements are promising).
Wonder if we could implement communication between thread and worker through signals rather than busy-spin on atomic variables, and if this would improve perfs.

test/regression/regression.conf
```diff
@@ -1,6 +1,6 @@
 # Configuration
 
-shared_preload_libraries = 'pg_duckdb'
+shared_preload_libraries = 'pg_duckdb.so'
```
JelteF, line 3: This shouldn't be necessary, and actually probably breaks OSX

JelteF:
@mkaruza do you want to close this one in favor of #477

mkaruza:
Closing

PR #88, opened 2024-08-02
Local object cache for cached_httpfs extension

mkaruza:
* To cache remote object, user need to explicilty use `duckdb.cache(path, type)` function. Path is remote HTTPFS/S3/GCS/R2 object path and type is either `parquet` or `csv` indicating remote object type. Caching will not be triggered on normal `SELECT` queries.

* When file is currently being cached and if we run in parallel another query targeting same remote object - second query will fail because it will not be able to take read lock.

* PG_DATADIR/duckdb_cache will be used as cache directory


third_party/cached_httpfs/include/http_file_cache.hpp
```diff
@@ -0,0 +1,99 @@
...
+		return file->capacity;
+	}
+	//! Return the size of the initialized file
+	idx_t GetSize() {
+		D_ASSERT(file->initialized);
+		return file->size;
+	}
+
+private:
+	unique_ptr<lock_guard<mutex>> lock;
```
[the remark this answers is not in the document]
mkaruza: Lock are not needed. Locking is done on directly on file with READ/WRITE lock.

README.md
```diff
@@ -58,6 +58,9 @@ CREATE EXTENSION pg_duckdb;
 - Install DuckDB extensions using `SELECT duckdb.install_extension('extension_name');`
 - Toggle DuckDB execution on/off with a setting:
 	- `SET duckdb.execution = true|false`
+- Cache remote object localy for faster execution using `SELECT duckdb.cache('path', 'type');` where
```
Tishj, line 61: localy -> locally

src/pgduckdb_options.cpp
```diff
@@ -145,6 +190,55 @@ DuckdbInstallExtension(Datum name) {
 	return true;
 }
 
+static bool
+CanCacheRemoteObject(std::string remote_object) {
```
Tishj, line 194: I'd be more comfortable with this:
```c++
static bool
CanCacheRemoteObject(std::string remote_object) {
    return (remote_object.rfind("https://", 0) == 0) ||
           (remote_object.rfind("http://", 0) == 0) ||
           (remote_object.rfind("s3://", 0) == 0) ||
           (remote_object.rfind("s3a://", 0) == 0) ||
           (remote_object.rfind("s3n://", 0) == 0) ||
           (remote_object.rfind("gcs://", 0) == 0) ||
           (remote_object.rfind("gs://", 0) == 0) ||
           (remote_object.rfind("r2://", 0) == 0);
}
```

Multiplying `std::string::npos` multiple times feels problematic

third_party/cached_httpfs/http_file_cache.cpp
```diff
@@ -0,0 +1,103 @@
...
+	} else if (cache_file) {
+		flags = FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE | FileLockType::WRITE_LOCK;
+		handle = fs.OpenFile(file_name, flags);
+	}
+
+	ReleaseDirectoryCacheLock();
+}
+
+CachedFile::~CachedFile() {
+	if (!initialized && handle) {
```
Tishj, line 25: The handle should probably be closed before we remove the file

Tishj: Also, I don't think this condition is correct, we use `initialized` to indicate that the file is ready for reading, not that it was created by this process.

Even when it was created by this process, that doesn't necessarily mean that the file can be removed as another process can still be using the file. I'm not entirely sure what the condition should be here, but perhaps we need a interprocess reference count for the file.

I guess if we try to remove the file while another process is still using it, do we hang until it's closed or is the file unlinked and scheduled for removal when the other process closes the file?

Tishj: We could be hitting open file descriptor limitations if we're caching any decent amount of files which is also something to keep in mind, but perhaps figuring that out is a future PR

third_party/cached_httpfs/httpfs.cpp
```diff
@@ -323,17 +334,17 @@ unique_ptr<ResponseWrapper> HTTPFileSystem::GetRequest(FileHandle &handle, strin
 				    hfh.state->total_bytes_received += data_length;
 			    }
 			    if (!hfh.cached_file_handle->GetCapacity()) {
-				    hfh.cached_file_handle->AllocateBuffer(data_length);
+				    hfh.cached_file_handle->Allocate(data_length);
```
Tishj, line 337: Are we making sure that these are initialized files earlier on?
Since CachedFile::Allocate will throw an InternalException if `initialized` is set, this should be verified elsewhere or be downgraded to a regular exception as InternalExceptions are unrecoverable

third_party/cached_httpfs/http_file_cache.cpp
```diff
@@ -0,0 +1,103 @@
...
+void CachedFileHandle::Allocate(idx_t size) {
+	if (file->initialized) {
+		throw InternalException("Cannot allocate cached file that was already initialized");
+	}
+	file->handle->Trim(0, size);
+	file->capacity = size;
+}
+
+void CachedFileHandle::GrowFile(idx_t new_capacity, idx_t bytes_to_copy) {
+	file->handle->Trim(bytes_to_copy, new_capacity);
```
Tishj, line 74: Does this not need the same `file->initialized` check?

third_party/cached_httpfs/httpfs.cpp
```diff
@@ -624,6 +643,18 @@ static optional_ptr<HTTPMetadataCache> TryGetMetadataCache(optional_ptr<FileOpen
 	return nullptr;
 }
 
+static optional_ptr<HTTPFileCache> TryGetFileCache(optional_ptr<FileOpener> opener, HTTPFileSystem &httpfs) {
+	auto db = FileOpener::TryGetDatabase(opener);
+	auto client_context = FileOpener::TryGetClientContext(opener);
+	if (!db) {
```
Tishj, line 649: What is this for? `db` isn't used after this condition so I'm a little confused

third_party/cached_httpfs/include/httpfs.hpp
```diff
@@ -194,6 +201,7 @@ class HTTPFileSystem : public FileSystem {
 	// Global cache
 	mutex global_cache_lock;
 	duckdb::unique_ptr<HTTPMetadataCache> global_metadata_cache;
+	duckdb::unique_ptr<HTTPFileCache> glob_file_cache;
```
Tishj, line 204: nit: `global_file_cache`

"glob" is a separate thing and could be confusing here

third_party/cached_httpfs/http_file_cache.cpp
mkaruza: When file doesn't exists, i.e when we cache file it will open file with WRITE_LOCK so all other processes cannot open file until that lock is released. 
This logic is intended to clean up file if user interrupts caching (initialize = false & file handle != null) but as we, for now, don't have ability to do this i'll remove this logic in destructor.
As discussed, improvements for this extension will be done later, but it is good that we keep track of what could be potential problems.

third_party/cached_httpfs/httpfs.cpp
mkaruza: Most of this code around file read/write/reserve is copied from in-memory cache which need to check conditions but in our case (cache file is not available until completely downloaded). Methods that are used in file caching are used only in GET request which is issued once we have WRITE_LOCK and opened file handle. Changed all this check to be ASSERT. Comment applies to next review change request.

src/pgduckdb_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+	return true;
+}
+
+std::string
+CreateOrGetDirectoryPath(std::string directory_name) {
+	StringInfo duckdb_data_directory = makeStringInfo();
+	appendStringInfo(duckdb_data_directory, "%s/%s", DataDir, directory_name.c_str());
+
+	if (!CheckDirectory(duckdb_data_directory->data)) {
+		if (mkdir(duckdb_data_directory->data, S_IRWXU | S_IRWXG | S_IRWXO) == -1) {
```
JelteF, line 48: Should use `MakePGDirectory` instead
```suggestion
		if (MakePGDirectory(duckdb_data_directory->data) == -1) {
```

src/pgduckdb_utils.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+	std::string directory(duckdb_data_directory->data);
+	pfree(duckdb_data_directory->data);
+	return directory;
+}
+
+} // namespace pgduckdb
+
+void
+DuckdbCreateCacheDirectory() {
```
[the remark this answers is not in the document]
wuputah: I think PG_DATADIR is the only thing we can be assured that Postgres can write to, aside from `mktemp`. But the intent was not to have that directory persist in backups, so that's a good point. A configurable option would be good.

wuputah (approved):

wuputah (approved):

wuputah (approved):

PR #92, opened 2024-08-06
Move httpfs extension outside of duckdb library.

mkaruza:
* Skip clang format for httpfs folder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,16 +77,13 @@ third_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src/$(DUCKDB_LIB):
 	DISABLE_SANITIZER=1 \
 	ENABLE_UBSAN=0 \
 	BUILD_UNITTESTS=OFF \
-	BUILD_HTTPFS=1 \
-	BUILD_JSON=1 \
-	CMAKE_EXPORT_COMPILE_COMMANDS=1 \
-	-j8
+	EXTENSION_CONFIGS="../pg_duckdb_extensions.cmake"
 
 install-duckdb:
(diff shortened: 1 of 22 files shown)


JelteF (dismissed):
I still don't understand why we need to vendor in a fork of the httpfs library. I'd much rather see that we add caching support into the upstream httpfs repo. If we don't want to wait for the next duckdb or httpfs release, we could still use this new upstream extension using a submodule. But copying all files over to this repo seems a bad idea in the long run to me.

wuputah (approved):
Appreciate the opinions but we are moving forward with this as this is a key feature for data lake use cases with Postgres+DuckDB. We can revisit moving this or merging the feature upstream in the future.

dentiny:
> I still don't understand why we need to vendor in a fork of the httpfs library. I'd much rather see that we add caching support into the upstream httpfs repo. If we don't want to wait for the next duckdb or httpfs release, we could still use this new upstream extension using a submodule. But copying all files over to this repo seems a bad idea in the long run to me.

Hi team, when I use duckdb httpfs, I suffer the same problem as you guys, so I made my extension.
Link here: https://github.com/dentiny/duck-read-cache-fs

A few key features:
- Supports both on-disk cache and in-memory cache, with block size tunable
- Supports parallel IO request, with request size tunable
- Supports basic profiling for IO latency and cache hit / miss ration, which provides an insight on workload characterization
- Exposes function to get cache size and cleanup cache
- Provide option to disable / enable cache, which could act as a drop-in replacement for httpfs

Wondering if you might be interested in these features. :)

wuputah:
Hi @dentiny thanks for posting this. I think your extension could be a good substitute, and we'd be happy to separate this out and make it opt-in.

dentiny:
> Hi @dentiny thanks for posting this. I think your extension could be a good substitute, and we'd be happy to separate this out and make it opt-in.

Thank you so much for the reply! Let me know if there's anything I could help!

Issue #56, opened 2024-06-24
Switch from replacement scans to using catalogs


Mytherin:
* This allows us to insert into tables by defining a custom insert operator
* This makes schemas work natively without needing the current work-around

JelteF:
Closing this since #97 was merged

PR #97, opened 2024-08-07
Add the `PostgresStorageExtension` to DuckDB 

Tishj:
This PR aims to implement #56 

It is still very barebones, a lot of the API's virtual methods are not implemented, but all existing tests are passing.

The existing functionality has been preserved as best as possible:
- the `PostgresIndexTable` TableCatalogEntry will use the `PostgresIndexScanFunction`
- the `PostgresHeapTable` TableCatalogEntry will use the `PostgresSeqScanFunction`
- VIEWs are still handled by the ReplacementScan, replacing the view with its view_definition, which will get bound again and hit a PostgresIndexTable / PostgresHeapTable

I could find no tests for the PostgresIndexScanFunction, it's possible that functionality has been broken, but I've tried to keep as much of the existing code in tact.


Tishj:
Now that the duckdb connection is a singleton this needs some more work, as it currently assumes the catalog to be thrown away after every query.
The created catalog entries need to be part of the Transaction now so they can be cleared on commit

wuputah:
those changes aren't on main

JelteF:
> those changes aren't on main, but it is something to think about going forward.

For clarity, @Y-- and me are working together for a POC that can be used to show motherduck. But those PRs are not being merged in main yet (they are not nearly ready)

JelteF:
Can you fix the merge conflicts with master? Then it should be much easier to try out. Also when compiling this PR locally I see a bunch of warnings.

src/scan/postgres_seq_scan.cpp
```diff
@@ -14,7 +14,7 @@ PostgresSeqScanGlobalState::PostgresSeqScanGlobalState(Relation relation, duckdb
       m_heap_reader_global_state(duckdb::make_shared_ptr<HeapReaderGlobalState>(relation)), m_relation(relation) {
 	m_global_state->InitGlobalState(input);
 	m_global_state->m_tuple_desc = RelationGetDescr(m_relation);
-	elog(DEBUG3, "-- (DuckDB/PostgresReplacementScanGlobalState) Running %lu threads -- ", MaxThreads());
+	elog(DEBUG3, "-- (DuckDB/PostgresReplacementScanGlobalState) Running %llu threads -- ", MaxThreads());
```
JelteF, line 17: This change is actually causing a compiler warning for me. (I'm on Linux)

```
src/scan/postgres_seq_scan.cpp: In constructor ‘pgduckdb::PostgresSeqScanGlobalState::PostgresSeqScanGlobalState(Relation, duckdb::TableFunctionInitInput&)’:
src/scan/postgres_seq_scan.cpp:17:22: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘idx_t’ {aka ‘long unsigned int’} [-Wformat=]
   17 |         elog(DEBUG3, "-- (DuckDB/PostgresReplacementScanGlobalState) Running %llu threads -- ", MaxThreads());
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~
      |                                                                                                           |
      |                                                                                                           idx_t {aka long unsigned int}
```

include/pgduckdb/catalog/pgduckdb_table.hpp
```diff
@@ -0,0 +1,71 @@
...
+	// -- Table API --
+	unique_ptr<BaseStatistics> GetStatistics(ClientContext &context, column_t column_id) override;
+	TableFunction GetScanFunction(ClientContext &context, unique_ptr<FunctionData> &bind_data) override;
+	TableStorageInfo GetStorageInfo(ClientContext &context) override;
+
+private:
+	Oid oid;
+};
+
+class PostgresIndexTable : public PostgresTable {
```
JelteF, line 55: What is an `IndexTable`? Would be good to add a small code comment explaining that.

src/catalog/pgduckdb_catalog.cpp
```diff
@@ -0,0 +1,126 @@
...
+PostgresCatalog::CreateSchema(CatalogTransaction transaction, CreateSchemaInfo &info) {
+	throw NotImplementedException("CreateSchema not supported yet");
+}
+
+optional_ptr<SchemaCatalogEntry>
+PostgresCatalog::GetSchema(CatalogTransaction transaction, const string &schema_name, OnEntryNotFound if_not_found,
+                           QueryErrorContext error_context) {
+	if (schema_name == DEFAULT_SCHEMA) {
+		return GetSchema(transaction, "public", if_not_found, error_context);
+	}
```
JelteF, line 66: Let's add a TODO comment here to make this work with the postgres search_path instead. Having `main` remapped to `public` only works when the Postgres search_path is unmodified.

src/scan/postgres_seq_scan.cpp
JelteF: I think this should work:
```suggestion
	elog(DEBUG3, "-- (DuckDB/PostgresReplacementScanGlobalState) Running " UINT64_FORMAT " threads -- ", MaxThreads());
```

src/pgduckdb_duckdb.cpp
```diff
@@ -151,9 +161,14 @@ DuckdbCreateConnection(List *rtables, PlannerInfo *planner_info, List *needed_co
...
 	/* Add DuckDB replacement scan to read PG data */
 	auto con = duckdb::make_uniq<duckdb::Connection>(db);
-	con->context->registered_state->Insert(
-	    "postgres_scan", duckdb::make_shared_ptr<PostgresReplacementScanDataClientContextState>(rtables, planner_info,
-	                                                                                            needed_columns, query));
+	auto &context = *con->context;
+	
+	context.registered_state->Insert(
+	    "postgres_state", duckdb::make_shared_ptr<PostgresContextState>(rtables, planner_info, needed_columns, query));
+	auto res = context.Query("set search_path='pgduckdb.main'", false);
```
JelteF, line 168: Regarding the tests that are currently failing:
```
2024-09-16 13:40:53.576 UTC [4126] WARNING:  (DuckDB) Catalog Error: Table with name test_table does not exist!
	Did you mean "pg_tables"?
	LINE 1: EXPLAIN SELECT count(*) AS count FROM test_table
```

I guess it's because the search_path in postgres is public. And here we don't include postgres. Given that search_path in postgres can be modified from the default (and often is). I think there are two options:
1. Always send fully-qualified tables to duckdb.
2. Update search_path in duckdb whenever the search_path in postgres changes.

I think 1 is probably the least error, and I think should be fairly easy. I think if we change the search_path of Postgres temporarily to `pg_catalog` while we call `pgduckdb_pg_get_querydef` then all tables will be fully qualified.

include/pgduckdb/catalog/pgduckdb_table.hpp
Tishj: I couldn't tell you, this is just me moving around the existing logic, I suspect it's an index scan of the table

Tishj: @mkaruza can you provide some clarification here so we can make this more descriptive / self-explanatory?

src/scan/postgres_seq_scan.cpp
Tishj: Ah that infamous define, yes I've seen it before, I think you're right

src/pgduckdb_duckdb.cpp
Tishj: I thought about this same issue on friday and this morning, I was hoping `pgduckdb_pg_get_querydef` would take care of it but that doesnt seem to be the case.

I thought about solution 2, but I came up with a different solution on duckdb's side which is to add an optional callback to the CatalogSearchPath that is invoked when `CatalogSearchPath::Get()` is called.

We could use that callback to ask postgres what the current `search_path` is and return `CatalogSearchEntry` (catalog + schema) objects for the entries in the `search_path`.

Tishj: > I think 1 is probably the least error, and I think should be fairly easy. I think if we change the search_path of Postgres temporarily to pg_catalog while we call pgduckdb_pg_get_querydef then all tables will be fully qualified.

I don't get what this accomplishes and why that would result in fully qualified tables?

Tishj: @JelteF I'm not sure why this is failing in the CI but it isn't failing locally:
```sql
-- search_path ORDER matters.
SET search_path TO other, public;
SELECT count(*) FROM t;
SELECT count(*) FROM t, public.t;
```

this should highlight the problem but these pass for me locally

Tishj: Ah actually my expected results are wrong, it is broken.
I'm going to see if the solution I proposed above can quickly be implemented to work here

JelteF: > I don't get what this accomplishes and why that would result in fully qualified tables?

The way `pgduckdb_pg_get_querydef` works is that it fully qualifies all tables in the query, unless that is unnecessary because their schema is in the current Postgres search_path. So by temporarily changing the Postgres search_path to `pg_catalog, non of the tables will be in search_path (except for tables in pg_catalog but we need a different solution for that anyway and we dissallow duckdb execution for those tables because of that). To be clear this is because `pg_get_querydef` works that way, and now that we vendored this function in we could change this behaviour, but it seems nicer not to touch the vendored in code unless strictly necessary. Especially because I can get Postgres to create fully qualified names using by applying this tiny diff to the callsite:

```diff
+	auto save_nestlevel = NewGUCNestLevel();
+	SetConfigOption("search_path", "pg_catalog", PGC_USERSET, PGC_S_SESSION);
	const char *query_string = pgduckdb_pg_get_querydef(query, false);
+	AtEOXact_GUC(false, save_nestlevel);
```

For reference the decision on whether to qualify or not qualify can be found here: https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/ruleutils.c#L12996-L12998

Tishj: Ah thanks that makes sense, we only unset the `search_path` for `pg_get_querydef`, not for parsing.

src/catalog/pgduckdb_catalog.cpp
JelteF: I think we can remove this if statement now that we have send fully qualified table references to duckdb.

src/pgduckdb_duckdb.cpp
```diff
@@ -151,9 +162,14 @@ DuckdbCreateConnection(List *rtables, PlannerInfo *planner_info, List *needed_co
...
 	/* Add DuckDB replacement scan to read PG data */
 	auto con = duckdb::make_uniq<duckdb::Connection>(db);
-	con->context->registered_state->Insert(
-	    "postgres_scan", duckdb::make_shared_ptr<PostgresReplacementScanDataClientContextState>(rtables, planner_info,
-	                                                                                            needed_columns, query));
+	auto &context = *con->context;
+
+	context.registered_state->Insert(
+	    "postgres_state", duckdb::make_shared_ptr<PostgresContextState>(rtables, planner_info, needed_columns, query));
+	auto res = context.Query("set search_path='pgduckdb.main'", false);
```
JelteF, line 169: Is this line still needed now?

src/pgduckdb_planner.cpp
```diff
@@ -106,7 +110,10 @@ CreatePlan(Query *query, const char *query_string, ParamListInfo bound_params) {
 
 PlannedStmt *
 DuckdbPlanNode(Query *parse, int cursor_options, ParamListInfo bound_params) {
+	auto save_nestlevel = NewGUCNestLevel();
+	SetConfigOption("search_path", "pg_catalog", PGC_USERSET, PGC_S_SESSION);
```
JelteF, line 114: Let's add some comment explaining why we're doing this. Also I realized setting search_path to the empty string is functionally equivalent to setting it to pg_catalog and it seems less confusing so let's do that instead.
```suggestion
	/* Temporarily clear search_path so that the query will contain fully qualified tables. If we don't do this tables are only fully-qualified if they are not part of the current search_path. */
	auto save_nestlevel = NewGUCNestLevel();
	SetConfigOption("search_path", "", PGC_USERSET, PGC_S_SESSION);
```

src/pgduckdb_duckdb.cpp
JelteF: At least searching the `main` schema by default isn't necessary anymore I think. Maybe change it to:
```suggestion
	auto res = context.Query("USE pgduckdb", false);
```

JelteF: Side-note: I think in a future PR we might want to change the the vendored in ruleutils to include the `pgduckdb` catalog in the fully table name in a query i.e. `pgduckdb.public.t` instead of `public.t`. But that can wait until we actually need it, e.g. to avoid ambiguities once we support having multiple databases open in duckdb.

test/regression/expected/views.out
```diff
@@ -27,10 +27,6 @@ create table "s.t" as select 42;
 create view vw1 as select * from s.t;
 create view vw2 as select * from "s.t";
 select * from vw1, vw2;
-WARNING:  (DuckDB) Binder Error: Referenced table "t" not found!
```
JelteF, line 30: Great to see that this is now fixed!

src/pgduckdb_planner.cpp
```diff
@@ -63,6 +64,9 @@ CreatePlan(Query *query, const char *query_string, ParamListInfo bound_params) {
 
 	PlannerInfo *query_planner_info = PlanQuery(query, bound_params);
 	auto duckdb_connection = pgduckdb::DuckdbCreateConnection(rtables, query_planner_info, vars, query_string);
+	if (!duckdb_connection) {
```
JelteF, line 67: In what cases is this expected to happen? Afaict when reading the `DuckdbCreateConnection` function this should not be possible.

src/scan/postgres_scan.cpp
```diff
@@ -168,58 +87,47 @@ PostgresReplacementScan(duckdb::ClientContext &context, duckdb::ReplacementScanI
...
-		auto table_function = duckdb::make_uniq<duckdb::TableFunctionRef>();
-		table_function->function =
-		    duckdb::make_uniq<duckdb::FunctionExpression>("postgres_seq_scan", std::move(children));
-		table_function->alias = table_name;
-		return std::move(table_function);
+	duckdb::Parser parser;
+	parser.ParseQuery(view_definition);
+	auto statements = std::move(parser.statements);
+	if (statements.size() != 1) {
+		elog(ERROR, "View definition contained more than 1 statement!");
```
JelteF, line 121: Let's throw exceptions here instead, given the guidance in `elog(ERROR)` provided in #176

src/pgduckdb_planner.cpp
Tishj: I made a modification to `DuckdbCreateConnection` so it could return nullptr, but it didn't stick - I'll remove this branch 👍

src/pgduckdb_duckdb.cpp
JelteF: okay, seems reasonable to me. Resolving this then.

include/pgduckdb/catalog/pgduckdb_catalog.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
```
JelteF, line 1: @bleskes was wondering what the benefits are of making this change. He thought you probably had some good reasons, but that it would be good to call those out clearly in the PR description.

Afaict one reason is to be able to support inserts using the custom INSERT. Are there more?

If it's only inserts, another approach to be to split inserts into two:

1. A read-only part, which is executed by DuckDB
2. A writing part that actually writes into the table, which takes the output from the read-only part

Tishj: Anything beyond a simple replacement scan will need this, that's why we're opting to add it sooner rather than later.
I already need it for the enum type support:

```
(lldb) p query 
(const std::string &) "SELECT a FROM tbl WHERE (a = 'sad'::mood)"
```

The produced query from postgres contains `::mood` which means it is going to do a lookup in the catalog for the `mood` type, but that doesn't exist

Tishj: > If it's only inserts, another approach for supporting inserts would to be to split inserts into two:

A read-only part, which is executed by DuckDB, this generates the data that needs to be inserted
A writing part that actually writes into the table, which is executed Postgres and takes the output from the read-only part as its input

Isn't that exactly what this facilitates?
By overloading `PlanInsert` we can create a PhysicalOperator that performs the insert into postgres, being fed data from the duckdb plan

This is `DuckCatalog::PlanInsert`
```c++
unique_ptr<PhysicalOperator> DuckCatalog::PlanInsert(ClientContext &context, LogicalInsert &op,
                                                     unique_ptr<PhysicalOperator> plan) {
	bool parallel_streaming_insert = !PhysicalPlanGenerator::PreserveInsertionOrder(context, *plan);
	bool use_batch_index = PhysicalPlanGenerator::UseBatchIndex(context, *plan);
	auto num_threads = TaskScheduler::GetScheduler(context).NumberOfThreads();
	if (op.return_chunk) {
		// not supported for RETURNING (yet?)
		parallel_streaming_insert = false;
		use_batch_index = false;
	}
	if (op.action_type != OnConflictAction::THROW) {
		// We don't support ON CONFLICT clause in batch insertion operation currently
		use_batch_index = false;
	}
	if (op.action_type == OnConflictAction::UPDATE) {
		// When we potentially need to perform updates, we have to check that row is not updated twice
		// that currently needs to be done for every chunk, which would add a huge bottleneck to parallelized insertion
		parallel_streaming_insert = false;
	}
	unique_ptr<PhysicalOperator> insert;
	if (use_batch_index && !parallel_streaming_insert) {
		insert = make_uniq<PhysicalBatchInsert>(op.types, op.table, op.column_index_map, std::move(op.bound_defaults),
		                                        std::move(op.bound_constraints), op.estimated_cardinality);
	} else {
		insert = make_uniq<PhysicalInsert>(
		    op.types, op.table, op.column_index_map, std::move(op.bound_defaults), std::move(op.bound_constraints),
		    std::move(op.expressions), std::move(op.set_columns), std::move(op.set_types), op.estimated_cardinality,
		    op.return_chunk, parallel_streaming_insert && num_threads > 1, op.action_type,
		    std::move(op.on_conflict_condition), std::move(op.do_update_condition), std::move(op.on_conflict_filter),
		    std::move(op.columns_to_fetch));
	}
	D_ASSERT(plan);
	insert->children.push_back(std::move(plan));
	return insert;
}
```

JelteF: > The produced query from postgres contains `::mood` which means it is going to do a lookup in the catalog for the `mood` type, but that doesn't exist

That's a good point. I'm wondering whether we can get around this by simply removing `::mood` from the query that is generated by postgres by changing `duckdb_pg_get_querydef` to not include that cast. Duckdb does not actually seem to need it when I try some queries locally. But I might be missing some more complex queries where it actually requires the cast to be present.

> Isn't that exactly what this facilitates?

What I meant is that we don't necessarily need to send the `INSERT` to duckdb. So if a user runs:

```
INSERT INTO t SELECT * FROM read_csv(...);
```

Then we instead send `SELECT * FROM read_csv(...)` to duckdb, and let postgres logic insert the resulting tuples into table `t`.

Tishj: Yea that could work, using a streaming result we could even start inserting while the select is still executing

But like I said INSERT is definitely not the only reason we need this

JelteF: Yeah makes sense. For completeness, I found a type of query where where the casts definitely matter, and we cannot just leave them out:

```sql
D CREATE TYPE numbers AS ENUM ('3', '2', '1');
D select '1' < '3';
┌─────────────┐
│ ('1' < '3') │
│   boolean   │
├─────────────┤
│ true        │
└─────────────┘
D select '1'::numbers < '3'::numbers;
┌───────────────────────────────────────────────┐
│ (CAST('1' AS numbers) < CAST('3' AS numbers)) │
│                    boolean                    │
├───────────────────────────────────────────────┤
│ false                                         │
└───────────────────────────────────────────────┘
```

Tishj: Just to add a little more context:
`SELECT a FROM tbl WHERE (a = 'sad'::mood)` was produced from the query string:
`SELECT a FROM tbl WHERE a = 'sad'`

This works without the explicit cast because comparison between string and enum is not allowed, so it does a cast from the string to the enum it's being compared to

Tishj:
when I merge with main I can't resolve a crash in `test/regression/sql/query_filter.sql`
`SELECT a, c FROM query_filter_output_column WHERE b = 't1';` is what triggers it, it's a EXC_BAD_ACCESS somewhere in the query profiler.

I suspect that the query data gets prematurely deleted, it's duckdb data it's trying and failing to access.
I don't know why this query specifically is causing the problem, I know there was recently work done on filter / projection pushdown?

JelteF (approved):

src/pgduckdb_planner.cpp
```diff
@@ -139,6 +151,25 @@ CreatePlan(Query *query, ParamListInfo bound_params) {
...
+	const char *query_string = pgduckdb_pg_get_querydef(parse, false);
+	AtEOXact_GUC(false, save_nestlevel);
+
+	if (ActivePortal && ActivePortal->commandTag == CMDTAG_EXPLAIN) {
+		if (duckdb_explain_analyze) {
+			query_string = psprintf("EXPLAIN ANALYZE %s", query_string);
+		} else {
+			query_string = psprintf("EXPLAIN %s", query_string);
+		}
+	}
```
JelteF, line 172: I think this can go away now

src/pgduckdb_planner.cpp
```diff
@@ -55,12 +56,23 @@ PlanQuery(Query *parse, ParamListInfo bound_params) {
...
 	/*
 	 * Copy the query, so the original one is not modified by the
 	 * subquery_planner call that PlanQuery does.
 	 */
 	Query *copied_query = (Query *)copyObjectImpl(query);
+	/*
+	    Temporarily clear search_path so that the query will contain only fully qualified tables.
+	    If we don't do this tables are only fully-qualified if they are not part of the current search_path.
+	    NOTE: This still doesn't fully qualify tables in pg_catalog or temporary tables, for that we'd need to modify
+	   pgduckdb_pg_get_querydef
```
JelteF, line 69: super nitpick: it looks a bit weird to me that there are no `*`'s in front of every line of the comment.

test/regression/Makefile
```diff
@@ -9,7 +9,7 @@ check-regression-duckdb:
 	--temp-instance=./tmp_check \
 	--temp-config regression.conf \
 	--load-extension=pg_duckdb \
-	--schedule schedule
+	$(if $(TEST),$(TEST),--schedule=schedule)
```
JelteF, line 12: What is this about

JelteF:
> when I merge with main I can't resolve a crash in test/regression/sql/query_filter.sql
SELECT a, c FROM query_filter_output_column WHERE b = 't1'; is what triggers it, it's a EXC_BAD_ACCESS somewhere in the query profiler.

Did you fix this? Because I see you merged with main afterwards, and tests seem to pass now on CI.

JelteF (approved):
Okay, I think this is pretty much good to merge. Left some final comments. The major one being the (seemingly) incorrect merge. But all should be easy to address afaict.

JelteF:
> I'd like to stop having to fix this with every PR that gets merged

Totally makes sense. Let's merge this quickly now.

Tishj:
> > when I merge with main I can't resolve a crash in test/regression/sql/query_filter.sql
> > SELECT a, c FROM query_filter_output_column WHERE b = 't1'; is what triggers it, it's a EXC_BAD_ACCESS somewhere in the query profiler.
> 
> Did you fix this? Because I see you merged with main afterwards, and tests seem to pass now on CI.

it's not, I can reproduce this on main and the PR that introduced it is https://github.com/duckdb/pg_duckdb/pull/195

Mytherin (approved):

PR #98, opened 2024-08-07
Set the version of clang-format to be the same as used in DuckDB

Tishj:
Behavior differs between versions of clang-format, to be sure that the entire codebase doesn't have to be reformatted when clang-format updates I've locked it to 11.0.1 (the same that DuckDB uses)

Primary reason I'm making this change is that I am currently having to switch back and forth my clang-format version, which I would prefer to not do.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -15,7 +15,6 @@ AllowAllConstructorInitializersOnNextLine: true
 AllowAllParametersOfDeclarationOnNextLine: true
 AlwaysBreakAfterReturnType: AllDefinitions
 AlignAfterOpenBracket: Align
-BracedInitializerIndentWidth: 4
 IncludeBlocks: Preserve
 IncludeCategories: # we want to ensure postgres.h appear first
   - Regex:           '^"postgres\.h"'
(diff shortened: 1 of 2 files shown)


Y-- (approved):
LGTM!

Tishj:
It's not running the CI, I'm having some trouble figuring out why it's not

JelteF:
Yeah it makes sense to align clang-format versions across different duckdb projects. And changing it here seems easiest by far. It might be worth bumping the clang-format version on the repos soonish again though, since it's 7 major versions behind at this point.

JelteF (approved):

JelteF:
> It's not running the CI, I'm having some trouble figuring out why it's not

Maybe try something like this:
```
git commit --allow-empty 'bump ci'
git push
```

GitHub actions sometimes have a bad moment.

wuputah:
ah, ci didnt run on that last commit because `.clang-format` isn't in the list here:
https://github.com/duckdb/pg_duckdb/blob/de8fa8e3590ff6cb1be86607ea3ab7af305a9590/.github/workflows/build_and_test.yaml#L4-L13

JelteF (approved):

Issue #36, opened 2024-05-27
Add `postgres_mode` to DuckDB that makes it simulate Postgres in a better capacity


Mytherin:
This mode is a (very long tail) mode that improves compatibility between DuckDB's SQL dialect and Postgres' SQL dialect.

e.g.:

* `SELECT COUNT(*) FROM tbl` returns a column named `count`, instead of `count_star()`

JelteF:
I think all the things in this issue have now been solved/worked around some other way. We might still need a "postgres_mode" feature in duckdb for other reasons, but let's create separate issues for these things when we find them.

PR #99, opened 2024-08-07
Read functions now supports setting optional arguments

mkaruza:
* For functions read_parquet / read_csv we can now set custom arguments


sql/pg_duckdb--0.0.1.sql
```diff
@@ -1,30 +1,97 @@
...
 BEGIN
     RAISE EXCEPTION 'Function `read_parquet(TEXT[])` only works with Duckdb execution.';
 END;
 $func$;
 
-CREATE OR REPLACE FUNCTION read_csv(path text)
+
+-- Arguments 'columns' and 'nullstr' are currently not supported for read_csv
+
+CREATE OR REPLACE FUNCTION read_csv(path text, all_varchar BOOLEAN DEFAULT FALSE,
```
wuputah, line 30: I see `columns` is missing, is that because we can't pass the hash/JSON syntax?

```
 columns => {
        'FlightDate': 'DATE',
        'UniqueCarrier': 'VARCHAR',
        'OriginCityName': 'VARCHAR',
        'DestCityName': 'VARCHAR'
    }```

File: sql/pg_duckdb--0.0.1.sql [hunk 2]
```diff
@@ -1,30 +1,97 @@
...
 RETURNS SETOF record LANGUAGE 'plpgsql' AS
 $func$
 BEGIN
     RAISE EXCEPTION 'Function `read_parquet(TEXT)` only works with Duckdb execution.';
 END;
 $func$;
 
-CREATE OR REPLACE FUNCTION read_parquet(path text[])
+CREATE OR REPLACE FUNCTION read_parquet(path text[], binary_as_string BOOLEAN DEFAULT FALSE,
+                                                     filename BOOLEAN DEFAULT FALSE,
```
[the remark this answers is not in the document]
Tishj: Right that was changed on `main`, I guess we can revisit this when updating to v1.1.0 in roughly a month

sql/pg_duckdb--0.0.1.sql
mkaruza: Yes, not sure is there is postgresql way to provide struct for this option.

wuputah: Supporting this option isn't a big deal in itself, but it's a bit annoying to say "everything is supported except `columns`."

Supporting `struct_pack` is close but it wants bare key names. 

The [JSON type][1] does have a way to convert to struct, but the `::struct` part would fail in Postgres, and is very verbose.

```sql
SELECT '{"duck": 42}'::JSON::STRUCT(duck INTEGER);
```

@Tishj any ideas on how to get structs from Postgres into DuckDB?

[1]: https://duckdb.org/docs/extensions/json#json-type

Tishj: Depends on if this will autocast on the duckdb side, providing the struct as a string could let it cast if that's the case

mkaruza: I think that closes to struct that we could have is to declare columns to be `JSON` type and argument as:

```
columns => '{"col1" : "type1", "col2" : "type2"}'
```

This is going to be received as `VARCHAR` type in `CSV extension` so probably we would need to cast/convert it to struct type (maybe do this only if postgresql mode is set #36).

wuputah (approved):

JelteF (approved):

JelteF (approved):

JelteF (approved):

PR #102, opened 2024-08-07
Make DuckDB a singleton per con process

Y--:
* add a `DuckDBManager` class to hold the singleton and manage instantiation
* give `PostgresReplacementScanData` an identifier (auto-incremented integer)
* extends `DuckDB::Connection` to hold replacement scan id and remove it when it is destroyed

### DIFF
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -10,8 +10,47 @@ extern "C" {
 
 namespace pgduckdb {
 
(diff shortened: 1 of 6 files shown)


Y--:
Note that replacement scans are not isolated per connections, so while each connection keeps track of the scans it adds, it is still possible that two (or more) `DuckDB::Connection` conflict with each others using the wrong scans.
In practice it seems unlikely though since one PG connection will likely only use one DuckDB connection at a time.

JelteF (approved):

PR #103, opened 2024-08-07
Process `~/.pg_duckrc` if it exists

Y--:


wuputah:
What is this for? It's useful for our development work maybe, but I wouldn't want this behavior on my production Postgres server.

Y--:
> What is this for? It's useful for our development work maybe, but I wouldn't want this behavior on my production Postgres server.

Hi @wuputah! Thanks for reviewing the PR :-) This PR is not meant to be merged in main as-is. Like you said it is something useful to have for development purposes, especially when you have a few commands to issue to connect to MotherDuck for example ;-)

wuputah:
ah, I missed this:

> 1 commit into [yl/ddb-per-con](https://github.com/duckdb/pg_duckdb/tree/yl/ddb-per-con) from [yl/pgduckrc](https://github.com/duckdb/pg_duckdb/tree/yl/pgduckrc)

We could put this on `main` behind a build-time flag (e.g. -DDEBUG) or similar.

JelteF (approved):

PR #107, opened 2024-08-08
Minimal create table support

Y--:


src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,141 @@
...
+	 * method. See the code comment in that function for more details.
+	 */
+	int ret = SPI_exec(R"(
+                INSERT INTO duckdb.tables(relid)
+                SELECT objid
+                FROM pg_catalog.pg_event_trigger_ddl_commands() cmds
+                JOIN pg_catalog.pg_class
+                ON cmds.objid = pg_class.oid
+                WHERE cmds.object_type = 'table'
+                AND pg_class.relam = (SELECT oid FROM pg_am WHERE amname = 'duckdb'))",
```
Y--, line 64: Sorry for the stupid question - what guarantees us that what we select here is the one and only statement that triggered the event which led us to this point? (i.e can't we have other connections/background tasks firing concurrently?)

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,141 @@
...
+	                   0);
+
+	/* Revert back to original privileges */
+	SetUserIdAndSecContext(saved_userid, sec_context);
```
Y--, line 68: Is it possible for `SPI_exec` to error out/longjmp leading this line to not be executed? (I see the error check just below, but maybe there are fatal cases where it just doesn't return?)

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+	if (ret != SPI_OK_INSERT)
+		elog(ERROR, "SPI_exec failed: error code %d", ret);
+
+	if (SPI_processed == 0) {
+		/* it was a regular postgres table, nothing todo */
+		SPI_finish();
+		PG_RETURN_NULL();
+	}
+	SPI_finish();
```
Y--, line 78: nit: should we just move `SPI_finish();` before the `if` since we're doing it in the two paths?

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,141 @@
...
+                ON cmds.objid = pg_class.oid
+                WHERE cmds.object_type = 'table'
+                AND pg_class.relam = (SELECT oid FROM pg_am WHERE amname = 'duckdb'))",
+	                   0);
+
+	/* Revert back to original privileges */
+	SetUserIdAndSecContext(saved_userid, sec_context);
+
+	if (ret != SPI_OK_INSERT)
+		elog(ERROR, "SPI_exec failed: error code %d", ret);
```
Y--, line 71: Is there a way to translate `ret` to a string here?

src/pgduckdb_ddl.cpp
JelteF: We're reading from `pg_catalog.pg_event_trigger_ddl_commands()` this only returns info about the ddl commands for this event trigger.

src/pgduckdb_ddl.cpp
JelteF: I had to look this up, because I wasn't 100% sure. But AbortTransaction rolls back any changes that SetUserIdAndSecContext made on error.

src/pgduckdb_ddl.cpp
JelteF: Done

src/pgduckdb_ddl.cpp
JelteF: done

PR #109, opened 2024-08-08
Cache pg & duckdb builds

Y--:
Bring the build time from [~4minutes](https://github.com/duckdb/pg_duckdb/actions/runs/10310110288/attempts/1) down to [~1 minute](https://github.com/duckdb/pg_duckdb/actions/runs/10310257549) by caching PG and DuckDB's builds.

* PG is cached based on the branch checked out
* DuckDB is cached based on the sha committed as the submodule

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -53,9 +53,26 @@ jobs:
       - name: Checkout pg_duckdb extension code
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


wuputah (approved):
Nice.

Tangential, but we could move to build off Postgres tags as well rather than building off of a release branch, would just mean swapping out `REL_16_STABLE` for something like `REL_16_4` (though it's also nice to test against the latest changes, too).

PR #110, opened 2024-08-09
POC: connecting to MotherDuck

JelteF:
This is a PR to our MotherDuck POC branch. This adds changes to be able to
connect to MotherDuck. It also rolls back the DuckDB version to v1.0.0, since
that is the latest supported by MotherDuck.

Add this to `~/.pg_duckrc`

```sql
LOAD 'motherduck';

SET motherduck_token='<your token>';

PRAGMA md_connect;
```

Then in `posgresql.conf` configure the following:
```conf
duckdb.default_db='my_motherduck_database'
```

Once you've done both of those things the following snippet should work and
create a `md_table` table in your MotherDuck database:

```sql
CREATE TABLE pg_table AS SELECT a FROM generate_series(1, 100) a;
CREATE TABLE md_table(b int) USING duckdb;
INSERT INTO md_table SELECT * FROM pg_table where a > 50;
SELECT count(*) FROM pg_table, md_table WHERE a = b;
```


Y-- (approved):
Awesome!

Makefile
```diff
@@ -79,13 +79,16 @@ third_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src/$(DUCKDB_LIB):
...
+	BUILD_JSON=1 \
+	CMAKE_EXPORT_COMPILE_COMMANDS=1 \
+	-j8
 
 install-duckdb:
 	$(install_bin) -m 755 third_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src/$(DUCKDB_LIB) $(DESTDIR)$(PG_LIB)
 
 clean-duckdb:
-	rm -rf third_party/duckdb/build
+#rm -rf third_party/duckdb/build
```
Y--, line 91: Should we keep that?

PR #111, opened 2024-08-09
Add support for CTAS USING duckdb w/ PG tables

Y--:
Add support for CTAS such as:
```sql
CREATE TABLE md_table2 USING duckdb AS SELECT * FROM pg_table;
```


Y--:
This seems to break the simple `CREATE TABLE md_table USING duckdb` though. I'll investigate deeper tomorrow.

src/pgduckdb_ddl.cpp
```diff
@@ -93,7 +94,16 @@ duckdb_create_table_trigger(PG_FUNCTION_ARGS) {
 	// TODO: This is a huge hack, we should build the query string from the parsetree
 	std::string query_string = std::regex_replace(debug_query_string, std::regex(R"((using|USING)\s+duckdb)"), "");
 
-	auto connection = pgduckdb::DuckdbCreateSimpleConnection();
+	CreateTableAsStmt *stmt = castNode(CreateTableAsStmt, trigdata->parsetree);
```
JelteF, line 97: This needs to be in an `if (IsA(parsetree, CreateTableAsStmt))` block. That's almost certainly why plain `CREATE TABLE` fails now.

JelteF (approved):

PR #114, opened 2024-08-13
Allow make install to use -j flag

JelteF:
The `install-duckdb` target would be run before the duckdb target was
actually built. This caused a failure when running make in parallel like
this:
```
make install -j20
```

At the top of the make output you would then see the following error:
```
/usr/bin/install: cannot stat 'third_party/duckdb/build/release/src/libduckdb.so': No such file or directory
make: *** [Makefile:83: install-duckdb] Error 1
make: *** Waiting for unfinished jobs....
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,12 +66,14 @@ clean-regression:
 
 installcheck: all install check-regression-duckdb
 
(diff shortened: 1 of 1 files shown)


wuputah (approved):
👍

PR #117, opened 2024-08-14
Fix a typo for global_state variable

japinli:
As $subject said, there is a variable name typo in [src/scan/postgres_seq_scan.cpp](https://github.com/duckdb/pg_duckdb/blob/main/src/scan/postgres_seq_scan.cpp#L113).


Tishj (approved):
Thanks!

hlinnaka (approved):

PR #123, opened 2024-08-14
Automatically enable DuckDB execution when DuckDB-only functions are used

JelteF:
Since #86 we disable `duckdb.execution` by default, because existing
queries might fail in DuckDB. However, if people use a DuckDB-only
function such as `read_parquet` there's no reason to try running it in
Postgres, because it only works when run in DuckDB.

This starts enabling DuckDB execution for queries automatically when
such a function is used. And also adds test coverage in general for
`read_csv` and `read_parquet`.


src/pgduckdb_metadata_cache.cpp
```diff
@@ -29,17 +85,32 @@ IsExtensionRegistered() {
...
 		schema_hash_value = GetSysCacheHashValue1(NAMESPACENAME, CStringGetDatum("duckdb"));
 
 		CacheRegisterSyscacheCallback(NAMESPACENAME, InvalidateCaches, (Datum)0);
 	}
 
+	cache.installed = get_extension_oid("pg_duckdb", true) != InvalidOid;
+	if (cache.installed) {
+		BuildDuckdbOnlyFunctions();
+	}
+	cache.valid = true;
```
Y--, line 99: Maybe I'm reading this wrong, but it looks like it's possible to end up with `cache.installed == false && cache.valid == true` - shouldn't the valid flag be set only in the `installed` path?

Y-- (commented):
LGTM, just a small question on the cache

sql/pg_duckdb--0.0.1.sql
```diff
@@ -1,38 +1,38 @@
 LOAD 'pg_duckdb';
 
-CREATE OR REPLACE FUNCTION read_parquet(path text)
+CREATE OR REPLACE FUNCTION public.read_parquet(path text)
```
JelteF, line 3: I'm adding public here so that these functions get created in a consistent schema.

src/pgduckdb_metadata_cache.cpp
JelteF: valid == is the cache initialized. I added a bunch of comments in the file to make this much clearer.

Y-- (approved):
Looks great!

sql/pg_duckdb--0.0.1.sql
JelteF: undid this change now that I find the functions in a different way.

JelteF:
@mkaruza I realized the initial version of this PR would conflict with your #99 PR, since you're changing the argument types there. Keeping these argument types in sync seems like it could get annoying and also not very readable. So instead I changed the code a bit to not need the argument types.

src/pgduckdb_metadata_cache.cpp
```diff
@@ -1,45 +1,157 @@
...
+	Assert(!cache.duckdb_only_functions);
+	Assert(cache.extension_oid != InvalidOid);
+
+	/*
+	 * We search the system cache for functions with these specific names. It's
+	 * possible that other functions with the same also exist, so we check if
+	 * each of the found functions is actually part of our extension before
+	 * caching its OID as a DuckDB-only function.
+	 */
+	const char *function_names[] = {"read_parquet", "read_csv", "iceberg_scan"};
```
mkaruza, line 80: Is there a way to fetch all functions that are installed via extension - rather than keep list hardcoded here?

src/pgduckdb_metadata_cache.cpp
```diff
@@ -1,45 +1,157 @@
...
+	 * Is the pg_duckdb extension installed? If this is false all the other
+	 * fields (except valid) should be ignored. It is totally fine to have
+	 * valid=true and installed=false, this happens when the extension is not
+	 * installed and we cached that information.
+	 */
 	bool installed;
+	/* The Postgres OID of the pg_duckdb extension. */
+	Oid extension_oid;
+	/* A list of Postgres OIDs of functions that can only be executed by DuckDB */
+	List *duckdb_only_functions;
```
mkaruza, line 38: Not mandatory but maybe use Bitmapset instead of List ?

JelteF: I did consider that, but `Oid` is `uint32` and Bitmapset can only store `int32`. So that doesn't work (afaict).

src/pgduckdb_metadata_cache.cpp
mkaruza: just quick note, functions are used only in duckdb live in `public` schema while this function can be found in `duckdb`. But i agree that this is something to be reconsider in future if there need for it.

src/pgduckdb_metadata_cache.cpp
mkaruza: Sure and as comment above - at this point list is going to be small to have cause any problems.

JelteF: I included a note about the datastructure choice in the codecomment now.

mkaruza (approved):

PR #130, opened 2024-08-16
Add support for CREATE TABLE AS

JelteF:
This adds support for queries like:

```
CREATE TABLE mytable AS SELECT column00 from read_csv("my_data.csv") AS (column00 int)
```

The table will be then be created in Postgres (not in DuckDB). The main
change that supports this is that we now start "deparsing" the parsed
`Query` type from Postgres to an actual query string, instead of sending
the original string that Postgres received to DuckDB.

This also accidentally fixes some differences such as the DuckDB
`count(*)` aggregate creating a `count_star` vs Postgres its `count`
column. So it reduces the need for #36 a bit. The reason this happens is
because it is deparsed very explicitly like `count(*) AS count`.

For similar reasons this also changes the behaviour of the following
query to only return the `id` and `name` columns, instead of all the
columns. It's arguable whether that's a good thing or not: it's more
verbose if you want those columns but it's also less surprising.

```
SELECT * FROM read_parquet(...) AS (id int, name text);
```

One especially surprising thing with the above query is that
if you used a column name that was not in the parquet file,
then your query would succeed, but you would not actually
get that column back in the result (only the actual columns
in the file).


test/regression/expected/views.out
```diff
@@ -27,9 +27,13 @@ create table "s.t" as select 42;
...
 create view vw2 as select * from "s.t";
 select * from vw1, vw2;
+WARNING:  (DuckDB) Binder Error: Referenced table "t" not found!
+Candidate tables: "s.t"
+LINE 3:    FROM ( SELECT t."?column?"
+                         ^
  ?column? | ?column? 
 ----------+----------
-       21 |       21
+       21 |       42
```
JelteF, line 36: The previous DuckDB pushdown was actually returning the wrong results here. So the DuckDB failure that this introduces is better than the alternative imho. Because now it warns and automatically falls back to Postgres, instead of returning a wrong result.

Tishj: Yea I noticed that as well, the postgres catalog extension solved the same problem

test/regression/sql/materialized_view.sql
```diff
@@ -1,5 +1,5 @@
-CREATE TABLE t(a INT, b TEXT);
-INSERT INTO t SELECT g % 1000, MD5(g::TEXT) FROM generate_series(1,20000) g;
+CREATE TABLE t(a INT, b VARCHAR);
+INSERT INTO t SELECT g % 1000, MD5(g::VARCHAR) FROM generate_series(1,20000) g;
```
JelteF, line 2: The `SELECT` in the `CREATE MATERIALIZED VIEW` command now also goes through DuckDB. This is a good thing imho, because that means that we can use `read_csv` and `read_parquet` in those queries too.

However, on PG16 our type mapping results in the following error:
```
 CREATE MATERIALIZED VIEW tv AS SELECT * FROM t WHERE a % 100 = 0;
+ERROR:  SELECT rule's target entry 2 has different type from column "b"
+DETAIL:  SELECT target entry has type text, but column has type character varying.
```

Surprisingly in PG17 this works just fine.

While suboptimal, I think this is an improvement overall.

src/pgduckdb_hooks.cpp
```diff
@@ -46,20 +46,18 @@ IsAllowedStatement(Query *query) {
...
 	}
-	/* For `SELECT ..` ActivePortal doesn't exist */
-	if (!ActivePortal)
-		return true;
-	/* `EXPLAIN ...` should be allowed */
-	if (ActivePortal->commandTag == CMDTAG_EXPLAIN)
-		return true;
-	return false;
+	/* We don't support modifying statements yet */
+	if (query->commandType != CMD_SELECT) {
```
mkaruza, line 50: We already check this condition in `DuckdbPlannerHook` if statement but this is better place to contain this check `commandType == SELECT`.

mkaruza (dismissed):

Tishj (approved):
Thanks this sounds like a good change to me 👍

src/pgduckdb_hooks.cpp
JelteF: I moved a few more checks from the DuckDBPlannerHook to IsAllowedStatement now.

mkaruza (approved):

PR #132, opened 2024-08-16
Speed up running tests when compiled with DUCKDB_BUILD=Debug

JelteF:
Without this change it takes more than a minute to run
`make installcheck` on my machine if I enable DUCKDB_BUILD=Debug.
Because both of these tests take more than thirty seconds. After this
change both execute in less than 2 seconds. Especially the first test
taking so long is confusing, because it seems like the tests are not
working.


Y-- (approved):
Very much in favor of this one, thanks!

Tishj (approved):
Thanks, that should help
Slightly off-topic:
We should have at least one "slow" test to test converting large (over 10k) result sets from duckdb -> pg, I'm not sure if we do currently

PR #136, opened 2024-08-19
Add EXPLAIN ANALYZE support

JelteF:
This adds support for forwarding `EXPLAIN ANALYZE` to DuckDB.

A useful side-effect for debugging issues of running `EXPLAIN ANALYZE`, is that it shows the actual query that was executed in duckdb.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ SRCS = src/scan/heap_reader.cpp \
 	   src/scan/postgres_scan.cpp \
 	   src/scan/postgres_seq_scan.cpp \
 	   src/utility/copy.cpp \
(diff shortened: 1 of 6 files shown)


mkaruza:
Can we have test for this?

JelteF:
Since the EXPLAIN ANALYZE output contains timings, which afaict are not possible to turn off in duckdb, the output of EXPLAIN ANALYZE wouldn't be stable. I'll look into copying over and modifying the testing framework that I built for PgBouncer though. It should be very easy to test with that, and we'll probably find more cases where pgregress isn't enough for testing.

mkaruza (approved):

PR #137, opened 2024-08-19
Add support for filtering with varchar

Y--:
Allow to filter with a `VARCHAR`, eg:

```sql
SELECT * FROM varchar_tbl WHERE a = 'test';
```

### DIFF
--- a/src/pgduckdb_filter.cpp
+++ b/src/pgduckdb_filter.cpp
@@ -28,11 +28,8 @@ StringFilterOperation(Datum &value, const duckdb::Value &constant) {
 	}
 
 	bool should_free = false;
-	auto detoasted_value = DetoastPostgresDatum(reinterpret_cast<varlena *>(value), &should_free);
-
-	const auto ptr = (text*)DatumGetPointer(detoasted_value);
-	text* tunpacked = pg_detoast_datum_packed(ptr);
-	const auto datum_sv =  std::string_view((char*)VARDATA_ANY(tunpacked), VARSIZE_ANY_EXHDR(tunpacked));
+	const auto detoasted_value = DetoastPostgresDatum(reinterpret_cast<varlena *>(value), &should_free);
+	const auto datum_sv =  std::string_view((const char*)VARDATA_ANY(detoasted_value), VARSIZE_ANY_EXHDR(detoasted_value));
 	const auto val = duckdb::StringValue::Get(constant);
 	const auto val_sv =  std::string_view(val);
 	const bool res = OP::Operation(datum_sv, val_sv);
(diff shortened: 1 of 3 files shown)


src/pgduckdb_filter.cpp
```diff
@@ -17,41 +19,49 @@ TemplatedFilterOperation(Datum &value, const duckdb::Value &constant) {
...
 
+template <class OP>
+bool
+StringFilterOperation(Datum &value, const duckdb::Value &constant) {
+	const auto ptr = (text*)DatumGetPointer(value);
+	if (ptr == nullptr || constant.IsNull()) {
+		return false; // Comparison to NULL always returns false.
+	}
+
+	text* tunpacked = pg_detoast_datum_packed(ptr);
```
mkaruza, line 30: This is going to be problematic/buggy  with multiple threads. There is equivalent function in pgduckdb_detoast.cpp - `DetoastPostgresDatum` which should be safe in duckdb execution.

mkaruza (changes requested):

src/pgduckdb_filter.cpp
```diff
@@ -17,41 +19,49 @@ TemplatedFilterOperation(Datum &value, const duckdb::Value &constant) {
...
 		Datum date_datum = static_cast<int32_t>(value + pgduckdb::PGDUCKDB_DUCK_DATE_OFFSET);
 		return TemplatedFilterOperation<int32_t, OP>(date_datum, constant);
-		break;
 	}
 	case TIMESTAMPOID: {
 		Datum timestamp_datum = static_cast<int64_t>(value + pgduckdb::PGDUCKDB_DUCK_TIMESTAMP_OFFSET);
 		return TemplatedFilterOperation<int64_t, OP>(timestamp_datum, constant);
-		break;
 	}
+	case VARCHAROID:
```
JelteF, line 63: Let's also make this work with the `text` type. Which is much more commonly used in the postgres ecosystem than `varchar`. (should have a test too)

```suggestion
	case TEXTOID:
	case VARCHAROID:
```

src/pgduckdb_filter.cpp
Y--: Thanks for the review @mkaruza! I got the inspiration for this from the [text_to_cstring](https://github.com/postgres/postgres/blob/28a1121fd91235e09793c9ce64cb5539719afa30/src/backend/utils/adt/varlena.c#L220-L221) PG function, and I got the sense from @JelteF that we cannot use multiple thread in DuckDB execution when touching PG data anyway.

I was also hoping not to have to allocate + copy the PG string if possible.

Happy to use `DetoastPostgresDatum` though.

mkaruza: @Y-- we can use multiple threads for heap scan (there is also GUC variable where you can specify max threads that can be spawned for single table scan). FYI, joining two heap tables also runs in separate threads.

Y--: I see, make sense, thanks! And I assume there's no way to have a no-alloc & no-copy version of `DetoastPostgresDatum` in this context then?

mkaruza (approved):
lgtm

PR #142, opened 2024-08-21
Add support for using http to connect S3

lss602726449:
Currently, We always use HTTPS to connect to S3. We should also support HTTP to do so.

### DIFF
--- a/sql/pg_duckdb--0.0.1.sql
+++ b/sql/pg_duckdb--0.0.1.sql
@@ -117,6 +117,7 @@ CREATE TABLE secrets (
     region TEXT,
     endpoint TEXT,
     r2_account_id TEXT,
+    use_ssl BOOLEAN DEFAULT true,
     CONSTRAINT type_constraint CHECK (type IN ('S3', 'GCS', 'R2'))
 );
 
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -118,7 +118,7 @@ DuckdbCreateConnection(List *rtables, PlannerInfo *planner_info, List *needed_co
 			appendStringInfo(secret_key, ", ACCOUNT_ID '%s'", secret.endpoint.c_str());
 		}
 		if (!secret.use_ssl) {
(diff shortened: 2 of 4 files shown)


sql/pg_duckdb--0.0.1.sql
```diff
@@ -117,6 +117,7 @@ CREATE TABLE secrets (
     region TEXT,
     endpoint TEXT,
     r2_account_id TEXT,
+    use_ssl BOOLEAN,
```
JelteF, line 120: nitpick, seems nicer to set a default here:
```suggestion
    use_ssl BOOLEAN DEFAULT true,
```

lss602726449: Good idea, Thanks for your suggestion.

src/pgduckdb_duckdb.cpp
```diff
@@ -117,6 +117,9 @@ DuckdbCreateConnection(List *rtables, PlannerInfo *planner_info, List *needed_co
 		if (is_r2_cloud_secret) {
 			appendStringInfo(secret_key, ", ACCOUNT_ID '%s'", secret.endpoint.c_str());
 		}
+		if (!secret.use_ssl) {
+			appendStringInfo(secret_key, ", USE_SSL '%s'", "FALSE");
```
wuputah, line 121: No reason to format here:

```suggestion
			appendStringInfo(secret_key, ", USE_SSL 'FALSE'");
```

wuputah (commented):
I'm a little surprised DuckDB supports this. It seems very dangerous to me. What's the use case here? The default is `true` so that's good, and the logic is simple enough, I'm just worried about some future bug.

We should really get some kind of test coverage here, though I'm not sure how we want to do that if it's actually going to S3. Could we somehow test that the secret is created correctly in DuckDB?

lss602726449:
I just use a local minio without HTTPS. So I suppose to add this feature. This is support in duckdb in the readme https://github.com/duckdb/duckdb/blob/862852fa395b99735e5713cb55d0cea1d9320659/test/sql/copy/s3/README.md?plain=1#L46

wuputah (approved):

PR #143, opened 2024-08-21
Add pytest-based testing framework

JelteF:
`pgregress` is easy to use an is part of Postgres, so it's often the first thing to use testing a Postgres extension. But it also quite limiting, since you can only execute SQL and a test fails as soon as any of the output is different.

This adds a Python based testing framework to the project, which allows to do more "application" style testing. It's based on the testing framework I wrote for PgBouncer, but I removed a bunch of stuff that only made sense when testing PgBouncer. Postgres has a similar framework included, but that uses Perl and I really don't want to write any Perl. Also `pytest` is a much nicer to use testing framework than what is available in Perl.

As an example of a test that cannot easily be written using `pgregress` I added a test for `EXPLAIN ANALYZE`. Support for `EXPLAIN ANALYZE` was added in #136, but writing a test was not possible because the Timing information would be flaky.


mkaruza (approved):
I don't mind adding this for testing purposes - so approving. Only thing that i would like to understand why we need `ruff`.

JelteF:
Ruff is for auto formatting

Tishj:
I love this, `pytest` is great and really powerful for testing additional side-effects that would not be possible with the postgres regression tests

Tishj:
This could also be used to make #88 testable, verifying that files are correctly downloaded, multiple processes can make use of it and not break etc..

PR #144, opened 2024-08-21
Add a NOTICE file which contains the Postgres license

JelteF:
The PostgreSQL license is very liberal but, like MIT and BSD, it does
require attribution and requires that the original license is included.
This adds a NOTICE file containing the license, because we've vendored
in a few files/functions from the Postgres repository and we'll likely need
to vendor in more.

If/when we decide to to merge #143 we should also include PgBouncer its
license. Which is the ISC license, which is also attribution-only.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,31 @@
+NOTICES
+
+This repoisitory contains some source code from other projects. The original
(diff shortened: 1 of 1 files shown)


wuputah (approved):

wuputah:
good catch, thanks!

PR #170, opened 2024-09-11
Materialized view test with duckdb query execution

mkaruza:
* Create materialized view from read_* duckdb functions


JelteF (approved):
I think this indeed makes sense to add a test for. I merged #140, so after you rebase on master I think this can be merged.

PR #174, opened 2024-09-12
Make duckdb a singleton

JelteF:
Based on work by @Y-- and PR #102 and #147. This basically extracts the work from there into a separate PR, so it can be merged before some of the open TODOs related to prepared statements/extended protocol are resolved.

### DIFF
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -10,7 +10,30 @@ extern "C" {
 
 namespace pgduckdb {
 
-duckdb::unique_ptr<duckdb::DuckDB> DuckdbOpenDatabase();
+class DuckDBManager {
+public:
+	static inline const DuckDBManager &
+	Get() {
+		static DuckDBManager instance;
+		return instance;
+	}
+
+	inline duckdb::DuckDB &
+	GetDatabase() const {
+		return *database;
+	}
+
+private:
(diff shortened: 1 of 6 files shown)


mkaruza (approved):

include/pgduckdb/scan/postgres_scan.hpp
```diff
@@ -39,14 +39,14 @@ class PostgresScanLocalState {
 	bool m_exhausted_scan;
 };
 
-struct PostgresReplacementScanData : public duckdb::ReplacementScanData {
+struct PostgresReplacementScanDataClientContextState : public duckdb::ClientContextState {
```
Tishj, line 42: Can we shorten this to `PostgresReplacementState`?

Also for future reference, I think a good portion of these class variables are unused, such as `m_rtables`

include/pgduckdb/pgduckdb_duckdb.hpp
```diff
@@ -10,7 +10,30 @@ extern "C" {
...
+public:
+	static inline const DuckDBManager &
+	Get() {
+		static DuckDBManager instance;
+		return instance;
+	}
+
+	inline duckdb::DuckDB &
+	GetDatabase() const {
+		return *database;
```
Tishj, line 23: Can we do a nullcheck here? I imagine this just causes a generic looking InternalException if `database` is null which will be hard to debug

Issue #93, opened 2024-08-06
Harden against Postgres interactions with C++ destructors


JelteF:
While looking at the the C++ guidelines in the CONTRIBUTING.md file proposed in #90, I realized/remembered that Postgres C and regular C++ are not automatically the good friends that the project currently assumes they are.

Specifically if Postgres throws an `ERROR`, it will use `sigsetjmp` and `longjump` to traverse the the stack until the next `PG_TRY`. This is really problematic when considering the way C++ destructors work. Because the longjump will jump over these destructors. This can cause many different problems like memory leaks or locks acquired by `lock_guard` not being released.

The other way around is similarly bad. Any C++ exceptions that are not caught will skip over any cleanup that Postgres does in `PG_CATCH`. And thus probably killing the whole Postgres process tree.

This is all listed in the postgres docs as well: https://www.postgresql.org/docs/current/xfunc-c.html#EXTEND-CPP

The only way to fix these problems in a managable way I think, is by making the boundaries of calling C++ from C and C from C++ much clearer than they are now (possibly by never including both duckdb headers and postgres headers in the same files). And then at these boundaries convert C++ exceptions to Postgres errors and the other way around. 

P.S. This same problem applies to Rust and was worked around by `pgrx` by implementing panic -> pgerror and pgerror -> panic conversion at every FFI boundary: https://github.com/pgcentralfoundation/pgrx/blob/develop/docs/src/design-decisions.md#protecting-the-rust-stack-read-lying-to-postgres

wuputah:
`-bug` as this does not have a specific bug report to fix, but instead is an approach to improve overall error handling. That said, we have agreed that this is a good idea and we want to move towards having all interactions between C and C++ be in a single file similar to how you would use FFI -- but we won't try to catch memory allocation errors.

JelteF:
I'm going to move this issue to 0.2.0, so we can start stability testing for 0.1.0. This is an important issue, but we fixed the worst cases. We'll fix more in the next release.

PR #176, opened 2024-09-13
Handle errors in execution

mkaruza:
* Unify reporting of WARNING with log
* Throw exception in functions that are used inside duckdb execution
* Report WARNING for functions that are part of "normal" postgres execution
* Cleanup duckdb_state before reporting error in DuckDBNode (clean up)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,10 +128,11 @@ coding styles.
 ## Error Handling
 
 * Use exceptions **only** when an error is encountered that terminates a query (e.g. parser error, table not found). Exceptions should only be used for **exceptional** situations. For regular errors that do not break the execution flow (e.g. errors you **expect** might occur) use a return value instead.
-* *Exception* should be used in code that is executed inside duckdb engine.
-* Use PostgreSQL *elog* API to report messages back to user. Using *ERROR* is strictly forbiden to use as it can brake execution and lead to
-unexpected memory problems - *ERROR* is only used in duckdb node to early bail out execution in case of query preparation error or execution interruption.
-* Calling PostgreSQL native functions needs to be used with *PostgresFunctionGuard*. *PostgresFunctionGuard* will handle correctly *ERROR* log messages that could be emmited from these functions.
+* There are two distinct parts of the code where error handling is done very differently: The code that executes before we enter DuckDB execution engine (e.g. initial part of the planner hook) and the part that gets executed inside the duckdb execution engine. Below are rules for how to handle errors in both parts of the code. Not following these guidelines can cause crashes, memory leaks and other unexpected problems.
+* Before we enter the DuckDB exection engine no exceptions should ever be thrown here. In cases where you would want to throw an exception here, use `elog(ERROR, ...)`. Any C++ code that might throw an exception is also problematic. Since C++ throws exceptions on allocation failures, this covers lots of C++ APIs. So try to use Postgres datastructures instead of C++ ones whenever possible (e.g. use `List` instead of `Vec`)
+* Inside the duckdb execution engine the opposite is true. `elog(ERROR, ...)` should never be used there, use exceptions instead.
+* Use PostgreSQL *elog* API can be used to report non-fatal messages back to user. Using *ERROR* is strictly forbiden to use in code that is executed inside the duckdb engine.
+* Calling PostgreSQL native functions from within DuckDB execution needs **extreme care**. Pretty much non of these functions are thread-safe, and they might throw errors using `elog(ERROR, ...)`. If you've solved the thread-safety issue by taking a lock (or by carefully asserting that the actual code is thread safe), then you can use *PostgresFunctionGuard* to solve the `elog(ERROR, ...) problem. *PostgresFunctionGuard* will correctly handle *ERROR* log messages that could be emmited from these functions.
 * Try to add test cases that trigger exceptions. If an exception cannot be easily triggered using a test case then it should probably be an assertion. This is not always true (e.g. out of memory errors are exceptions, but are very hard to trigger).
 * Use `D_ASSERT` to assert. Use **assert** only when failing the assert means a programmer error. Assert should never be triggered by user input. Avoid code like `D_ASSERT(a > b + 3);` without comments or context.
 * Assert liberally, but make it clear with comments next to the assert what went wrong when the assert is triggered.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,6 +128,10 @@ coding styles.
 ## Error Handling
 
 * Use exceptions **only** when an error is encountered that terminates a query (e.g. parser error, table not found). Exceptions should only be used for **exceptional** situations. For regular errors that do not break the execution flow (e.g. errors you **expect** might occur) use a return value instead.
+* *Exception* should be used in code that is executed inside duckdb engine.
+* Use PostgreSQL *elog* API to report messages back to user. Using *ERROR* is strictly forbiden to use as it can brake execution and lead to
+unexpected memory problems - *ERROR* is only used in duckdb node to early bail out execution in case of query preparation error or execution interruption.
+* Calling PostgreSQL native functions needs to be used with *PostgresFunctionGuard*. *PostgresFunctionGuard* will handle correctly *ERROR* log messages that could be emmited from these functions.
 * Try to add test cases that trigger exceptions. If an exception cannot be easily triggered using a test case then it should probably be an assertion. This is not always true (e.g. out of memory errors are exceptions, but are very hard to trigger).
 * Use `D_ASSERT` to assert. Use **assert** only when failing the assert means a programmer error. Assert should never be triggered by user input. Avoid code like `D_ASSERT(a > b + 3);` without comments or context.
 * Assert liberally, but make it clear with comments next to the assert what went wrong when the assert is triggered.
--- a/include/pgduckdb/pgduckdb_utils.hpp
+++ b/include/pgduckdb/pgduckdb_utils.hpp
@@ -46,8 +46,8 @@ PostgresFunctionGuard(FuncType postgres_function, FuncArgs... args) {
 }
 
 template <typename FuncType, typename... FuncArgs>
-bool
-PostgresVoidFunctionGuard(FuncType postgres_function, FuncArgs... args) {
+std::optional<bool>
+PostgresFunctionGuard(FuncType postgres_function, FuncArgs... args) {
 	bool error = false;
 	// clang-format off
 	PG_TRY();
--- a/include/pgduckdb/pgduckdb_utils.hpp
+++ b/include/pgduckdb/pgduckdb_utils.hpp
@@ -4,11 +4,11 @@ extern "C" {
 #include "postgres.h"
 }
 
+#include "duckdb/common/exception.hpp"
+
 #include <vector>
 #include <string>
 #include <sstream>
-#include <cstdio>
-#include <optional>
 
 namespace pgduckdb {
 
@@ -23,11 +23,16 @@ TokenizeString(char *str, const char delimiter) {
 	return v;
 };
 
+/*
+ * DuckdbGlobalLock should be held before calling.
+ */
 template <typename T, typename FuncType, typename... FuncArgs>
-std::optional<T>
+T
 PostgresFunctionGuard(FuncType postgres_function, FuncArgs... args) {
 	T return_value;
 	bool error = false;
+	MemoryContext ctx = CurrentMemoryContext;
+	ErrorData *edata = nullptr;
 	// clang-format off
 	PG_TRY();
 	{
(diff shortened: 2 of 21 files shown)


mkaruza:
Should address #93

Tishj (approved):
Thanks for these efforts, I don't think #93 can be fully closed because it's something we have to be conscientiously aware of going forward.

Also I wonder if there are methods that are used both during duckdb execution and outside of it, which could be a problem if one or the other is assumed

wuputah:
I certainly don't want #93 to be open forever :joy: but we can maybe add something to CONTRIBUTING.md as a note as well. We should try to move towards issues being specific rather than generic "fix all the things" e.g. with either a specific reproducible test case or ideally pointing to something specific in the code. Of course this can be hard with issues like memory leaks.

If there's still specific things you think we should add or change, I'd suggest we open new issue(s) for those.

src/pgduckdb_detoast.cpp
```diff
@@ -78,20 +81,22 @@ ToastDecompressDatum(struct varlena *attr) {
...
+		throw duckdb::InvalidInputException("(PGDuckDB/ToastDecompressDatum) Invalid compression method id %d",
+		                                    TOAST_COMPRESS_METHOD(attr));
 		return NULL; /* keep compiler quiet */
 	}
 }
 
 static struct varlena *
 ToastFetchDatum(struct varlena *attr) {
-	Relation toastrel;
+	Relation toast_rel;
```
JelteF, line 92: sensible change, but lets do that in a separate PR (or commit) to keep history understandable.

include/pgduckdb/pgduckdb_utils.hpp
```diff
@@ -19,4 +23,44 @@ TokenizeString(char *str, const char delimiter) {
 	return v;
 };
 
+template <typename T, typename FuncType, typename... FuncArgs>
+std::optional<T>
+PostgresFunctionGuard(FuncType postgres_function, FuncArgs... args) {
+	T return_value;
+	bool error = false;
```
JelteF, line 30: Needs to be marked `volatile`

JelteF (changes requested):

CONTRIBUTING.md
```diff
@@ -128,6 +128,10 @@ coding styles.
 ## Error Handling
 
 * Use exceptions **only** when an error is encountered that terminates a query (e.g. parser error, table not found). Exceptions should only be used for **exceptional** situations. For regular errors that do not break the execution flow (e.g. errors you **expect** might occur) use a return value instead.
+* *Exception* should be used in code that is executed inside duckdb engine.
+* Use PostgreSQL *elog* API to report messages back to user. Using *ERROR* is strictly forbiden to use as it can brake execution and lead to
+unexpected memory problems - *ERROR* is only used in duckdb node to early bail out execution in case of query preparation error or execution interruption.
+* Calling PostgreSQL native functions needs to be used with *PostgresFunctionGuard*. *PostgresFunctionGuard* will handle correctly *ERROR* log messages that could be emmited from these functions.
```
JelteF, line 134:
```suggestion
* There are two distinct parts of the code where error handling is done very differently: The code that executes before we enter DuckDB execution engine (e.g. initial part of the planner hook) and the part that gets executed inside the duckdb execution engine. Below are rules for how to handle errors in both parts of the code. Not following these guidelines can cause crashes, memory leaks and other unexpected problems.
* Before we enter the DuckDB exection engine no exceptions should ever be thrown here. In cases where you would want to throw an exception here, use `elog(ERROR, ...)`. Any C++ code that might throw an exception is also problematic. Since C++ throws exceptions on allocation failures, this covers lots of C++ APIs. So try to use Postgres datastructures instead of C++ ones whenever possible (e.g. use `List` instead of `Vec`)
* Inside the duckdb execution engine the opposite is true. `elog(ERROR, ...)` should never be used there, use exceptions instead.
* Use PostgreSQL *elog* API can be used to report non-fatal messages back to user. Using *ERROR* is strictly forbiden to use in code that is executed inside the duckdb engine as 
* Calling PostgreSQL native functions from within DuckDB execution needs **extreme care**. Pretty much non of these functions are thread-safe, and they might throw errors using `elog(ERROR, ...)`. If you've solved the thread-safety issue by taking a lock (or by carefully asserting that the actual code is thread safe), then you can use *PostgresFunctionGuard* to solve the `elog(ERROR, ...) problem. *PostgresFunctionGuard* will correctly handle *ERROR* log messages that could be emmited from these functions.
```

include/pgduckdb/pgduckdb_utils.hpp
```diff
@@ -19,4 +23,44 @@ TokenizeString(char *str, const char delimiter) {
 	return v;
 };
 
+template <typename T, typename FuncType, typename... FuncArgs>
+std::optional<T>
+PostgresFunctionGuard(FuncType postgres_function, FuncArgs... args) {
```
JelteF, line 28: I think we should have a similar function for the other way around, i.e. where we call DuckDB C++ APIs from Postgres C code.

src/pgduckdb_detoast.cpp
```diff
@@ -106,15 +111,31 @@ ToastFetchDatum(struct varlena *attr) {
...
-	if (attrsize == 0)
+	if (attrsize == 0) {
 		return result;
+	}
 
 	DuckdbProcessLock::GetLock().lock();
-	toastrel = table_open(toast_pointer.va_toastrelid, AccessShareLock);
-	table_relation_fetch_toast_slice(toastrel, toast_pointer.va_valueid, attrsize, 0, attrsize, result);
-	table_close(toastrel, AccessShareLock);
+	toast_rel = try_table_open(toast_pointer.va_toastrelid, AccessShareLock);
```
JelteF, line 119: `try_table_open` might allocate memory an throw an `ERROR` that way. So this needs a `PostgresFunctionGuard` too.

src/pgduckdb_detoast.cpp
```diff
@@ -106,15 +111,31 @@ ToastFetchDatum(struct varlena *attr) {
...
+		throw duckdb::InternalException("(PGDuckDB/ToastFetchDatum) Error opening toast relation");
+	}
+
+	bool error_fetch_toast = false;
+	if (PostgresFunctionGuard(table_relation_fetch_toast_slice, toast_rel, toast_pointer.va_valueid, attrsize, 0,
+	                          attrsize, result).value()) {
+		error_fetch_toast = true;
+	}
+
+	table_close(toast_rel, AccessShareLock);
```
JelteF, line 132: Even `table_close` can throw some ERROR, see this quote from the PG source code where they consider memory context reset callbacks possibly throwing an error.

```
	/*
	 * It's not entirely clear whether 'tis better to do this before or after
	 * delinking the context; but an error in a callback will likely result in
	 * leaking the whole context (if it's not a root context) if we do it
	 * after, so let's do it before.
	 */
	MemoryContextCallResetCallbacks(context);
```

JelteF (changes requested):

JelteF:
> I certainly don't want #93 to be open forever 😂

Totally agreed. But I think there's two more things we need to do at minimum:
1. Confirm that #120 is fixed
2. Guarding (pretty much all) C++ API calls from postgres C code, because if those throw an exception (e.g. on OOM) they crash postgres.

I think neither necessarily needs to be part of this initial PR. I personally would prefer merging this PR as soon as we agree on nice to use Guard APIs. Because then we can use those Guard APIs for new code, without having to wait for the problems to be addressed in all existing code. 

Finally, this is an issue where it's easy to miss one place. So I definitely want to take another full pass myself over the codebase, after the initial push, to try and find some missed callsites before closing #93.

JelteF:
> Finally, this is an issue where it's easy to miss one place. 

As a side-note to this: It would be great if we could make it harder to make mistakes here. One idea I had was to not directly import duckdb C++ headers into postgres code. And just as well not directly import postgres headers into C++ code. But instead create dedicated files from where we create safe to use versions with the correct guards. That would make it easy to spot mistakes during review, because it's not expected that you import `postgres.h`/`duckdb.hpp` in certain files. And for the files where it is expected you take extra care during review.

include/pgduckdb/pgduckdb_utils.hpp
JelteF: Actually it seems this is probably not necessary. I was thinking about the problem described here, but that does not apply since `error` is not changed in the PG_TRY section: https://github.com/postgres/postgres/blob/master/src/include/utils/elog.h#L369-L374

include/pgduckdb/pgduckdb_utils.hpp
```diff
@@ -19,4 +23,44 @@ TokenizeString(char *str, const char delimiter) {
...
+	T return_value;
+	bool error = false;
+	// clang-format off
+	PG_TRY();
+	{
+		return_value = postgres_function(args...);
+	}
+	PG_CATCH();
+	{
+		error = true;
```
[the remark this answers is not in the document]
mkaruza: I'm not sure is it correct to throw from this. I would prefer that this function outputs error message but still caller needs to handle this.

CONTRIBUTING.md
mkaruza: Changed according to suggestion.

src/pgduckdb_detoast.cpp
mkaruza: Guarded.

include/pgduckdb/pgduckdb_utils.hpp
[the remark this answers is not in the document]
JelteF: what do you mean with "correct" in this case? 

Do you mean is it allowed to throw an exception in PG_CATCH? In which case, I think it's fine, but indeed I'm also not 100% sure. But if you throw an exception after the PG_CATCH if `error` is `true`, that should definitely be fine.  

Or do you mean is it what the caller should do? I think in at least 90% of the cases the caller will simply want to propagate the error message in the form of an exception. In the last 10% maybe they want more control, in which case returning the error message is indeed what we would want. But I think we should make it as easy as possible to write correct code, so I think we should also have a helper that simply throws an exception.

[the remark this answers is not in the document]
mkaruza: @JelteF  guard function will now throw InternalException with error message constructed from postgresql log

src/pgduckdb_utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+	{
+		MemoryContextSwitchTo(ctx);
+		edata = CopyErrorData();
+		FlushErrorState();
+		error = true;
+	}
+	PG_END_TRY();
+	// clang-format on
+	if (error) {
+		throw duckdb::InternalException("(PGDuckDB/PostgresFunctionGuard) %s", edata->message);
```
JelteF, line 33: I don't think it makes sense to add this prefix, the error didn't really originate here, we are simply transforming it, so having a reference to the guard won't help you find where it originated. Especially if we'll also add conversion the other way around I can see many prefixes being added to a simple error message.
```suggestion
		throw duckdb::InternalException("%s", edata->message);
```

src/pgduckdb_utils.cpp
```diff
@@ -0,0 +1,63 @@
...
```
JelteF, line 59: I think throwing an InternalException is acceptable for now, so that we can start using these guards in new code ASAP. But I think it would be great, if we would save not only the error message, but also the code/detail/context and maybe more in the `extra_data` of the `Exception`. Or probably easier: just store the whole `edata` (which we'd then probably have to allocate in `TopMemoryContext`).

That way we could convert such an `Exception` back to the equivalent `ereport` when surfacing the error to the user.

Tishj:
In the spirit of this PR, I am going to add some generic "ScopedPostgresResource" class so we can make sure never to leak postgres resources like `Relation` in the event of a C++ exception

src/pgduckdb_node.cpp
```diff
@@ -77,16 +77,16 @@ ExecuteQuery(DuckdbScanState *state) {
 			auto &executor = duckdb::Executor::Get(*connection->context);
 			// Wait for all tasks to terminate
 			executor.CancelTasks();
-
 			// Delete the scan state
 			CleanupDuckdbScanState(state);
 			// Process the interrupt on the Postgres side
 			ProcessInterrupts();
-			elog(ERROR, "Query cancelled");
+			elog(ERROR, "(PGDuckDB/ExecuteQuery) Query cancelled");
```
JelteF, line 84: I don't think the prefix really adds anything useful here.
```suggestion
			elog(ERROR, "Query cancelled");
```

Side-note: We should be throwing the expected error code here.

JelteF (changes requested):
Overall, I think this is pretty much in a merge-able state now. 

I'm not sure I'm a fan of all the `(PGDuckDB/ReplaceView)` type prefixes, but that might just be a matter of me not being used to them. I left two comments on where I think we shouldn't add them for sure though.

Apart from that: Tests are failing. Which should obviously be fixed before merging.

JelteF (approved):

mkaruza:
@JelteF will merge this PR to have some foundation to start with, but i would imagine that there is going to be changes in future until we are satisfied how to handle guarding / error logging / error handling in project.

PR #178, opened 2024-09-15
support postgresql 14

Leo-XM-Zeng:
Postgres 14
370f3892b310b4e608e27058b9a433f
make check
727051217ca72a6cca0e24506cf2f2c

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,8 @@ SRCS = src/scan/heap_reader.cpp \
 OBJS = $(subst .cpp,.o, $(SRCS))
 
 
-C_SRCS = src/vendor/pg_ruleutils_16.c \
+C_SRCS = src/vendor/pg_ruleutils_14.c \
+		 src/vendor/pg_ruleutils_16.c \
 		 src/vendor/pg_ruleutils_17.c
 OBJS += $(subst .c,.o, $(C_SRCS))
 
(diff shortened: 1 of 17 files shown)


Leo-XM-Zeng:
Hello, I can try to support postgresql 15 if needed

JelteF (changes requested):
Thanks a lot for your effort here. A few things before I think this can be merged:

1. I think merging this, requires adding PG15 support as well. Leaving a gap of the versions we support between PG14 and PG16 seems too confusing.
2. PG14 (and PG15) should be added to Github Actions testing
3. The original PG ruleutils should be added in a separate commit from the rest of the changes, just like I did in #140. That will make it easy to track where certain lines of code come from using `git blame`.

Leo-XM-Zeng:
> Thanks a lot for your effort here. A few things before I think this can be merged:
> 
> 1. I think merging this, requires adding PG15 support as well. Leaving a gap of the versions we support between PG14 and PG16 seems too confusing.
> 2. PG14 (and PG15) should be added to Github Actions testing
> 3. The original PG ruleutils should be added in a separate commit from the rest of the changes, just like I did in [Vendor in Postgres ruleutils and handle public.read_csv #140](https://github.com/duckdb/pg_duckdb/pull/140). That will make it easy to track where certain lines of code come from using `git blame`.

Ok, thank you for your guidance. Then I will finish PG15 first as required

Issue #154, opened 2024-08-28
[omnibus] platforms, PG versions, binaries, etc.


wuputah:
This is an omnibus recognition of the targets we don't have yet in order to consolidate these feature requests in one place and avoid duplication of very similar issues.

Operating Systems
* [x] Linux
* [x] Mac
* [ ] Windows

Postgres Versions
* :x: Postgres 12 - EOL is Nov 14, 2024 - Support is not planned
* :x: Postgres 13 - EOL is Nov 13, 2025 - Support is not planned
* [x] Postgres 14
* [x] Postgres 15
* [x] Postgres 16
* [x] Postgres 17  
* :x: Greenplum 6 - Support is not planned as Greenplum 6 is based on Postgres 9.4.

Platforms
* [x] x86_64
* [x] arm64 / aarch64

Binary package types
* [x] apt -> Ubuntu 22.04, Ubuntu 24.04, Debian bookworm, ...
  * supplied [via pgxman](http://pgx.sh/pg_duckdb), which supports Ubuntu Jammy, Ubuntu Noble, and Debian Bookworm. We may consider publishing packages directly in the future.
* [x] [pgxman](http://pgx.sh/pg_duckdb)
* [x] [Docker image](https://github.com/duckdb/pg_duckdb/blob/main/docker/README.md)
* [ ] homebrew
* [ ] rpm -> RHEL, RockyLinux, ...

Leo-XM-Zeng:
#178 Postgres 14

Leo-XM-Zeng:
#180 Postgres 15

JelteF:
As discussed in person. We won't support PG12 and PG13. And PG14 not for the 0.1.0 release, we might support PG14 for a later release. 

PG15 we'll support once #180 is merged for 0.1.0, but might remove support if it turns out to conflict with some of the future features we want to support.

wuputah:
all targets expected for 0.1.0 have been met; removing from milestone.

Vonng:
Currently, there are three DuckDB-related extensions: `duckdb_fdw`, `pg_mooncake`, and `pg_duckdb`, each using its own version of libduckdb. Consequently, it is not feasible to install these extensions simultaneously. Is there a possibility for `pg_duckdb` to utilize a standalone libduckdb shared library instead of building its own in the PostgreSQL library directory? Adopting a shared library could significantly reduce compilation times and package sizes, and also eliminate extension conflicts.

[...]

PR #180, opened 2024-09-16
support postgresql 15

Leo-XM-Zeng:


Leo-XM-Zeng:
image

**But there's a little problem I don't know how to deal with**
image
**It cannot be overwritten directly here for fear of affecting versions higher than PG15**
image

Leo-XM-Zeng:
'\getenv pwd PWD' is better transformed into '\set pwd' pwd ', which is more general and suitable for more PG versions. #178

Leo-XM-Zeng:
> Ragarding the problem, you need to apply the equivalent of this change to the ruleutils.c file too: [d62ba97](https://github.com/duckdb/pg_duckdb/commit/d62ba97b5fe6f1628493ae7aba087cf9621b0872)

That's one step missing. Thank you for your help
image

Leo-XM-Zeng:
@JelteF Hi JelteF, I have a question and I hope you can help me. I wonder if I need to move similar code to "pg_ruleutils_xx.c"? But it doesn't actually belong to "pg_ruleutils". 6025369b695bf743dec83773be3e2f5eb59776e1
image

Leo-XM-Zeng:
Or is there something that needs to be improved in this pull request and is looking forward to your suggestions

src/utility/copy.cpp
```diff
@@ -136,4 +165,4 @@ DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment
 	auto chunk = res->Fetch();
 	*processed = chunk->GetValue(0, 0).GetValue<uint64_t>();
 	return true;
-}
+}
```
JelteF, line 168: please install editorconfig and save this file again. We want a final newline in the file. https://editorconfig.org/

Leo-XM-Zeng: ok

JelteF:
Can you make sure this is also run in CI, by adding `REL_15_STABLE` here: https://github.com/duckdb/pg_duckdb/blob/ee3526d4f065f1f756caea80758653abfac1eda8/.github/workflows/build_and_test.yaml#L57

src/vendor/pg_ruleutils_15.c
```diff
@@ -1,20 +1,18 @@
 /*-------------------------------------------------------------------------
  *
- * ruleutils.c
+ * pg_ruleutils_15.c
  *	  Functions to convert stored expressions/querytrees back to
  *	  source text
  *
- * Portions Copyright (c) 1996-2022, PostgreSQL Global Development Group
+ * Portions Copyright (c) 1996-2023, PostgreSQL Global Development Group
```
JelteF, line 7: I think this date change was probably done by accident.

Leo-XM-Zeng: > I think this date change was probably done by accident. Can you revert that?

Ok, I'll be more careful next time

Leo-XM-Zeng:
> Can you make sure this is also run in CI, by adding here: `REL_15_STABLE`
> 
> https://github.com/duckdb/pg_duckdb/blob/ee3526d4f065f1f756caea80758653abfac1eda8/.github/workflows/build_and_test.yaml#L57

Okay, thanks. I'll try.

src/utility/copy.cpp
```diff
@@ -59,12 +70,25 @@ CreateRelationCopyParseState(ParseState *pstate, const CopyStmt *stmt, List **va
 		    lappend(*vars, makeVar(1, cur, tuple_desc->attrs[cur - 1].atttypid, tuple_desc->attrs[cur - 1].atttypmod,
 		                           tuple_desc->attrs[cur - 1].attcollation, 0));
 	}
+#else
+	foreach(cur, attnums)
+	{
+		int	attno = lfirst_int(cur) - FirstLowInvalidHeapAttributeNumber;
```
JelteF, line 76: Does this work? That should allow us to remove the `ListCell` declaration at the top of the function. If not we should update these macros to work for PG15.

```suggestion
	foreach_int(cur, attnums)
	{
		int	attno = cur - FirstLowInvalidHeapAttributeNumber;
```

Leo-XM-Zeng: This part refers to part of the source code of PG15
image

JelteF: Hmm, I didn't realize that we copied a part of this code initially. I might have kept the `foreach` macro then when I originally replaced it. But let's stay consistent here for now and use the new `foreach_int` macro in both places.

Leo-XM-Zeng: I have just read the relevant code and understood your meaning. Try to use the relevant definition of "pg_list.hpp" in the project, and I will modify it. 
```c
#define foreach_ptr(type, var, lst) foreach_internal(type, *, var, lst, lfirst)
#define foreach_int(var, lst)	foreach_internal(int, , var, lst, lfirst_int)
#define foreach_oid(var, lst)	foreach_internal(Oid, , var, lst, lfirst_oid)
#define foreach_xid(var, lst)	foreach_internal(TransactionId, , var, lst, lfirst_xid)

/*
 * The internal implementation of the above macros.  Do not use directly.
 *
 * This macro actually generates two loops in order to declare two variables of
 * different types.  The outer loop only iterates once, so we expect optimizing
 * compilers will unroll it, thereby optimizing it away.
 */
#define foreach_internal(type, pointer, var, lst, func) \
	for (type pointer var = 0, pointer var##__outerloop = (type pointer) 1; \
		 var##__outerloop; \
		 var##__outerloop = 0) \
		for (ForEachState var##__state = {(lst), 0}; \
			 (var##__state.l != NIL && \
			  var##__state.i < var##__state.l->length && \
			 (var = func(&var##__state.l->elements[var##__state.i]), true)); \
			 var##__state.i++)
```

Leo-XM-Zeng: Sorry, using part of PG code has become a habit, I did not understand your meaning quickly

src/scan/postgres_scan.cpp
```diff
@@ -201,7 +205,7 @@ PostgresReplacementScan(duckdb::ClientContext &context, duckdb::ReplacementScanI
 	}
 
 	/* SELECT query will have nodePath so we can return cardinality estimate of scan */
-	Cardinality nodeCardinality = node_path ? node_path->rows : 1;
```
JelteF, line 204: I don't think this is needed for PG15 support (probably this was for PG14 support). So let's remove it for now. Also it's the wrong type afaict. The Cardinality is a typedef for `double`.

src/utility/copy.cpp
```diff
@@ -108,7 +131,12 @@ DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment
 		raw_stmt->stmt_location = pstmt->stmt_location;
 		raw_stmt->stmt_len = pstmt->stmt_len;
 
-		rewritten = pg_analyze_and_rewrite_fixedparams(raw_stmt, query_string, NULL, 0, NULL);
+		#if PG_VERSION_NUM >= 150000
+			rewritten = pg_analyze_and_rewrite_fixedparams(raw_stmt, query_string, NULL, 0, NULL);
+		#else
+			rewritten = pg_analyze_and_rewrite(raw_stmt, query_string, NULL, 0, NULL);
+		#endif
```
JelteF, line 138: I guess this change is only needed for PG14 support. Let's remove it from this PR for now.

src/pgduckdb_hooks.cpp
```diff
@@ -52,7 +52,11 @@ ContainsDuckdbFunctions(Node *node, void *context) {
 
 	if (IsA(node, Query)) {
 		Query *query = (Query *)node;
-		return query_tree_walker(query, ContainsDuckdbFunctions, context, 0);
+#if PG_VERSION_NUM >= 160000
+	return query_tree_walker(query, ContainsDuckdbFunctions, context, 0);
```
JelteF, line 56: nitpick: This needs another indent (same for block below), to make it clear that it's part of the `if`

src/pgduckdb_hooks.cpp
```diff
@@ -62,12 +66,20 @@ ContainsDuckdbFunctions(Node *node, void *context) {
...
+#endif	
 }
 
 static bool
 NeedsDuckdbExecution(Query *query) {
+#if PG_VERSION_NUM >= 160000
 	return query_tree_walker(query, ContainsDuckdbFunctions, NULL, 0);
+#else
+	return query_tree_walker(query, (bool (*)()) ((void *) ContainsDuckdbFunctions), NULL, 0);
+#endif
```
JelteF, line 82: nitpick: this has some trailing whitespace space. Now that you have editorconfig installed, saving this file again should fix that automatically.
```suggestion
#endif
```

src/pgduckdb_hooks.cpp
```diff
@@ -52,7 +52,11 @@ ContainsDuckdbFunctions(Node *node, void *context) {
 
 	if (IsA(node, Query)) {
 		Query *query = (Query *)node;
-		return query_tree_walker(query, ContainsDuckdbFunctions, context, 0);
+#if PG_VERSION_NUM >= 160000
+	return query_tree_walker(query, ContainsDuckdbFunctions, context, 0);
+#else
+	return query_tree_walker(query, (bool (*)()) ((void *) ContainsDuckdbFunctions), context, 0);
```
JelteF, line 58: Side-note: We might want to create a helper function with a consistent calling convention around such functions, so that we don't need so many `#if PG_VERSION_NUM` statements throughout the codebase. I expect query_tree_walker could become one of those. For now I think this is fine though, since we only call it twice.

But to be clear, let's not do that in this PR.

JelteF (changes requested):
Some final suggestions, but after that it's good to merge IMO.

@wuputah @mkaruza do any of you have some feedback on this. If not I'll squash some of the commits after my final feedback is addressed, and then merge this in a way that preserves the history nicely (specifically the initial copy of the PG15 ruleutils).

src/utility/copy.cpp
Leo-XM-Zeng: Ok, I will submit this paragraph again at PG14

src/scan/postgres_scan.cpp
Leo-XM-Zeng: Ok, here the type definition basis function "CreateFunctionIndexScanArguments" parameter
image

src/pgduckdb_hooks.cpp
Leo-XM-Zeng: ok

src/pgduckdb_hooks.cpp
Leo-XM-Zeng: Sorry, I didn't think of a good implementation here, can you help me think of a more elegant implementation to replace this

src/scan/postgres_scan.cpp
JelteF: Hmm, I guess Postgres and DuckDB cardinalities have a different type (uint64 for Duckdb and double for Postgres). Let's leave it as is for now since this only applies to PG14, but let's add a comment explaining this difference on the PG14 PR.

src/pgduckdb_hooks.cpp
Leo-XM-Zeng: Thank you for your guidance.

wuputah:
> do any of you have some feedback on this ...

Go for it. Tests pass so 👍

JelteF:
@Leo-XM-Zeng sorry for the delay. We had some discussion in person about PG version support, given this is a young project and would like to move fast. The result of that discussion can be fonud here: https://github.com/duckdb/pg_duckdb/issues/154#issuecomment-2382890707

tl;dr we'd like to merge this PG15 support.

Could you fix the merge conflict? Then I'll merge it after.

JelteF (approved):

Leo-XM-Zeng:
> @Leo-XM-Zeng sorry for the delay. We had some discussion in person about PG version support, given this is a young project and would like to move fast. The result of that discussion can be fonud here: [#154 (comment)](https://github.com/duckdb/pg_duckdb/issues/154#issuecomment-2382890707)
> 
> tl;dr we'd like to merge this PG15 support.
> 
> Could you fix the merge conflict? Then I'll merge it after.
> 
> Thank you for all your work on this.

Ok, just a moment, please.

Leo-XM-Zeng:
@JelteF  oh, this error is due to unsupported syntax in the current psql15 version
image

JelteF:
to be clear, not remove the whole basic.sql just the lines that use `\bind`

Leo-XM-Zeng:
> to be clear, not remove the whole basic.sql just the lines that use `\bind`

understand

JelteF:
@wuputah or @Y-- Could one of you approve this. I've now cleaned up the commits, and I'll "merge" this by pushing straight to master to keep `git blame` working well (just like I did when previously vendoring in the PG16 and PG17 ruleutils files).

Y-- (approved):

PR #182, opened 2024-09-16
update duckdb to v1.1.0

wuputah:


Y-- (approved):

wuputah:
hmm I didn't expect the tests to fail! more work to do here...

bleskes:
++ to run on nightly. That said, ideally, we shouldn't go deeper than the basic CPP api, so I also didn't expect much breakage from the upgrade.

JelteF:
Based on the error in CI the problem seems to be that the EXPLAIN output has changed in a way that doesn't match the test assertion. i.e. the string "Timing:" is not present anymore in the `EXPLAIN ANALYZE` output.

JelteF (approved):

Y-- (approved):

JelteF:
FYI, I quickly fixed the testing issue and merged this because I was running into some bug that was fixed in DuckDB 1.1.0 while working on #147

Issue #184, opened 2024-09-17
Extension Installing Fails because of duckdb build git release version


ramarnat:
Installing Extensions using `SELECT duckdb.install_extension('azure');` fails.

```
WARNING:  (duckdb_install_extension) HTTP Error: Failed to download extension "azure" at URL "http://extensions.duckdb.org/17d598fc44/linux_amd64/azure.duckdb_extension.gz" (HTTP 403)
Extension "azure" is an existing extension.
```

The reason for this is the `17d598fc44`, which is the git release for duckdb. I have tried using `OVERRIDE_GIT_DESCRIBE=v0.1.1 GEN=ninja make` to have duckdb built with the right version, but that didnt work. 

Any recommendations on how to make sure the duckdb version is overridden?

wuputah:
try `OVERRIDE_GIT_DESCRIBE=v1.1.0` 😁

ramarnat:
ha. sorry that was a copy pasta, but it is `v1.1.0` 🤷🏽‍♂️

wuputah:
Hmm I see, yeah, I'm not sure how this works exactly. Something we will need to solve for functional builds / binaries. @Tishj do you know?

ramarnat:
I dont think that's the issue, because I am doing this in Docker with a specific version of duckdb, if I was using that git hash, the error would be: `at URL "http://extensions.duckdb.org/**fa5c2fe15f3d**/linux_amd64/azure.duckdb_extension.gz" (HTTP 403)`

wuputah:
Exactly, the URL is wrong. Try using my branch, or applying #185 if you can.

ramarnat:
I see that the azure extension is available at 
```
http://extensions.duckdb.org/fa5c2fe15f/linux_amd64/azure.duckdb_extension.gz
```

but its also available at 
```
http://extensions.duckdb.org/v1.1.0/linux_amd64/azure.duckdb_extension.gz
```

And I'd like to be able to checkout a named tag and have that be used in the duckdb call....

wuputah:
Right, those two commits are the same. 👍 

But otherwise, yeah I get the point. I'm not sure how to accomplish that, that's a "building duckdb" question I don't have an answer to, but it's not specific to pg_duckdb.

wuputah:
[...]

PR #185, opened 2024-09-17
remove hash in .gitmodules

wuputah:
~completes the update made in ee3526d, possibly addresses #184~

removes the hash from .gitmodules as it has no effect

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "third_party/duckdb"]
 	path = third_party/duckdb
 	url = https://github.com/duckdb/duckdb.git
-	hash = 17d598fc4472c64969f47f86c30fce75c4d64ed4


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "third_party/duckdb"]
 	path = third_party/duckdb
 	url = https://github.com/duckdb/duckdb.git
-	hash = 17d598fc4472c64969f47f86c30fce75c4d64ed4
+	hash = fa5c2fe15f3da5f32397b009196c0895fce60820
```
Y--, line 4: I didn't know `hash` property was a thing? And I don't see it in the [doc](https://git-scm.com/docs/gitmodules/en) either? Does that do anything?

wuputah: Oh my, it seems it may not, though it _seems_ to have some effect, that could be some other problem. Looking at the history, Thijs added this when we pinned circa 1.0.

I do really dislike git submodules. We could just put the duckdb directory into gitignore and have this all handled by the Makefile instead... food for thought.

wuputah: ok I just confirmed that on a clean checkout then `git submodule update --init --recursive` this appears to do nothing.

wuputah: changed PR to remove this instead

Y-- (approved):

PR #188, opened 2024-09-18
Add `ScopedPostgresResource`

Tishj:
This PR feels like a continuation of #176 and is also related to #93 

When we throw a C++ exception we have to make sure that all postgres related resources get released, this is done through the destructor of the ScopedPostgresResource class

Opening as a draft, hopefully we can talk about the implementation or even alternatives that solve the same issue


JelteF:
I think this goal of this PR makes a lot of sense, but I think this needs a review from someone that's more experienced in C++ than me. @Y-- what do you think about this? Are there easier/better ways to achieve the same?

Y--:
> I think this goal of this PR makes a lot of sense, but I think this needs a review from someone that's more experienced in C++ than me. @Y-- what do you think about this? Are there easier/better ways to achieve the same?

This LGTM, a pretty typical use of RAII pattern :-)

Tishj:
Thanks, then I'll add it to more places as required and undraft 👍

JelteF:
Moved this to 0.2.0, in that release we'll do a more holistic approach to solving these issues.

JelteF:
@Y-- Does this still make sense? Or should we close this one? It seems like you haven't needed this for your cleanup so far?

PR #192, opened 2024-09-19
Fix for pg_duckdb cache invalidation

### LINKED ISSUE #187: Server crash related to `duckdb.execution` setting
Reproduction

```
SET duckdb.execution TO true;
drop extension pg_duckdb;
create extension pg_duckdb;
SET duckdb.execution TO true;
```

Results in:
```
!?> 
```

mkaruza:
* List `duckdb_only_functions` will be cleaned nly if cache is considered installed (i.e. list should be populated at this point).
* Should fix #187


test/regression/sql/basic.sql
```diff
@@ -20,3 +20,8 @@ SET client_min_messages TO default;
 
 DROP TABLE t;
 DROP TABLE empty;
+
+-- Check that DROP / CREATE extension works
+
```
Tishj, line 25: I don't quite see how this is testing the problem, as it occurred from running `create extension pg_duckdb;` without having loaded the `pg_duckdb` extension, but never the less the changes fixed it 👍

Tishj (approved):

src/pgduckdb_metadata_cache.cpp
```diff
@@ -62,7 +62,10 @@ InvalidateCaches(Datum arg, int cache_id, uint32 hash_value) {
 		return;
 	}
 	cache.valid = false;
-	list_free(cache.duckdb_only_functions);
+	if (cache.installed) {
+		list_free(cache.duckdb_only_functions);
+		cache.duckdb_only_functions = nullptr;
```
JelteF, line 67: nit: for consistency with Postgres code it would be better to set this to NIL (which is defined as `(List *) NULL`)
```suggestion
		cache.duckdb_only_functions = NIL;
```

JelteF (approved):
LGTM

JelteF (approved):

PR #193, opened 2024-09-19
Enum type support

Tishj:
This PR adds support for Postgres's ENUM types.

In Postgres ENUM types can not be ephemeral (exist only for the duration of a query), they *have* to be registered, and thus we can't *just* recreate the ENUM type when converting back from DuckDB -> Postgres at the end of execution.

This detail caused the biggest development overhead, we need to vendor and inherit from a templated derived class that contains the guts of the DuckDB ENUM type so we can add Postgres's ENUM member oids to it.

```c++
template <class T>
class PGDuckDBEnumTypeInfo : public EnumTypeInfoTemplated<T> {
public:
	PGDuckDBEnumTypeInfo(Vector &values_insert_order_p, idx_t dict_size_p, Vector &oid_vec)
	    : EnumTypeInfoTemplated<T>(values_insert_order_p, dict_size_p), oid_vec(oid_vec) {
	}

public:
	const Vector &
	GetMemberOids() const {
		return oid_vec;
	}

private:
	Vector oid_vec;
};
```

We save the ENUM member oids into this, as this is what has to be returned in the query result.


src/pgduckdb_types.cpp
```diff
@@ -702,6 +850,31 @@ ConvertDecimal(const NumericVar &numeric) {
 	return (NumericIsNegative(numeric) ? -base_res : base_res);
 }
 
+int
+GetEnumPosition(Datum enum_oid) {
```
JelteF, line 854: I think for clarity it would be good to not use the name `enum_oid` at all in this PR, but use `enum_type_oid` and `enum_member_oid`. Right now you use `enum_oid` for both which is a bit confusing.

src/pgduckdb_types.cpp
```diff
@@ -546,12 +659,33 @@ ConvertPostgresToDuckColumnType(Form_pg_attribute &attribute) {
...
+			return ConvertPostgresEnumToDuckEnum(type);
+		}
 		std::string name = "UnsupportedPostgresType (Oid=" + std::to_string(type) + ")";
 		return duckdb::LogicalType::USER(name);
 	}
 	}
 }
 
+Oid
+GetEnumTypeOid(const Vector &oids) {
```
JelteF, line 672: Afaict we can remove this whole function if we store the oid of the enum type also in the `PGDuckDBEnumTypeInfo` when we create it. Not needing code is always good, especially when the code causes us to cross the Postgres <-> DuckDB boundary.

JelteF (changes requested):

include/pgduckdb/duckdb_vendor/enum_type_info_templated.hpp
```diff
@@ -0,0 +1,54 @@
+#pragma once
```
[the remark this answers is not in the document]
Tishj: Yea that's entirely possible, I just wanted to do the changes that are required to get it working first.
I'll PR it upstream so it's not necessary

I believe we want this to get into the 0.1 release though, so I think the vendoring will be needed for that

include/pgduckdb/types/pgduckdb_enum.hpp
```diff
@@ -0,0 +1,59 @@
...
+	const Vector &
+	GetMemberOids() const {
+		return oid_vec;
+	}
+
+private:
+	Vector oid_vec;
+};
+
+struct PGDuckDBEnum {
```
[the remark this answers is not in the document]
Tishj: Sure I'll add the comments 👍 

I can't extend `EnumTypeInfo` without subclassing, which I'm doing with `PGDuckDBEnumTypeInfo`, but the internals of `EnumTypeInfo` are assuming (in things like `EnumTypeInfo::GetPos`) that the derived class is `EnumTypeInfoTemplated` and it will cast to that.

Since `PGDuckDBEnumTypeInfo` is a templated class I can't add a static method there[1]
So the next best option is to add `PGDuckDBEnum` just to contain static methods (probably should move the enum-related methods from `pgduckdb_types.cpp` into this as well for cohesion)


[1] I can do this, but then it has to be called like `PGDuckDBEnumTypeInfo<uint8_t>::CreateType(...)` even if we don't know what type we're creating yet.
Which either looks confusing or we would need to move that if/else tree outside of the function, which defeats the purpose of having this static function.

src/pgduckdb_types.cpp
Tishj: Yea agreed 👍

include/pgduckdb/duckdb_vendor/enum_type_info_templated.hpp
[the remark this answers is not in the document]
JelteF: Makes sense. I'm not at all opposed to temporarily vendoring it in until the next DuckDB release. But less vendored code is always good, so I think it's good to get support for this usage in upstream. Also, I think any vendored code should have a comment explaining why we need it, and if there's something we're waiting for until we can remove it again. (I realize I didn't actually do that myself for the pg_ruleutils_xx.c files, so I'll update those comments soonish)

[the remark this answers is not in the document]
Tishj: Alright that makes sense, I'll make the changes 👍 

https://github.com/duckdb/duckdb/pull/14038

include/pgduckdb/types/pgduckdb_enum.hpp
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+public:
+	const Vector &
+	GetMemberOids() const {
+		return oid_vec;
+	}
+
+private:
+	Vector oid_vec;
```
Mytherin, line 32: Instead of storing these oids - can't we do a reverse lookup of the oid based on the string name when converting back? That way we don't need to modify the DuckDB enum type.

JelteF: I don't think there's really a way around creating a custom DuckDB enum type. I agree in theory we could get away with not storing these OIDs, but to be able to look them up we'd need to know the oid (or name) of the enum type itself. The only way I can think of to find that OID is by storing it in the type DuckDB enum type again. Which brings us back to the point where we need to modify the DuckDB enum type, and if we have to do that we might as well store the oid_vec in there for efficience too.

To clarify postgres behaviour: enums even if they have the exact same members, are not the same. And the members have different oids based on the specific enum they are part of: 
```
> create type abc as enum ('a', 'b', 'c');
> create type abc2 as enum ('a', 'b', 'c');
> select 'a'::abc = 'a'::abc2;
ERROR:  42883: operator does not exist: abc = abc2
LINE 1: select 'a'::abc = 'a'::abc2;
> select * from pg_enum;
  oid   │ enumtypid │ enumsortorder │ enumlabel
────────┼───────────┼───────────────┼───────────
 101258 │    101256 │             1 │ a
 101260 │    101256 │             2 │ b
 101262 │    101256 │             3 │ c
 101266 │    101264 │             1 │ a
 101268 │    101264 │             2 │ b
 101270 │    101264 │             3 │ c
```

src/pgduckdb_types.cpp
```diff
@@ -702,6 +850,31 @@ ConvertDecimal(const NumericVar &numeric) {
...
+
+	if (!HeapTupleIsValid(enum_tuple)) {
+		elog(ERROR, "could not find enum with OID %u", DatumGetObjectId(enum_oid));
+	}
+
+	// Get the pg_enum structure from the tuple
+	enum_entry = (Form_pg_enum)GETSTRUCT(enum_tuple);
+
+	// The enumsortorder field gives us the position of the enum value
+	enum_position = (int)enum_entry->enumsortorder;
```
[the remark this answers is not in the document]
JelteF: Actually, I misremembered the performance part of this comment. SearchSysCache is relatively efficient because it does not allocate, it's SearchSysCacheCopy that is particularly problematic performance wise.

include/pgduckdb/types/pgduckdb_enum.hpp
Tishj: One way we might be able to get away with this is by storing the extra data in the `Vector values_insert_order;`
Vector's don't have a size, that's why the EnumTypeInfo also has a accompanying `idx_t dict_size;`, I could allocate extra data in the Vector to store this data, if the `dict_size` stays the same it should never be looked at

(This will probably fail in debug mode because of (de)serialization roundtrip verification)

src/pgduckdb_types.cpp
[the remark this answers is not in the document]
JelteF: > I think it makes more sense to cache a mapping of postgres enum_oid to duckdb position in your custom enum type, so that we can get the position of an enum_oid without having to access postgres catalogs.

Instead of doing that, we could also simply iterate over the already existing `enum_member_oids` vec. Since the number of members is usually relatively small I think this will be fine performance wise, at least for a first version.

include/pgduckdb/types/pgduckdb_enum.hpp
Tishj:
```patch
diff --git a/include/pgduckdb/types/pgduckdb_enum.hpp b/include/pgduckdb/types/pgduckdb_enum.hpp
index 7cc7e1c..f141b70 100644
--- a/include/pgduckdb/types/pgduckdb_enum.hpp
+++ b/include/pgduckdb/types/pgduckdb_enum.hpp
@@ -46,12 +46,12 @@ private:
 };
 
 struct PGDuckDBEnum {
-	static LogicalType CreateEnumType(Vector &ordered_data, idx_t size, Vector &enum_member_oids);
+	//static LogicalType CreateEnumType(Vector &ordered_data, idx_t size, Vector &enum_member_oids);
 	static idx_t GetDuckDBEnumPosition(duckdb::Value &val);
 	static int GetEnumPosition(Datum enum_member_oid);
 	static bool IsEnumType(Oid type_oid);
 	static Oid GetEnumTypeOid(const Vector &oids);
-	static const Vector &GetMemberOids(const duckdb::LogicalType &type);
+	static Vector GetMemberOids(const duckdb::LogicalType &type);
 };
 
 } // namespace pgduckdb
diff --git a/src/pgduckdb_types.cpp b/src/pgduckdb_types.cpp
index e7609ac..e73f08e 100644
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -327,7 +327,7 @@ ConvertDuckToPostgresArray(TupleTableSlot *slot, duckdb::Value &value, idx_t col
 void
 ConvertDuckToPostgresEnumValue(TupleTableSlot *slot, duckdb::Value &val, idx_t col) {
 	auto position = PGDuckDBEnum::GetDuckDBEnumPosition(val);
-	auto &enum_oids = PGDuckDBEnum::GetMemberOids(val.type());
+	auto enum_oids = PGDuckDBEnum::GetMemberOids(val.type());
 	auto enum_value_oid = duckdb::FlatVector::GetData<Oid>(enum_oids)[position];
 	slot->tts_values[col] = ObjectIdGetDatum(enum_value_oid);
 }
@@ -544,10 +544,13 @@ ConvertPostgresEnumToDuckEnum(Oid enum_type_oid) {
 
 	std::sort(enum_members.begin(), enum_members.end(), sort_order_cmp);
 
-	auto duck_enum_vec = duckdb::Vector(duckdb::LogicalType::VARCHAR, enum_members.size());
-	auto enum_oid_vec = duckdb::Vector(duckdb::LogicalType::UINTEGER, enum_members.size());
+	idx_t allocation_size = enum_members.size();
+	allocation_size += enum_members.size() / 4;
+	allocation_size += (enum_members.size() % 4) != 0;
+
+	auto duck_enum_vec = duckdb::Vector(duckdb::LogicalType::VARCHAR, allocation_size);
 	auto enum_vec_data = duckdb::FlatVector::GetData<duckdb::string_t>(duck_enum_vec);
-	auto enum_oid_data = duckdb::FlatVector::GetData<uint32_t>(enum_oid_vec);
+	auto enum_oid_data = (uint32_t*)(enum_vec_data + enum_members.size());
 	for (idx_t i = 0; i < enum_members.size(); i++) {
 		auto &member = enum_members[i];
 		auto enum_data = (Form_pg_enum)GETSTRUCT(member);
@@ -556,7 +559,7 @@ ConvertPostgresEnumToDuckEnum(Oid enum_type_oid) {
 	}
 
 	PostgresFunctionGuard(ReleaseCatCacheList, list);
-	return PGDuckDBEnum::CreateEnumType(duck_enum_vec, enum_members.size(), enum_oid_vec);
+	return EnumTypeInfo::CreateType(duck_enum_vec, enum_members.size());
 }
 
 duckdb::LogicalType
@@ -683,7 +686,7 @@ GetPostgresDuckDBType(duckdb::LogicalType type) {
 		}
 	}
 	case duckdb::LogicalTypeId::ENUM: {
-		auto &member_oids = PGDuckDBEnum::GetMemberOids(type);
+		auto member_oids = PGDuckDBEnum::GetMemberOids(type);
 		return PGDuckDBEnum::GetEnumTypeOid(member_oids);
 	}
 	default: {
diff --git a/src/types/pgduckdb_enum.cpp b/src/types/pgduckdb_enum.cpp
index 9b85ef5..a98aa66 100644
--- a/src/types/pgduckdb_enum.cpp
+++ b/src/types/pgduckdb_enum.cpp
@@ -5,27 +5,27 @@ namespace pgduckdb {
 
 // This is copied from 'duckdb::EnumTypeInfo::CreateType' with our additional metadata added
 
-LogicalType
-PGDuckDBEnum::CreateEnumType(Vector &ordered_data, idx_t size, Vector &enum_member_oids) {
-	// Generate EnumTypeInfo
-	shared_ptr<ExtraTypeInfo> info;
-	auto enum_internal_type = EnumTypeInfo::DictType(size);
-	switch (enum_internal_type) {
-	case PhysicalType::UINT8:
-		info = make_shared_ptr<PGDuckDBEnumTypeInfo<uint8_t>>(ordered_data, size, enum_member_oids);
-		break;
-	case PhysicalType::UINT16:
-		info = make_shared_ptr<PGDuckDBEnumTypeInfo<uint16_t>>(ordered_data, size, enum_member_oids);
-		break;
-	case PhysicalType::UINT32:
-		info = make_shared_ptr<PGDuckDBEnumTypeInfo<uint32_t>>(ordered_data, size, enum_member_oids);
-		break;
-	default:
-		throw InternalException("Invalid Physical Type for ENUMs");
-	}
-	// Generate Actual Enum Type
-	return LogicalType(LogicalTypeId::ENUM, info);
-}
+//LogicalType
+//PGDuckDBEnum::CreateEnumType(Vector &ordered_data, idx_t size, Vector &enum_member_oids) {
+//	// Generate EnumTypeInfo
+//	shared_ptr<ExtraTypeInfo> info;
+//	auto enum_internal_type = EnumTypeInfo::DictType(size);
+//	switch (enum_internal_type) {
+//	case PhysicalType::UINT8:
+//		info = make_shared_ptr<PGDuckDBEnumTypeInfo<uint8_t>>(ordered_data, size, enum_member_oids);
+//		break;
+//	case PhysicalType::UINT16:
+//		info = make_shared_ptr<PGDuckDBEnumTypeInfo<uint16_t>>(ordered_data, size, enum_member_oids);
+//		break;
+//	case PhysicalType::UINT32:
+//		info = make_shared_ptr<PGDuckDBEnumTypeInfo<uint32_t>>(ordered_data, size, enum_member_oids);
+//		break;
+//	default:
+//		throw InternalException("Invalid Physical Type for ENUMs");
+//	}
+//	// Generate Actual Enum Type
+//	return LogicalType(LogicalTypeId::ENUM, info);
+//}
 
 idx_t
 PGDuckDBEnum::GetDuckDBEnumPosition(duckdb::Value &val) {
@@ -103,20 +103,14 @@ PGDuckDBEnum::GetEnumTypeOid(const Vector &oids) {
 	return result;
 }
 
-const Vector &
-PGDuckDBEnum::GetMemberOids(const duckdb::LogicalType &type) {
+Vector PGDuckDBEnum::GetMemberOids(const duckdb::LogicalType &type) {
 	auto type_info = type.AuxInfo();
 	auto &enum_type_info = type_info->Cast<duckdb::EnumTypeInfo>();
-	switch (enum_type_info.DictType(enum_type_info.GetDictSize())) {
-	case PhysicalType::UINT8:
-		return enum_type_info.Cast<PGDuckDBEnumTypeInfo<uint8_t>>().GetMemberOids();
-	case PhysicalType::UINT16:
-		return enum_type_info.Cast<PGDuckDBEnumTypeInfo<uint16_t>>().GetMemberOids();
-	case PhysicalType::UINT32:
-		return enum_type_info.Cast<PGDuckDBEnumTypeInfo<uint32_t>>().GetMemberOids();
-	default:
-		throw InternalException("Invalid Physical Type for ENUMs");
-	}
+	auto &enum_sort_order = enum_type_info.GetValuesInsertOrder();
+	idx_t dict_size = enum_type_info.GetDictSize();
+
+	auto vector_data = duckdb::FlatVector::GetData<duckdb::string_t>(enum_sort_order);
+	return Vector(duckdb::LogicalType::UINTEGER, (duckdb::data_ptr_t)(vector_data + dict_size));
 }
 
 } // namespace pgduckdb
diff --git a/third_party/duckdb b/third_party/duckdb
--- a/third_party/duckdb
+++ b/third_party/duckdb
@@ -1 +1 @@
-Subproject commit fa5c2fe15f3da5f32397b009196c0895fce60820
+Subproject commit fa5c2fe15f3da5f32397b009196c0895fce60820-dirty
```

This works 🤢

Tishj: And it's pushed

src/pgduckdb_types.cpp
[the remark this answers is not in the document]
Tishj: I made the change 👍 
```c++
idx_t
PGDuckDBEnum::GetEnumPosition(Datum enum_member_oid_p, const duckdb::LogicalType &type) {
	auto enum_member_oid = DatumGetObjectId(enum_member_oid_p);

	auto &enum_type_info = type.AuxInfo()->Cast<duckdb::EnumTypeInfo>();
	auto dict_size = enum_type_info.GetDictSize();

	auto enum_member_oids = PGDuckDBEnum::GetMemberOids(type);
	auto oids_data = duckdb::FlatVector::GetData<uint32_t>(enum_member_oids);

	uint32_t *begin = oids_data;
	uint32_t *end = oids_data + dict_size;
	uint32_t *result = std::find(begin, end, enum_member_oid);

	if (result == end) {
		throw duckdb::InternalException("Could not find enum_member_oid: %d", enum_member_oid);
	}
	return (idx_t)(result - begin);
}
```

src/pgduckdb_types.cpp
```diff
@@ -495,6 +519,53 @@ ChildTypeFromArray(Oid array_type) {
...
+		} else {
+			return false; // v1 == v2
+		}
+	};
+
+	std::sort(enum_members.begin(), enum_members.end(), sort_order_cmp);
+
+	idx_t allocation_size = enum_members.size();
+	allocation_size += enum_members.size() / 4;
+	allocation_size += (enum_members.size() % 4) != 0;
```
JelteF: This definitely needs a code comment to explain what's going on. Afaict you are allocating enough additional space some extra space after. And I'm guessing the 4 is for the size of the `Oid`? If so let's make that clear:
```suggestion
	allocation_size += enum_members.size() / sizeof(Oid);
	allocation_size += (enum_members.size() % sizeof(Oid)) != 0;
```

Does this take care of allignment?

include/pgduckdb/types/pgduckdb_enum.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "access/htup_details.h"
+}
+
+namespace pgduckdb {
+
+using duckdb::idx_t;
+using duckdb::LogicalType;
+using duckdb::Vector;
+
+struct PGDuckDBEnum {
```
JelteF, line 20: Should this be changed to a namespace? Now that this struct doesn't have any fields anymore?

include/pgduckdb/types/pgduckdb_enum.hpp
Tishj: Yea I talked about it with Mark, there isn't a good solution and indeed all of them have a certain ick to them

include/pgduckdb/types/pgduckdb_enum.hpp
Tishj: Yea possibly, in our codebase we're not too strict on static-membered-structs vs namespaces, I think when we also need a constexpr value here we can't use a namespace but can use a struct, so I almost always go for using a struct

src/pgduckdb_types.cpp
Tishj: I'm not sure, but `string_t` is 16 bytes so I assume the answer is no since it's divisable by 8 (`sizeof(void *)`) ?

Tishj:
I'm waiting on <https://github.com/duckdb/pg_duckdb/pull/97> so I can add the tests with `a > 'sad'` etc, that will involve adding support for looking up type catalog entries, which is probably a nice bonus (but might be better as a separate PR)

include/pgduckdb/types/pgduckdb_enum.hpp
Tishj: Hmm now trying to make `select a > 'sad' from tbl;` work, which involves storing the LogicalType in the TypeCatalogEntry, this invokes `ExtraTypeInfo::Copy` and breaks this "solution"

```c++
shared_ptr<ExtraTypeInfo> EnumTypeInfo::Copy() const {
	Vector values_insert_order_copy(LogicalType::VARCHAR, false, false, 0);
	values_insert_order_copy.Reference(values_insert_order);
	return make_shared_ptr<EnumTypeInfo>(values_insert_order_copy, dict_size);
}
```

Tishj: The actual issue was this:
```patch
diff --git a/src/pgduckdb_filter.cpp b/src/pgduckdb_filter.cpp
index 081271d..11c9bb2 100644
--- a/src/pgduckdb_filter.cpp
+++ b/src/pgduckdb_filter.cpp
@@ -71,8 +71,7 @@ FilterOperationSwitch(Datum &value, duckdb::Value &constant, Oid type_oid) {
 	case VARCHAROID:
 		return StringFilterOperation<OP>(value, constant);
 	default:
-		throw duckdb::InvalidTypeException(
-		    duckdb::string("(DuckDB/FilterOperationSwitch) Unsupported duckdb type: %d", type_oid));
+		throw duckdb::InvalidTypeException(duckdb::StringUtil::Format("(DuckDB/FilterOperationSwitch) Unsupported duckdb type: %d", type_oid));
 	}
 }
```

Tishj: This invoked this `std::string` constructor:
```
basic_string( const basic_string& other, size_type pos,
              const Allocator& alloc = Allocator() );
```

Which flooded my output with *every* string literal that duckdb contains 🙃

wuputah:
Regarding the `0.2.0` milestone: while it's not a blocker for 0.1.0 but it would be great for this to land, especially considering it's basically done. Use of ENUMs is certainly not uncommon in Postgres.

src/types/pgduckdb_enum.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+using duckdb::idx_t;
+using duckdb::LogicalTypeId;
+using duckdb::PhysicalType;
+
+// This is medium hacky, we create extra space in the Vector that would normally only hold the strings corresponding to
+// the enum's members. In that extra space we store the enum member oids.
+
+// We do this so we can find the OID from the offset (which is what DuckDB stores in a Vector of type ENUM) and return
+// that when converting the result from DuckDB -> Postgres
```
JelteF, line 15: This is not true anymore right? i.e. you switched back to the first with the vendored in template (until the next duckdb release)

include/pgduckdb/duckdb_vendor/enum_type_info_templated.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "duckdb/common/extra_type_info.hpp"
+#include "duckdb/common/vector.hpp"
+#include "duckdb/common/types/vector.hpp"
+#include "duckdb/common/string_map_set.hpp"
+#include "duckdb/common/serializer/deserializer.hpp"
+
+// This is copied directly without changes from 'duckdb/src/common/extra_type_info.cpp'
+// The reason this is copied is to be able to inherit from it.
```
JelteF, line 10:
```suggestion
// The reason this is copied is to be able to inherit from it.
// This can be removed once DuckDB 1.2.0 is released, which contains this PR: 
// https://github.com/duckdb/duckdb/pull/14038
```

src/types/pgduckdb_enum.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+LogicalType
+PGDuckDBEnum::CreateEnumType(std::vector<HeapTuple> &enum_members) {
```
JelteF, line 18: I think it makes sense to merge this function `ConvertPostgresEnumToDuckEnum`. They are so strongly coupled that I don't think it has any benefit to separate them like this. i.e. `ConvertPostgresEnumToDuckEnum` is the only caller of this function and that this takes a vector of HeapTuples, which should already be sorted. If we'd keep them separate that requirement on already sorted should be documented in this function its comment, but if we merge the functions such a comment is not necessary.

PR #194, opened 2024-09-19
Explicitly convert datum values before filter operation

### LINKED ISSUE #189: Comparisons with double precision type return wrong results
Something is very wrong here (found while creating tests for #147)

```sql
create table t(d double precision);
insert into t values (8.0);
select count(*) from t where d < 8.1; -- returns 1 (correct)
set duckdb.execution = true;
select count(*) from t where d < 8.1; -- returns 0 (wrong)
```

mkaruza:
Datums can contain a pass-by-value or a pass-by-reference type. We were simply casting the `Datum` for a pass by value type, but that is not always correct. Specifically for floating point numbers this would not always result in the correct floating point number. And int64 values are pass-by-reference on 32bit systems, but pass-by-value on 64bit systems. Postgres provides the `DatumGetXyz` macros/functions to work around these problems. This starts using these functions in `FilterOperationSwitch` to avoid these casting problems there.

Fixes #189

### DIFF
--- a/src/pgduckdb_filter.cpp
+++ b/src/pgduckdb_filter.cpp
@@ -5,7 +5,8 @@ extern "C" {
 #include "postgres.h"
 #include "catalog/pg_type.h"
 #include "utils/builtins.h"
-
+#include "utils/date.h"
+#include "utils/timestamp.h"
 }
 
 #include "pgduckdb/pgduckdb_filter.hpp"
@@ -16,8 +17,8 @@ namespace pgduckdb {
 
 template <class T, class OP>
 bool
-TemplatedFilterOperation(Datum &value, const duckdb::Value &constant) {
-	return OP::Operation((T)value, constant.GetValueUnsafe<T>());
+TemplatedFilterOperation(T value, const duckdb::Value &constant) {
+	return OP::Operation(value, constant.GetValueUnsafe<T>());
 }
 
 template <class OP>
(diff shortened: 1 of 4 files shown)


test/regression/sql/query_filter.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE query_filter_int(a INT);
+INSERT INTO query_filter_int SELECT g FROM generate_series(1,100) g;
+SELECT COUNT(*) FROM query_filter_int WHERE a  <= 50;
+DROP TABLE query_filter_int;
+
+CREATE TABLE query_filter_float(a FLOAT8);
+INSERT INTO query_filter_float VALUES (0.9), (1.0), (1.1);
+SELECT COUNT(*) = 1 FROM query_filter_float WHERE a < 1.0;
+SELECT COUNT(*) = 2 FROM query_filter_float WHERE a <= 1.0;
+SELECT COUNT(*) = 2 FROM query_filter_float WHERE a < 1.1;
```
JelteF, line 10: In my experience you should generally avoid returning booleans from queries in pgregress tests, because it makes it harder to debug the test when it starts failing. With your current code you only know that the test failed, but not why. With the following, you could see how many rows were found, which can be helpful with debugging. 
```suggestion
SELECT COUNT(*) FROM query_filter_float WHERE a < 1.0;
SELECT COUNT(*) FROM query_filter_float WHERE a <= 1.0;
SELECT COUNT(*) FROM query_filter_float WHERE a < 1.1;
```

src/pgduckdb_filter.cpp
```diff
@@ -55,9 +55,9 @@ FilterOperationSwitch(Datum &value, duckdb::Value &constant, Oid type_oid) {
 	case INT8OID:
 		return TemplatedFilterOperation<int64_t, OP>(value, constant);
 	case FLOAT4OID:
-		return TemplatedFilterOperation<float, OP>(value, constant);
+		return TemplatedFilterOperation<float, OP>(DatumGetFloat4(value), constant);
```
JelteF, line 58: It seems like we should probably do this explicit conversion for all types here.

JelteF (changes requested):

src/pgduckdb_filter.cpp
```diff
@@ -45,26 +45,26 @@ static bool
...
 	case FLOAT8OID:
-		return TemplatedFilterOperation<double, OP>(value, constant);
+		return TemplatedFilterOperation<double, OP>(DatumGetFloat8(value), constant);
 	case DATEOID: {
-		Datum date_datum = static_cast<int32_t>(value + pgduckdb::PGDUCKDB_DUCK_DATE_OFFSET);
-		return TemplatedFilterOperation<int32_t, OP>(date_datum, constant);
+		int32_t date = static_cast<int32_t>(value + pgduckdb::PGDUCKDB_DUCK_DATE_OFFSET);
```
JelteF, line 62:
```suggestion
		int32_t date = static_cast<int32_t>(DatumGetDate(value) + pgduckdb::PGDUCKDB_DUCK_DATE_OFFSET);
```

src/pgduckdb_filter.cpp
```diff
@@ -45,26 +45,26 @@ static bool
...
+		return TemplatedFilterOperation<int32_t, OP>(date, constant);
 	}
 	case TIMESTAMPOID: {
-		Datum timestamp_datum = static_cast<int64_t>(value + pgduckdb::PGDUCKDB_DUCK_TIMESTAMP_OFFSET);
-		return TemplatedFilterOperation<int64_t, OP>(timestamp_datum, constant);
+		int64_t timestamp = static_cast<int64_t>(value + pgduckdb::PGDUCKDB_DUCK_TIMESTAMP_OFFSET);
```
JelteF, line 66:
```suggestion
		int64_t timestamp = static_cast<int64_t>(DatumGetTimestamp(value) + pgduckdb::PGDUCKDB_DUCK_TIMESTAMP_OFFSET);
```

JelteF (approved):
Two minor suggestions, other than that this is good IMO.

JelteF:
I updated the PR description with some additional info/explanation of the problem.

mkaruza:
Update commit message.

JelteF:
> Update commit message.

Afaict the repo is configured to use the PR description as the commit message when merging by default, when doing a squash-merge. That's why I usually like to the PR description to explain a bit what the PR is doing before merging. (it's even better if the PR description is helpful when opening the PR ofcourse, to make it easy for reviewing, but in this case I was able to understand what the issue was from context I had).

JelteF (approved):

PR #195, opened 2024-09-19
Use correct output column id list for query processing 

### LINKED ISSUE #190: Equality comparison with varchar crashes
The following SQL consistently produces a crash on my machine. Looking at debugger some memory is clearly uninitialized/freed because it is repeating the `\xa5` byte over and over again.

```sql
create table t(a varchar);
insert into t values('t1');
set duckdb.execution = true;
select count(*) from t where a = 't1';
```

Possibly related to #189

mkaruza:
* When reading postgres tables we need to know which columns need to to
  written to output vector. Based on filtering and query we need to
  either get this information from `projection_ids` or `columns_ids`
  list. Projection ids list will be used when that are fetched columns
  from table but those columns are not needed in further query processing.
  Otherwise columns_ids list will be used.
* Fixes #190

### DIFF
--- a/include/pgduckdb/scan/postgres_scan.hpp
+++ b/include/pgduckdb/scan/postgres_scan.hpp
@@ -23,8 +23,8 @@ public:
 	TupleDesc m_tuple_desc;
 	std::mutex m_lock; // Lock for one replacement scan
 	bool m_count_tuples_only;
-	duckdb::map<duckdb::idx_t, duckdb::idx_t> m_columns;
-	duckdb::map<duckdb::idx_t, duckdb::idx_t> m_projections;
+	duckdb::map<duckdb::idx_t, duckdb::column_t> m_read_columns_ids;
+	duckdb::map<duckdb::idx_t, duckdb::column_t> m_output_columns_ids;
 	duckdb::TableFilterSet *m_filters = nullptr;
 	std::atomic<std::uint32_t> m_total_row_count;
 };
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -982,12 +982,18 @@ InsertTupleIntoChunk(duckdb::DataChunk &output, duckdb::shared_ptr<PostgresScanG
 	}
 
 	/* FIXME: all calls to duckdb_malloc/duckdb_free should be changed in future */
-	Datum *values = (Datum *)duckdb_malloc(sizeof(Datum) * scan_global_state->m_columns.size());
-	bool *nulls = (bool *)duckdb_malloc(sizeof(bool) * scan_global_state->m_columns.size());
+	Datum *values = (Datum *)duckdb_malloc(sizeof(Datum) * scan_global_state->m_output_columns_ids.size());
+	bool *nulls = (bool *)duckdb_malloc(sizeof(bool) * scan_global_state->m_read_columns_ids.size());
 
 	bool valid_tuple = true;
 
-	for (auto const &[columnIdx, valueIdx] : scan_global_state->m_columns) {
+	/* First we are fetching all required columns oredered by column id
+	 * and than we need to write this tuple into output vector. Output column id list
+	 * could be out of order so we need to match column values from first
+	 */
+
+	/* Read heap tuple with all required columns. */
+	for (auto const &[columnIdx, valueIdx] : scan_global_state->m_read_columns_ids) {
 		values[valueIdx] = HeapTupleFetchNextColumnDatum(scan_global_state->m_tuple_desc, tuple, heap_tuple_read_state,
 		                                                 columnIdx + 1, &nulls[valueIdx]);
 		if (scan_global_state->m_filters &&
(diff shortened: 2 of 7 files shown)


test/regression/sql/query_filter.sql
```diff
@@ -10,3 +10,10 @@ SELECT COUNT(*) = 2 FROM query_filter_float WHERE a <= 1.0;
 SELECT COUNT(*) = 2 FROM query_filter_float WHERE a < 1.1;
 
 DROP TABLE query_filter_float;
+
+CREATE TABLE query_filter_varchar(a VARCHAR);
+INSERT INTO query_filter_varchar VALUES ('t1'), ('t2'), ('t1');
+SELECT COUNT(*) = 2 FROM query_filter_varchar WHERE a = 't1';
+SELECT COUNT(a) = 2 FROM query_filter_varchar WHERE a = 't1';
```
JelteF, line 17: Same suggestion about returning the actual count as in #194

src/scan/postgres_scan.cpp
```diff
@@ -44,11 +44,11 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
 
 	if (input.CanRemoveFilterColumns()) {
 		for (duckdb::idx_t i = 0; i < input.projection_ids.size(); i++) {
-			m_projections[i] = input.column_ids[input.projection_ids[i]];
+			m_column_selection[i] = input.column_ids[input.projection_ids[i]];
 		}
 	} else {
-		for (duckdb::idx_t i = 0; i < input.projection_ids.size(); i++) {
```
JelteF, line 50: Just to be clear, this line contains the fix right? the rest of the PR is  just doing a field rename.

include/pgduckdb/scan/postgres_scan.hpp
```diff
@@ -24,7 +24,7 @@ class PostgresScanGlobalState {
 	std::mutex m_lock; // Lock for one replacement scan
 	bool m_count_tuples_only;
 	duckdb::map<duckdb::idx_t, duckdb::idx_t> m_columns;
-	duckdb::map<duckdb::idx_t, duckdb::idx_t> m_projections;
+	duckdb::map<duckdb::idx_t, duckdb::idx_t> m_column_selection;
```
JelteF, line 27:
```suggestion
	duckdb::map<duckdb::idx_t, duckdb::column_t> m_column_selection;
```

JelteF (changes requested):

src/scan/postgres_scan.cpp
```diff
@@ -44,11 +44,11 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
 
 	if (input.CanRemoveFilterColumns()) {
 		for (duckdb::idx_t i = 0; i < input.projection_ids.size(); i++) {
```
JelteF, line 46: Should this be changed to column_ids too?
```suggestion
		for (duckdb::idx_t i = 0; i < input.column_ids.size(); i++) {
```

src/scan/postgres_scan.cpp
```diff
@@ -44,11 +44,11 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
 
 	if (input.CanRemoveFilterColumns()) {
```
JelteF, line 45: I think adding some comments to this section would be very helpful. It's not super clear to me what the differences between these branches is (primarily because the comments in duckdb itself on the relevant functions and fields are non-existing).

src/pgduckdb_types.cpp
```diff
@@ -982,12 +982,18 @@ InsertTupleIntoChunk(duckdb::DataChunk &output, duckdb::shared_ptr<PostgresScanG
...
 	/* FIXME: all calls to duckdb_malloc/duckdb_free should be changed in future */
-	Datum *values = (Datum *)duckdb_malloc(sizeof(Datum) * scan_global_state->m_columns.size());
-	bool *nulls = (bool *)duckdb_malloc(sizeof(bool) * scan_global_state->m_columns.size());
+	Datum *values = (Datum *)duckdb_malloc(sizeof(Datum) * scan_global_state->m_output_columns_ids.size());
+	bool *nulls = (bool *)duckdb_malloc(sizeof(bool) * scan_global_state->m_read_columns_ids.size());
 
 	bool valid_tuple = true;
 
-	for (auto const &[columnIdx, valueIdx] : scan_global_state->m_columns) {
+	/* First we are fetching all required columns oredered by column id
```
JelteF, line 990:
```suggestion
	/* First we are fetching all required columns ordered by column id
```

test/regression/expected/query_filter.out
```diff
@@ -28,3 +28,51 @@ SELECT COUNT(*) FROM query_filter_float WHERE a < 1.1;
...
+-- Projection ids list will be used (column `a`is not needed after scan)
+SELECT b FROM query_filter_output_column WHERE a = 2;
+ b  
+----
+ t1
+ t2
+(2 rows)
+
+-- Column ids list used because both of fetched column are used after scan
+SELECT a, b FROM query_filter_output_column WHERE b = 't1';
```
JelteF, line 63: From your comments and my closer reading of the relevant duckdb code, we are expected to change the handle the column order changes. So let's include that in one of these tests
```suggestion
-- Column ids list used because both of fetched column are used after scan, we also should swap column order correctly
SELECT b, a FROM query_filter_output_column WHERE b = 't1';
```

src/scan/postgres_scan.cpp
JelteF: Turns out the answer is no

test/regression/expected/query_filter.out
JelteF: Let's also add a test where we only re-order columns, but still output all of them.

JelteF (dismissed):
Overall I think this looks pretty good.

src/scan/postgres_scan.cpp
mkaruza: In this branch, we need to use projection_ids list.

JelteF (approved):

Tishj:
I get this crash on `main`, likely related to this PR?

```
➜  pg_duckdb git:(main) ✗ psql postgres
 pg_backend_pid 
----------------
          29479
(1 row)

DROP EXTENSION
psql:/Users/thijs/.psqlrc:3: WARNING:  To actually execute queries using DuckDB you need to run "SET duckdb.execution TO true;"
CREATE EXTENSION
SET
psql (16.0)
Type "help" for help.

postgres=# SELECT a, c FROM query_filter_output_column WHERE b = 't1';
!?> 
```

PR #198, opened 2024-09-20
Fix cases where Postgres tuples don't store all columns

JelteF:
Since https://github.com/postgres/postgres/commit/16828d5c0273b4fe5f10f42588005f16b415b2d8 you can add columns with defaults in a way that does not require a table rewrite. The way this is implemented is that the default is stored for the attribute and when reading the tuple, if it has fewer columns than you expect, then you need to fetch the default. We were not doing this, so we would always read those columns as NULL.


src/pgduckdb_types.cpp
```diff
@@ -901,13 +902,19 @@ HeapTupleFetchNextColumnDatum(TupleDesc tupleDesc, HeapTuple tuple, HeapTupleRea
...
 	bool hasnulls = HeapTupleHasNulls(tuple);
+	bool natts = HeapTupleHeaderGetNatts(tup);
 	int attnum;
 	char *tp;
 	uint32 off;
 	bits8 *bp = tup->t_bits;
 	bool slow = false;
 	Datum value = (Datum)0;
 
+	if (natts + 1 < att_num) {
```
JelteF, line 913: I think this shouldn't have the `+ 1`, but removing that breaks other tests even more...
```suggestion
	if (natts < att_num) {
```

src/pgduckdb_types.cpp
```diff
@@ -901,13 +902,19 @@ HeapTupleFetchNextColumnDatum(TupleDesc tupleDesc, HeapTuple tuple, HeapTupleRea
                               int att_num, bool *is_null) {
 	HeapTupleHeader tup = tuple->t_data;
 	bool hasnulls = HeapTupleHasNulls(tuple);
+	bool natts = HeapTupleHeaderGetNatts(tup);
```
mkaruza, line 905: maybe int?

JelteF: Yes, that was it... Clearly time for the weekend...

Y-- (approved):
LGTM!

PR #200, opened 2024-09-22
Only load pg_duckdb extension if set in shared_preload_variables

mkaruza:
* We want to enforce that pg_duckdb extension is loaded only if it is set in shared_preload_variables.

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -1,5 +1,6 @@
 extern "C" {
 #include "postgres.h"
+#include "miscadmin.h"
 #include "utils/guc.h"
 }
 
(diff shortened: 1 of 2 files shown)


wuputah:
This is bad for users who cant modify that setting. Is it possible that this can this be avoided, or it this simply inevitably necessary given the direction of the design?

JelteF:
I agree that if we can avoid requiring this then that would be really nice. And I think it's possible at least for now. I thought introducing background workers would require it, but even that seems not true because `RegisterDynamicBackgroundWorker` can be used to spawn background workers without needing `shared_preload_libraries`.

mkaruza:
So i'm closing this at this point, and if needed in future we can return to this restriction.

PR #204, opened 2024-09-23
Create a dedicated clean-all rule

JelteF:
DuckDB takes a long time to compile (even with `ccache`/`sccache`) and often I don't actually want to rebuild it. For instance I only want to rebuild the actual extension whenever I switch between Postgres versions, the DuckDB library can stay the same. This changes the `clean` rule to only clean the actual extension files and adds a clean-all rule to clean everything.


wuputah (approved):
technically you should add it to `.PHONY` but otherwise lgtm

PR #205, opened 2024-09-23
Use Postgres its header dependency tracking instead of custom one

JelteF:
Building the `.depend` file takes ~1.5 seconds on my laptop (in power saver mode). And this is done every time I execute `make`. Postgres its Makefile.global has built in support for an equivalent feature that does not have this downside. The only downside is that it is not enabled on many Postgres builds, only if you compile with the `--enable-depend` flag.

This replaces our custom support for dependency tracking with version copied from the Postgres Makefile. This way even contributors using Postgres installations that come from their package repositories (which probably won't have `--enalbe-depend`) won't have to run `make clean` all the time.


wuputah (approved):
glad to see this, the `.depend` generation was definitely a nuisance.

Up to you, but maybe move the "vendored" rules into its own file and `include` it. Not sure where a good place to put it is though.

wuputah:
that works for me.

wuputah (approved):

PR #206, opened 2024-09-23
Make DuckDB connection a singleton

JelteF:
In #174 we made the DuckDB database a singleton, but a new DuckDB connection was still being created for every query. This means that any connection state, such as SET commands or temporary tables don't survive a single query, which is pretty confusing. This addresses that by also making the `duckdb::Connection` a singleton, this way a Postgres connection and a DuckDB connection are linked 1-to-1.


JelteF:
FYI I'm postponing working on this until later, since it's not needed for the first release and the current implementation has bugs with non-obvious causes.

JelteF:
Closing this in favor of #413

PR #208, opened 2024-09-24
Add the ability to run a single test

Tishj:
By setting `TEST=<test_name>` it will override the schedule and instead run only the test(s?) that was provided.


wuputah (approved):

JelteF:
Merged this, since I want to use it

PR #210, opened 2024-09-24
Cleanup after incorrect merge in #97

JelteF:
I think @Tishj misunderstood my review comment in #97 about the merge being incorrect. I meant that the whole `pgduckdb_pg_get_querydef` stuff could be removed from `DuckdbPlanNode`, since that is now moved to `DuckdbPrepare`.


Tishj (approved):
Ah I see, `query_string` is completely unused here because we prepared it earlier

PR #211, opened 2024-09-24
Makefile improvements

wuputah:
This fixes issues with `-j` when on a clean build.

* cleanup dependency chain of duckdb
* use wildcard for c and cpp files
* build duckdb completely before building pg_duckdb
* complete the build before installing
* change CI caching for Postgres to use sha

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,8 @@ ifeq ($(UNAME_S),Linux)
 	DUCKDB_LIB = libduckdb.so
 endif
 
-all: duckdb $(OBJS)
+all: $(OBJS)
+all-lib: duckdb
 
 include Makefile.global
 
--- a/Makefile
+++ b/Makefile
@@ -4,39 +4,11 @@ MODULE_big = pg_duckdb
 EXTENSION = pg_duckdb
 DATA = pg_duckdb.control $(wildcard sql/pg_duckdb--*.sql)
 
-SRCS = src/scan/heap_reader.cpp \
-	   src/scan/index_scan_utils.cpp \
-	   src/scan/postgres_index_scan.cpp \
-	   src/scan/postgres_scan.cpp \
-	   src/scan/postgres_seq_scan.cpp \
-	   src/utility/copy.cpp \
-	   src/vendor/pg_explain.cpp \
-	   src/pgduckdb_metadata_cache.cpp \
-	   src/pgduckdb_detoast.cpp \
-	   src/pgduckdb_duckdb.cpp \
-	   src/pgduckdb_filter.cpp \
-	   src/pgduckdb_hooks.cpp \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,40 +4,12 @@ MODULE_big = pg_duckdb
...
-	   src/pgduckdb_types.cpp \
-	   src/pgduckdb.cpp \
-	   src/catalog/pgduckdb_storage.cpp \
-	   src/catalog/pgduckdb_schema.cpp \
-	   src/catalog/pgduckdb_table.cpp \
-	   src/catalog/pgduckdb_transaction.cpp \
-	   src/catalog/pgduckdb_transaction_manager.cpp \
-	   src/catalog/pgduckdb_catalog.cpp
-
+SRCS = $(wildcard src/*.cpp src/*/*.cpp)
```
JelteF, line 7: I love love this

Makefile
```diff
@@ -120,7 +92,7 @@ lintcheck:
 	ruff check
 
 .depend:
-	$(RM) -f .depend
+	$(RM) .depend
```
JelteF, line 95: Let's undo this to avoid merge conflicts with #205

Makefile
```diff
@@ -68,7 +40,8 @@ ifeq ($(UNAME_S),Linux)
 	DUCKDB_LIB = libduckdb.so
 endif
 
-all: duckdb $(OBJS) .depend
+all: $(OBJS) .depend
+all-lib: duckdb
```
JelteF, line 44: I'm not sure I understand the point of this change.

wuputah: This was the most effective way I could find to make sure duckdb was built before pg_duckdb starts building.

I'm open to other suggestions, I'm certainly not a `make` wizard. 😁

wuputah: ok this isnt having the effect I thought it was having.

I also tried:

```
$(OBJS): duckdb
```

but this caused all of pg_duckdb to be rebuilt on every `make`, which didn't make sense to me.

JelteF: I think this might work:
```
$(OBJS): $(FULL_DUCKDB_LIB)
```

Things depending on a .PHONY means they always needs to be rebuilt (because make cannot do timestamp checks on a .PHONY target)

JelteF: If that doesn't work let's remove this all/all-lib change from this PR, because all the other things look reasonable.

JelteF (approved):

Makefile
wuputah: Finally realized this was some kind of race condition with the `install` step, so this was inadvertently exposed by using `make install` rather than `make` then `make install`. Adding a dependency to completely finish the build before installing seems to have fixed it.

JelteF (approved):

PR #221, opened 2024-09-27
Switch CI to Ubuntu 24.04

wuputah:
ubuntu-latest is actually 22.04

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -54,13 +54,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04]
         version: [REL_16_STABLE, REL_17_STABLE]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Test details
-        run: echo Build and test pg_duckdb on ${{ matrix.os }} with PostgreSQL ${{ matrix.version }} branch
+        run: echo Build and test pg_duckdb with PostgreSQL ${{ matrix.version }} branch
 
       - name: Install build deps
         run: |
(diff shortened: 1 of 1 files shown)


wuputah:
There's no underlying motivation to change this; I believe it was an oversight / assumption that ubuntu-latest was already 24.04.

* base branch atm `jd/makefile` as that is to be merged first, so this is a draft until then
* will need to disable then switch the main branch requirements to the new matrix values

wuputah:
This can be reviewed as-is but will need an admin to force a merge (bypass branch requirements) then adjust the `main` branch requirements after merging.

JelteF (approved):

JelteF:
> This can be reviewed as-is but will need an admin to force a merge (bypass branch requirements) then adjust the `main` branch requirements after merging.

Maybe let's remove the OS from the matrix while we're at it, that will save us from this same problem in ~2 years time.

wuputah:
> Maybe let's remove the OS from the matrix while we're at it, that will save us from this same problem in ~2 years time.

Sure, done in 2ceee6082d944ddac141196df22bd4f67454c5f9.

JelteF (approved):

wuputah:
24.04 is becoming the default for `ubuntu-latest` quite soon anyway, so without this PR things will randomly break at some point since we will need to bump the libtinfo dependency. I still think it's better for us to "hardcode" the version rather than be subjected to Github rollout schedule. More info is here: https://github.com/actions/runner-images/issues/10636

Mark is on PTO this week; probably fine to just wait for him to get back for this PR.

At252:
https://github.com/duckdb/pg_duckdb/pull/221#issue-2553791142

Mytherin:
Thanks for the PR! I've updated the branch protection rules to the new set of tests

PR #230, opened 2024-09-30
Document casing of C implementations of SQL functions

JelteF:
This is to make it easy to grep for these functions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,8 +86,8 @@ coding styles.
 ### Postgres C Guidelines
 
 * Use memory contexts to allocate memory, e.g. `palloc` or `palloc0`.
-* Casing is very inconsistent in Postgres, we use `CamelCase` for function names and `snake_case` for variables.
-* If you copied code from Postgres, try to keep the Postgres style and prefer to keep it in a separate function so that it is easy to update when the Postgres code changes.
+* Casing is very inconsistent in Postgres, we use `CamelCase` for function names and `snake_case` for variables. For the C implementation of a SQL function (e.g. `duckdb.install_extension(...)`, use snake_case.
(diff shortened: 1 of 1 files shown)


wuputah (approved):

JelteF:
Ugh... Why is skipping builds so hard if you also want to have branch-protections... : https://www.pantsbuild.org/blog/2022/10/10/skipping-github-actions-jobs-without-breaking-branch-protection

wuputah:
Mmmm, more CI hackery. We could turn off the path restrictions given the CI is quite quick now?

That said... while I'm certainly not opposed to the branch protection rules, if they're just causing problems, maybe we don't need them. Personally, I'm a fan of setting rules and trust people will do the right thing, rather than enforce the rules in software. In this case, that would mean no branch protection rules, but trust that committers will only merge if the tests pass. I don't think we've ever merged without tests passing, so we're not really solving a problem with the branch protection rules.

But... the auto-merge feature is nice.

JelteF:
> Mmmm, more CI hackery. We could turn off the path restrictions given the CI is quite quick now?

Yeah, I think that's probably the most pragmatic solution. In ~95% of the PRs we change some of these paths anyway. Really only doc-only PRs benefit from not running CI.

> Personally, I'm a fan of setting rules and trust people will do the right thing, rather than enforce the rules in software. 
In this case, that would mean no branch protection rules [for CI], but trust that committers will only merge if the tests pass. I don't think we've ever merged without tests passing, so we're not really solving a problem with the branch protection rules.

Agreed

> But... the auto-merge feature is nice.

Yeah, this is the main reason I would like to keep the branch protection rules. If we could have auto-merge work without them, I'd be all for not having required checks in the branch protection.

PR #231, opened 2024-09-30
Add DuckDB temporary tables

### LINKED ISSUE #224: Temporary tables stored in-memory in DuckDB
```
CREATE TEMP TABLE t(a int) using duckdb;
```

JelteF:
This adds support to create temporary tables using the DuckDB in-memory storage engine. This can be useful for certain ETL workflows and it's an important stepping-stone to implement MotherDuck support. This is done by using a new `duckdb` [Table Access Method (AM)](https://www.postgresql.org/docs/current/tableam.html). However, we don't really use any of the hooks that the Table AM interface provides us with. Basically all of these functions are not implement. The reason for this is that we only really care about the `USING duckdb` syntax as a way to mark certain tables as being stored in DuckDB. When we notice such a marker DuckDB completely takes over the execution of the query.

An important implication of this approach is that you cannot create Postgres indexes on DuckDB tables, which is a feature that the table access method interface normally gives you for free if you actually implement the methods. We don't care about this though since DuckDB wouldn't be able to use these indexes anyway.

Another similar interface that Postgres provides is [FOREIGN TABLEs](https://www.postgresql.org/docs/current/sql-createforeigntable.html). This is a way to define a table that is actually backed by an external data source. This has too many limitation to be useful for what we want to do though. For example, you cannot use CREATE TABLE AS with a FOREIGN TABLE. Also, you cannot create temporary foreign tables, which is exactly the feature that this PR provides.

Note: this PR does not implement CREATE TEMP TABLE AS yet. That will be done in a follow-up PR, so that this PR does not get too big.

Fixes #224


src/pgduckdb_hooks.cpp
```diff
@@ -67,38 +96,41 @@ ContainsDuckdbFunctions(Node *node, void *context) {
...
-	return query_tree_walker(query, ContainsDuckdbFunctions, NULL, 0);
-#else
-	return query_tree_walker(query, (bool (*)()) ((void *) ContainsDuckdbFunctions), NULL, 0);
-#endif
+	return ContainsDuckdbItems((Node *)query, NULL);
 }
 
 static bool
-IsAllowedStatement(Query *query) {
+IsAllowedStatement(Query *query, bool throw_error) {
```
Y--, line 111: nit- we could have kept `throw_error = false` by default? (would reduce the diff/simplify API?)

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,309 @@
...
+	 * write queries in transactions we'll need to link Postgres and DuckdB
+	 * their transaction lifecycles.
+	 */
+	PreventInTransactionBlock(true, "DuckDB queries");
+
+	for (auto proc = 0; proc < SPI_processed; proc++) {
+		HeapTuple tuple = SPI_tuptable->vals[proc];
+
+		char *object_identity = SPI_getvalue(tuple, SPI_tuptable->tupdesc, 1);
+		// TODO: Handle std::string creation in a safe way for allocation failures
```
Y--, line 236: Well the connection would leak too anyway :-)

src/pgduckdb_ruleutils.cpp
```diff
@@ -16,4 +28,215 @@ pgduckdb_function_name(Oid function_oid) {
...
+ * pgduckdb_get_tabledef returns the definition of a given table. This
+ * definition includes table's schema, default column values, not null and check
+ * constraints. The definition does not include constraints that trigger index
+ * creations; specifically, unique and primary key constraints are excluded.
+ *
+ * TODO: Add support indexes, primary keys and unique constraints.
+ *
+ * This function is inspired by the pg_get_tableschemadef_string function in
+ * the following patch that I (Jelte) submitted to Postgres in 2023:
+ * https://www.postgresql.org/message-id/CAGECzQSqdDHO_s8=CPTb2+4eCLGUscdh=KjYGTunhvrwcC7ZSQ@mail.gmail.com
```
Y--, line 76: I'm actually surprised there's not a readily available function in PG that does the job :-)

src/pgduckdb_ruleutils.cpp
```diff
@@ -16,4 +28,215 @@ pgduckdb_function_name(Oid function_oid) {
...
+ */
+char *
+pgduckdb_get_tabledef(Oid relation_oid) {
+	Relation relation = relation_open(relation_oid, AccessShareLock);
+	char *relation_name = pgduckdb_relation_name(relation_oid);
+	char *schema_name = get_namespace_name_or_temp(relation->rd_rel->relnamespace);
```
Y--, line 82: nit
```suggestion
	const char *relation_name = pgduckdb_relation_name(relation_oid);
	const char *schema_name = get_namespace_name_or_temp(relation->rd_rel->relnamespace);
```

src/pgduckdb_ruleutils.cpp
```diff
@@ -16,4 +28,215 @@ pgduckdb_function_name(Oid function_oid) {
...
+
+		/*
+		 * We disregard the inherited attributes (i.e., attinhcount > 0) here.
+		 * The reasoning behind this is that Citus implements declarative
+		 * partitioning by creating the partitions first and then sending
+		 * "ALTER TABLE parent_table ATTACH PARTITION .." command. This may
+		 * not play well with regular inherited tables, which isn't a big
+		 * concern from Citus' perspective.
+		 */
+		if (!column->attisdropped) {
```
Y--, line 129: nit
```suggestion
		if (column->attisdropped)
			continue;
```

src/pgduckdb_ruleutils.cpp
```diff
@@ -16,4 +28,215 @@ pgduckdb_function_name(Oid function_oid) {
...
+ * creations; specifically, unique and primary key constraints are excluded.
+ *
+ * TODO: Add support indexes, primary keys and unique constraints.
+ *
+ * This function is inspired by the pg_get_tableschemadef_string function in
+ * the following patch that I (Jelte) submitted to Postgres in 2023:
+ * https://www.postgresql.org/message-id/CAGECzQSqdDHO_s8=CPTb2+4eCLGUscdh=KjYGTunhvrwcC7ZSQ@mail.gmail.com
+ */
+char *
+pgduckdb_get_tabledef(Oid relation_oid) {
```
Y--, line 79: I wonder if we could re-use the C++ test framework I started sketching to unit-test this method? (seems like a good candidate :-)

Y-- (approved):
Really nice piece of work! Added a few minor comments, but nothing blocking. Would be interested to here @mkaruza and/or @Tishj thoughts too.

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,309 @@
...
+	int ret = SPI_exec(R"(
+                INSERT INTO duckdb.tables(relid)
+                SELECT DISTINCT objid
+                FROM pg_catalog.pg_event_trigger_ddl_commands() cmds
+                JOIN pg_catalog.pg_class
+                ON cmds.objid = pg_class.oid
+                WHERE cmds.object_type = 'table'
+                AND pg_class.relam = (SELECT oid FROM pg_am WHERE amname = 'duckdb')
+				RETURNING relid)",
+	                   0);
```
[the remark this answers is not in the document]
JelteF: Hmm, that's definitely by accident. I'll fix that.

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,309 @@
...
+ */
+void
+DuckdbTruncateTable(Oid relation_oid) {
+	auto db = pgduckdb::DuckDBManager::Get().GetDatabase();
+	auto connection = duckdb::make_uniq<duckdb::Connection>(db);
+	auto &context = *connection->context;
+	auto query = std::string("TRUNCATE ") + pgduckdb_relation_name(relation_oid);
+	auto result = context.Query(query, false);
+	if (result->HasError()) {
+		elog(ERROR, "(PGDuckDB/DuckdbTruncateTable) Could not truncate table: %s", result->GetError().c_str());
```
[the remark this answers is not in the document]
JelteF: Yeah, makes sense. I indeed thought I'd let you figure out a bit how we should handle this in an elegant way, and do it in a crude but working way now. Especially these DDL commands are hopefully not run nearly as often as the regular SELECT queries, so a relatively small leak would probably not be the end of the world.

I think I'll leave it like this for this PR. Especially since your solution isn't perfect yet afaict, because it would still leak the error string. We'd have to copy that over to a memory context, using e.g. `pstrdup` and then manually deallocate (destroy?) the string.

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,309 @@
...
+		elog(ERROR, "DuckDB does not support CREATE TABLE AS yet");
+	}
+}
+
+/*
+ * Throws an error when an unsupported ON COMMIT clause is used. DuckDB does
+ * not support ON COMMIT DROP, and it's difficult to emulate because Postgres
+ * does not call any hooks or event triggers when dropping a table due to an ON
+ * COMMIT DROP clause. We could simplify the below check to only check for
+ * ONCOMMIT_DROP, but this semes safer in case
```
Tishj, line 62: seems*
also, is that last sentenced finished?

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,309 @@
...
+	EventTriggerData *trigdata = (EventTriggerData *)fcinfo->context;
+	Node *parsetree = trigdata->parsetree;
+
+	SPI_connect();
+
+	/* Temporarily escalate privileges to superuser so we can insert into duckdb.tables */
+	Oid saved_userid;
+	int sec_context;
+	auto save_nestlevel = NewGUCNestLevel();
+	SetConfigOption("search_path", "", PGC_USERSET, PGC_S_SESSION);
```
Tishj, line 95: Why do we need to set the search path to `""`?

src/pgduckdb_ddl.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+	/* Revert back to original privileges */
+	SetUserIdAndSecContext(saved_userid, sec_context);
+	AtEOXact_GUC(false, save_nestlevel);
+
+	if (ret != SPI_OK_INSERT_RETURNING)
+		elog(ERROR, "SPI_exec failed: error code %s", SPI_result_code_string(ret));
+
+	/* if we inserted a row it was a duckdb table */
+	auto isDuckdbTable = SPI_processed > 0;
```
Tishj, line 122: NIT: snake case-style for variables?

sql/pg_duckdb--0.0.1.sql
```diff
@@ -139,12 +139,47 @@ CREATE TABLE extensions (
     enabled BOOL DEFAULT TRUE
 );
 
+CREATE TABLE tables (
```
Tishj, line 142: I am a little confused why we need this, I thought a Relation already contained a pointer to the table AM struct, which we can use to check if it's a duckdb temp table?

src/pgduckdb_ruleutils.cpp
Tishj: nit on your nit:
```
if (column->attisdropped) {
    continue;
}
```

src/pgduckdb_ruleutils.cpp
```diff
@@ -16,4 +28,215 @@ pgduckdb_function_name(Oid function_oid) {
...
+				/*
+				 * convert expression to node tree, and prepare deparse
+				 * context
+				 */
+				Node *default_node = (Node *)stringToNode(default_value->adbin);
+
+				/* deparse default value string */
+				char *default_string = pgduckdb_deparse_expression(default_node, relation_context, false, false);
+
+				if (column->attgenerated == ATTRIBUTE_GENERATED_STORED) {
```
Tishj, line 170: DuckDB does support generated virtual columns, which should not be set as a `DEFAULT`

JelteF: Yeah, I realized that. But postgres does not support VIRTUAL columns. I updated the code and comment to be more clear about that.

src/pgduckdb_ruleutils.cpp
JelteF: Yeah, it's quite annoying that there isn't.

src/pgduckdb_ruleutils.cpp
Y--: ++ to always have a block :-)

src/pgduckdb_ruleutils.cpp
JelteF: I think it could be, but I think getting good test coverage here with the normal regression tests is actually pretty easy. And it's actually good that we have these end to end tests here. For now I'll leave this as is.

JelteF:
I believe I addressed all feedback here btw.

mkaruza (approved):

Y-- (approved):

PR #237, opened 2024-10-01
Use -Werror in CI

JelteF:
The PG15 support that was introduced in #180 introduced a bunch of warnings when compiling. To avoid getting numb to warnings, this adds the -Werror to our CI builds, so we catch warnings before merging to main.

This also starts to ignore the register warning which was introduced by PG15 support.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -113,7 +113,7 @@ jobs:
         id: build
         run: |
           pushd duckdb
-          make -j8 install
+          make -j8 install CFLAGS=-Werror CXXFLAGS=-Werror
 
       - name: Run make installcheck
         id: installcheck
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
         id: build
         run: |
           pushd duckdb
-          make -j8 install
+          make -j8 install CFLAGS=-Werror CXXFLAGS=-Werror
```
Y--, line 116: Why not using it in the Makefile as well? Too annoying in dev cycles?
In this case, maybe we want a dedicated make target?

Y--: Fair enough - if we need this we can always add it later.

Y-- (approved):

PR #240, opened 2024-10-01
upgrade to duckdb 1.1.1

wuputah:


wuputah:
@mkaruza looks like there's a diff to apply to httpfs

wuputah:
oh this is actually due to `-Wall`. easy fix though.

JelteF (approved):

PR #241, opened 2024-10-02
Add support for CREATE TABLE AS for duckdb TableAM

JelteF:
This is a follow up PR to #231 which adds support for CREATE TABLE AS for tables created with `USING duckdb`. With this change you can now run queries such as the following:

```
CREATE TEMP TABLE webpages USING duckdb AS SELECT * FROM read_csv('../../data/web_page.csv') as (column00 int, column01 text, column02 date);
```

The implementation does have the following limitations:
1. `EXPLAIN ANALYZE CREATE TABLE ... USING duckdb AS ...` does not work
2. It's not possible to use parameters of a prepared statement in the query that's part of the CREATE TABLE AS.

### DIFF
--- a/test/regression/sql/temporary_tables.sql
+++ b/test/regression/sql/temporary_tables.sql
@@ -146,7 +146,25 @@ DROP TABLE t;
 -- unsupported
 CREATE TEMP TABLE t(a int) ON COMMIT DROP;
 
+-- CTAS fully in Duckdb
 CREATE TEMP TABLE webpages USING duckdb AS SELECT * FROM read_csv('../../data/web_page.csv') as (column00 int, column01 text, column02 date);
+SELECT * FROM webpages ORDER BY column00 LIMIT 2;
+
+CREATE TEMP TABLE t_heap(a int) USING heap;
+INSERT INTO t_heap VALUES (1);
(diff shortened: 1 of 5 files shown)


mkaruza (approved):

test/regression/sql/temporary_tables.sql
```diff
@@ -146,7 +146,25 @@ DROP TABLE t;
...
+SELECT * FROM webpages ORDER BY column00 LIMIT 2;
+
+CREATE TEMP TABLE t_heap(a int) USING heap;
+INSERT INTO t_heap VALUES (1);
+
+-- CTAS from postgres table to duckdb table
+CREATE TEMP TABLE t(b) USING duckdb AS SELECT * FROM t_heap;
+SELECT * FROM t;
+
+-- CTAS from DuckDB table to postgres table
```
dpxcc, line 160: I think this should be `CTAS from postgres table to postgres table`, but not `CTAS from DuckDB table to postgres table`

And do we plan to support write to postgres table using DuckDB?
I noticed that DuckDB's Postgres storage extension supports that: https://github.com/duckdb/duckdb-postgres

Issue #245, opened 2024-10-03
Query lead to crash


Y--:
This query crashes PG:
```
CREATE TABLE foo AS SELECT 'bar' AS t;
BEGIN; SET duckdb.execution = true; SELECT t::integer AS t1 FROM foo; ROLLBACK;
SET duckdb.execution = true; 
SELECT 1 FROM foo;
```

JelteF:
This specific issue is fixed in #231, because that starts to disallow duckdb queries transaction blocks. I also think that the simplification of our connection management logic due to removal of index scan support might resolve this too. Whenever we have something merged that fixes this we should add this as a regression test.

PR #246, opened 2024-10-03
Call `GetActiveSnapshot` when creating TX

Y--:
When moving from the creation of a DuckDB instance for each query to a singleton, we kept the same snapshot. This is obviously not correct.

This PR calls `GetActiveSnapshot` under a lock when creating a transaction.

Fixes https://github.com/duckdb/pg_duckdb/issues/245

### DIFF
--- a/src/catalog/pgduckdb_transaction_manager.cpp
+++ b/src/catalog/pgduckdb_transaction_manager.cpp
@@ -10,13 +10,7 @@ PostgresTransactionManager::PostgresTransactionManager(AttachedDatabase &db_p, P
 
 Transaction &
 PostgresTransactionManager::StartTransaction(ClientContext &context) {
-	Snapshot snapshot;
-	{
-		std::lock_guard<std::mutex> lock(pgduckdb::DuckdbProcessLock::GetLock());
-		snapshot = GetActiveSnapshot();
-	}
-
-	auto transaction = make_uniq<PostgresTransaction>(*this, context, catalog, snapshot);
+	auto transaction = make_uniq<PostgresTransaction>(*this, context, catalog, GetActiveSnapshot());
 	auto &result = *transaction;
 	lock_guard<mutex> l(transaction_lock);
 	transactions[result] = std::move(transaction);
--- a/src/catalog/pgduckdb_transaction_manager.cpp
+++ b/src/catalog/pgduckdb_transaction_manager.cpp
@@ -1,5 +1,6 @@
 #include "pgduckdb/catalog/pgduckdb_transaction_manager.hpp"
 #include "duckdb/main/attached_database.hpp"
+#include "pgduckdb/pgduckdb_process_lock.hpp"
 
 namespace duckdb {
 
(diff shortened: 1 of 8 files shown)


src/catalog/pgduckdb_transaction_manager.cpp
```diff
@@ -31,6 +30,7 @@ PostgresTransactionManager::RollbackTransaction(Transaction &transaction) {
 
 void
 PostgresTransactionManager::Checkpoint(ClientContext &context, bool force) {
+	printf("PostgresTransactionManager::Checkpoint\n");
```
Y--, line 33: Leftover, will remove

src/catalog/pgduckdb_transaction_manager.cpp
```diff
@@ -3,16 +3,15 @@
...
-    : TransactionManager(db_p), catalog(catalog), snapshot(snapshot) {
+PostgresTransactionManager::PostgresTransactionManager(AttachedDatabase &db_p, PostgresCatalog &catalog)
+    : TransactionManager(db_p), catalog(catalog) {
 }
 
 Transaction &
 PostgresTransactionManager::StartTransaction(ClientContext &context) {
-	auto transaction = make_uniq<PostgresTransaction>(*this, context, catalog, snapshot);
-	auto &result = *transaction;
 	lock_guard<mutex> l(transaction_lock);
```
JelteF, line 12: Seems like this is probably not the right lock. I think we need to take the global postgres lock.

JelteF (approved):
I think it's good @Tishj can you take a final look too, since you have worked on the catalog extension.

test/regression/sql/transaction_errors.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE foo AS SELECT 'bar' AS t;
+BEGIN; SET duckdb.execution = true; SELECT t::integer AS t1 FROM foo; ROLLBACK;
+SET duckdb.execution = true;
+SELECT 1 FROM foo;
```
JelteF, line 4: missing newline at end of file

Y--: I would expect `make format` to catch these - is it normal that it doesn't?

JelteF: We should probably add a script to check for that. `make format` only handles C, C++ and Python files so we need something different for all others (SQL, markdown, _schedule). If you install editorconfig it should do this on save though.

Y--: Yep I did now, but I didn't go through all the files in this PR earlier (now I did :)

src/catalog/pgduckdb_transaction_manager.cpp
```diff
@@ -1,15 +1,21 @@
...
-    : TransactionManager(db_p), catalog(catalog), snapshot(snapshot) {
+PostgresTransactionManager::PostgresTransactionManager(AttachedDatabase &db_p, PostgresCatalog &catalog)
+    : TransactionManager(db_p), catalog(catalog) {
 }
 
 Transaction &
 PostgresTransactionManager::StartTransaction(ClientContext &context) {
+	Snapshot snapshot;
+	{
+		std::lock_guard<std::mutex> lock(pgduckdb::DuckdbProcessLock::GetLock());
```
Tishj, line 15: Does it need the lock? I thought that snapshots were transaction-local, created by postgres when a transaction started and GetActiveSnapshot just references that cached instance?

Or is it created lazily and that's why we need to protect the catalog from getting modified?
I would think postgres has their own locks to protect against this, we're not changing any global data in the shared duckdb db right?

Tishj (approved):
Thanks, this looks good to me
I raised a question around the locking of GetActiveSnapshot, but that's more for my own understanding (and maybe would benefit from a comment)

src/catalog/pgduckdb_transaction_manager.cpp
JelteF: Agreed, the lock should is not necessary. GetActiveSnapshot only does a read, so no need to serialize those. There are ofcourse functions that change the activesnapshot, but we don't call those functions.

JelteF (approved):

PR #247, opened 2024-10-03
Revert index scan

### LINKED ISSUE #183: Revert index scan functionality
@Tishj mentioned in #97 that we don't have tests for index scans. So I tried adding some basic tests for this. And as it turns out index scans are very much (and possibly completely) broken. This trivial example crashes the server for me:

```sql
CREATE TABLE test_table (id int primary key, name text);
SELECT * FROM test_table where id = 1;
```

mkaruza:
Reverting index scan feature as explained in #183. Removed query planning from execution. 

Three commits that should not be squashed when merging.

Fixes #183


src/catalog/pgduckdb_table.cpp
```diff
@@ -36,7 +35,7 @@ PostgresTable::PopulateColumns(CreateTableInfo &info, Oid relid, Snapshot snapsh
 	auto tupleDesc = RelationGetDescr(rel);
 
 	if (!tupleDesc) {
-		elog(ERROR, "Failed to get tuple descriptor for relation with OID %u", relid);
```
JelteF, line 39: I think this is probably a good change, but let's move it to a separate PR. It's nice to keep behavioral and refactoring changes separate, both for review and possible future `git revert`.

include/pgduckdb/catalog/pgduckdb_table.hpp
```diff
@@ -52,20 +52,4 @@ class PostgresHeapTable : public PostgresTable {
...
-
-public:
-	// -- Table API --
-	unique_ptr<BaseStatistics> GetStatistics(ClientContext &context, column_t column_id) override;
-	TableFunction GetScanFunction(ClientContext &context, unique_ptr<FunctionData> &bind_data) override;
-	TableStorageInfo GetStorageInfo(ClientContext &context) override;
-
-private:
-	Path *path;
-	PlannerInfo *planner_info;
```
JelteF, line 68: Do we still need the PlannerInfo to be passed into the sequence scan now? i.e. I think we were only using it to choose between sequence scan and index scan, so I think we can get rid of storing it in the `"postgres_state"` key of `context->registered_state`.

Tishj: I think you're right, I think that might also remove the entire need for `PostgresContextState`, I believe `m_rtables` and `m_needed_columns` as well as `m_query_string` are not used ?

src/catalog/pgduckdb_table.cpp
```diff
@@ -47,14 +47,25 @@ PostgresTable::PopulateColumns(CreateTableInfo &info, Oid relid, Snapshot snapsh
...
 		     duck_type.ToString().c_str());
 	}
 
 	RelationClose(rel);
 	return true;
 }
 
+Cardinality
+PostgresTable::GetTableCardinality(Oid relid) {
+	auto rel = RelationIdGetRelation(relid);
```
JelteF, line 60: I think this probably needs the `DuckdbProcessLock`.

include/pgduckdb/catalog/pgduckdb_table.hpp
JelteF: I'm also fine with doing that in a follow up PR btw.

JelteF (approved):
Apart from the missing process lock I think this can be merged. 

Like @Tishj said I think we can remove more code. But I'm also fine with that happening in a follow up PR.

src/catalog/pgduckdb_table.cpp
mkaruza: Let me give another pass through code; I'm seeing this `RelationIdGetRelation` used on few other places (unprotected by lock) maybe it is worth checking if this function can be used in only one function and than we would pass pointer to its data.

mkaruza: @Tishj can you look into latest commit it should centralize opening/closing relation for execution

src/catalog/pgduckdb_table.cpp
```diff
@@ -25,18 +27,29 @@ extern "C" {
...
 bool
-PostgresTable::SetTableInfo(CreateTableInfo &info, Oid relid, Snapshot snapshot) {
-	auto rel = RelationIdGetRelation(relid);
+PostgresTable::SetTableInfo(CreateTableInfo &info, ::Relation rel) {
 	auto tupleDesc = RelationGetDescr(rel);
 
 	if (!tupleDesc) {
-		elog(WARNING, "Failed to get tuple descriptor for relation with OID %u", relid);
+		elog(WARNING, "Failed to get tuple descriptor for relation with OID %u", rel->rd_id);
 		RelationClose(rel);
```
JelteF, line 53: This one should now be removed I think.

JelteF (approved):
One small comment, but other than that this looks good now.

src/catalog/pgduckdb_table.cpp
mkaruza: Yes, thanks. I'm thinking if this check is even needed? If we have relation object opened before should we always have tuple desc information?

JelteF:
Oh one more small note (I forgot it last time too when merging a PR without squashing). Can you update the commit messages to include the PR number at the end of the title in parenthesis, ie. ` (#247)`

src/catalog/pgduckdb_table.cpp
JelteF: Yeah, that sounds reasonable. There's no way for this to fail. It's a simple macro that reads a field:
```
#define RelationGetDescr(relation) ((relation)->rd_att)
```

JelteF (approved):

PR #248, opened 2024-10-03
fix builds that use older clang versions

wuputah:
pass PG_CPPFLAGS and PG_CXXFLAGS to COMPILE.cxx.bc like we did for COMPILE.cc.bc

this fixes the build on ubuntu:22.04

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,15 +29,15 @@ override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register ${DUCKDB_BUIL
 
 SHLIB_LINK += -Wl,-rpath,$(PG_LIB)/ -lpq -Lthird_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src -L$(PG_LIB) -lduckdb -lstdc++ -llz4
 
-COMPILE.cc.bc = $(CXX) -Wno-ignored-attributes -Wno-register $(BITCODE_CXXFLAGS) $(CXXFLAGS) $(PG_CPPFLAGS) $(PG_CXXFLAGS) -I$(INCLUDEDIR_SERVER) -emit-llvm -c
-COMPILE.cxx.bc = $(CLANG) -xc++ -Wno-ignored-attributes $(BITCODE_CXXFLAGS) $(CPPFLAGS) $(PG_CPPFLAGS) $(PG_CXXFLAGS) -flto=thin -emit-llvm -c
-
 include Makefile.global
 
 # We need the DuckDB header files to build the .o files. We depend on the
 # duckdb Makefile, because that target pulls in the submodule which includes
 # those header files.
 $(OBJS): third_party/duckdb/Makefile
+
+COMPILE.cc.bc += $(PG_CPPFLAGS) $(PG_CXXFLAGS)
+COMPILE.cxx.bc += $(PG_CPPFLAGS) $(PG_CXXFLAGS)
 
 # shlib is the final output product - make duckdb and all .o dependencies
 $(shlib): $(FULL_DUCKDB_LIB) $(OBJS)
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,8 @@ include Makefile.global
 # those header files.
 $(OBJS): third_party/duckdb/Makefile
 
-COMPILE.cc.bc += $(PG_CPPFLAGS) $(PG_CXXFLAGS)
-COMPILE.cxx.bc += $(PG_CPPFLAGS) $(PG_CXXFLAGS)
+COMPILE.cc.bc += $(PG_CPPFLAGS)
+COMPILE.cxx.bc += $(PG_CXXFLAGS)
 
 # shlib is the final output product - make duckdb and all .o dependencies
 $(shlib): $(FULL_DUCKDB_LIB) $(OBJS)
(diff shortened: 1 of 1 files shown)


wuputah:
Another cause looks like our `%.bc: %.cpp` rule isn't being picked up as it's defined before the `include`. However, if moved after the `include`, that rule doesn't seem to work.

```g++ -Wno-ignored-attributes -Wno-register -fno-strict-aliasing -fwrapv -Xclang -no-opaque-pointers -O2  -Wall -Wpointer-arith -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wshadow=compatible-local -Wformat-security -fno-strict-aliasing -fwrapv -moutline-atomics -g -g -O2 -flto=auto -ffat-lto-objects -flto=auto -ffat-lto-objects -fstack-protector-strong -Wformat -Werror=format-security -std=c++17 -Wno-sign-compare -Wno-register  -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -Iinclude -Ithird_party/duckdb/src/include -Ithird_party/duckdb/third_party/re2 -std=c++17 -Wno-sign-compare -Wno-register  -I/usr/include/postgresql/16/server -emit-llvm -c -Wl,-rpath,/usr/lib/postgresql/16/lib/ -lpq -Lthird_party/duckdb/build/release/src -L/usr/lib/postgresql/16/lib -lduckdb -lstdc++ -llz4  -I -o src/pgduckdb.bc src/pgduckdb.cpp
g++: error: unrecognized command-line option '-Xclang'
g++: error: unrecognized command-line option '-no-opaque-pointers'```

**Review by Y-- — APPROVED** [review]
LGTM if it works :-)

File: Makefile [hunk 1]
```diff
@@ -30,9 +30,7 @@ override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register ${DUCKDB_BUIL
 SHLIB_LINK += -Wl,-rpath,$(PG_LIB)/ -lpq -Lthird_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src -L$(PG_LIB) -lduckdb -lstdc++ -llz4
 
 COMPILE.cc.bc = $(CXX) -Wno-ignored-attributes -Wno-register $(BITCODE_CXXFLAGS) $(CXXFLAGS) $(PG_CPPFLAGS) $(PG_CXXFLAGS) -I$(INCLUDEDIR_SERVER) -emit-llvm -c
-
-%.bc : %.cpp
-	$(COMPILE.cc.bc) $(SHLIB_LINK) -I$(INCLUDE_SERVER) -o $@ $<
+COMPILE.cxx.bc = $(CLANG) -xc++ -Wno-ignored-attributes $(BITCODE_CXXFLAGS) $(CPPFLAGS) $(PG_CPPFLAGS) $(PG_CXXFLAGS) -flto=thin -emit-llvm -c
```
Y--, line 33: Two minor comments actually:
* do we need to tweak the link time optimizations here? Sounds unrelated no?
* can we factorize some of the flags with `COMPILE.cc.bc`? (maybe overkill though)

Makefile
```diff
@@ -30,9 +30,7 @@ override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register ${DUCKDB_BUIL
 SHLIB_LINK += -Wl,-rpath,$(PG_LIB)/ -lpq -Lthird_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src -L$(PG_LIB) -lduckdb -lstdc++ -llz4
 
 COMPILE.cc.bc = $(CXX) -Wno-ignored-attributes -Wno-register $(BITCODE_CXXFLAGS) $(CXXFLAGS) $(PG_CPPFLAGS) $(PG_CXXFLAGS) -I$(INCLUDEDIR_SERVER) -emit-llvm -c
```
JelteF, line 32: It seems confusing that the `CXXFLAGS` are passed to `COMPILE.cc.bc`. 
```suggestion
COMPILE.cc.bc = $(CXX) -Wno-ignored-attributes -Wno-register $(BITCODE_CFLAGS) $(CPPFLAGS) $(PG_CPPFLAGS) -I$(INCLUDEDIR_SERVER) -emit-llvm -c
```

Makefile
JelteF: Afaict this is mostly copied from the postgres makefile. where it looks like this:
```make
ifndef COMPILE.c.bc
# -Wno-ignored-attributes added so gnu_printf doesn't trigger
# warnings, when the main binary is compiled with C.
COMPILE.c.bc = $(CLANG) -Wno-ignored-attributes $(BITCODE_CFLAGS) $(CPPFLAGS) -flto=thin -emit-llvm -c
endif

ifndef COMPILE.cxx.bc
COMPILE.cxx.bc = $(CLANG) -xc++ -Wno-ignored-attributes $(BITCODE_CXXFLAGS) $(CPPFLAGS) -flto=thin -emit-llvm -c
endif
```

So I think this one looks okay, i.e. the only difference is these two additions: $(PG_CPPFLAGS) $(PG_CXXFLAGS)

JelteF: Would be good to add a comment about that though. Something like. These two lines were copied from Postgres, the only difference is that we added $(PG_CPPFLAGS) and $(PG_CXXFLAGS).

Makefile
wuputah: I believe this is now addressed via 151afafdc31a0026846bcbaf5460b37109201ca0

Makefile
```diff
@@ -29,18 +29,16 @@ override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register ${DUCKDB_BUIL
...
-
 include Makefile.global
 
 # We need the DuckDB header files to build the .o files. We depend on the
 # duckdb Makefile, because that target pulls in the submodule which includes
 # those header files.
 $(OBJS): third_party/duckdb/Makefile
 
+COMPILE.cc.bc += $(PG_CPPFLAGS)
+COMPILE.cxx.bc += $(PG_CXXFLAGS)
```
JelteF, line 40: I'm guessing this still needs the PG_CPPFLAGS too.
```suggestion
COMPILE.cxx.bc += $(PG_CXXFLAGS) $(PG_CPPFLAGS)
```

wuputah:
Build is failing for the `-Wall` issue I mentioned in #240 -- now that duckdb is being recompiled after switching to 24.04.

Makefile
wuputah: doesn't seem so! everything looking good on 22.04, 24.04, and locally on MacOS.

JelteF (approved):

PR #251, opened 2024-10-04
Fix memory leaks in custom executor

Y--:
When one calls `elog(ERROR,...` it `longjmp` out of the current stack leaving no chance for the C++ to be executed.

This PR is a first iteration toward what we want to rationalize (cf. https://github.com/duckdb/pg_duckdb/issues/93 ):
* keep a function with a pure C++ "flavor" (object creation, exceptions, etc.)
* call it in a new wrapper that copies out the error message and uses PG mechanism to throw the error

This fixes https://github.com/duckdb/pg_duckdb/issues/120, even though there still seem to be a small leak with DuckDB's `PrepareQuery` leftover (to be followed-up separately)

### DIFF
--- a/src/pgduckdb_node.cpp
+++ b/src/pgduckdb_node.cpp
@@ -124,7 +124,7 @@ ExecuteQuery(DuckdbScanState *state) {
 
 	auto pending = prepared.PendingQuery(duckdb_params, true);
 	if (pending->HasError()) {
-		throw std::runtime_error(pending->GetError());
+		return pending->ThrowError();
 	}
 
 	duckdb::PendingExecutionResult execution_result;
@@ -144,7 +144,7 @@ ExecuteQuery(DuckdbScanState *state) {
 	} while (!duckdb::PendingQueryResult::IsResultReady(execution_result));
 
 	if (execution_result == duckdb::PendingExecutionResult::EXECUTION_ERROR) {
-		throw std::runtime_error(pending->GetError());
+		return pending->ThrowError();
 	}
 
 	query_results = pending->Execute();
--- a/src/pgduckdb_node.cpp
+++ b/src/pgduckdb_node.cpp
@@ -93,18 +93,20 @@ ExecuteQuery(DuckdbScanState *state) {
 		ParamExternData tmp_workspace;
 
 		/* give hook a chance in case parameter is dynamic */
-		if (pg_params->paramFetch != NULL)
+		if (pg_params->paramFetch != NULL) {
 			pg_param = pg_params->paramFetch(pg_params, i + 1, false, &tmp_workspace);
-		else
+		} else {
 			pg_param = &pg_params->params[i];
+		}
 
 		if (pg_param->isnull) {
 			duckdb_params.push_back(duckdb::Value());
-		} else {
-			if (!OidIsValid(pg_param->ptype)) {
-				elog(ERROR, "parameter with invalid type during execution");
-			}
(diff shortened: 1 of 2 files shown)


src/pgduckdb_node.cpp
```diff
@@ -92,25 +105,28 @@ ExecuteQuery(DuckdbScanState *state) {
...
+			std::ostringstream oss;
+			oss << "parameter " << i << " has an invalid type (" << pg_param->ptype << ") during query execution";
+			throw duckdb::Exception(duckdb::ExceptionType::EXECUTOR, oss.str().c_str());
 		}
 	}
 
 	auto pending = prepared.PendingQuery(duckdb_params, true);
 	if (pending->HasError()) {
-		elog(ERROR, "DuckDB execute returned an error: %s", pending->GetError().c_str());
+		throw std::runtime_error(pending->GetError());
```
Tishj, line 127: Can use `pending->ThrowError()` instead

Tishj (approved):
Thanks, if I'm reading this correctly, we add DuckDBFunctionGuard to catch a c++ exception so we can perform cleanup on the postgres side before propagating this error upward?

src/pgduckdb_node.cpp
```diff
@@ -121,16 +137,16 @@ ExecuteQuery(DuckdbScanState *state) {
...
 			ProcessInterrupts();
-			elog(ERROR, "Query cancelled");
+			throw duckdb::Exception(duckdb::ExceptionType::EXECUTOR, "Query cancelled");
 		}
 	} while (!duckdb::PendingQueryResult::IsResultReady(execution_result));
+
 	if (execution_result == duckdb::PendingExecutionResult::EXECUTION_ERROR) {
-		CleanupDuckdbScanState(state);
-		elog(ERROR, "(PGDuckDB/ExecuteQuery) %s", pending->GetError().c_str());
+		throw std::runtime_error(pending->GetError());
```
Tishj, line 147: Same comment as above

src/pgduckdb_node.cpp
Y--: That's what I did initially but sadly it throws an ugly JSON then... Maybe I missed something, I'll try again.

Y--:
> Thanks, if I'm reading this correctly, we add DuckDBFunctionGuard to catch a c++ exception so we can perform cleanup on the postgres side before propagating this error upward?

Yep, that's correct!

src/pgduckdb_node.cpp
Y--: I tried again and I got something like this:

```
ERROR:  (PGDuckDB/ExecuteQuery) {"exception_type":"Conversion","exception_message":"Could not convert string 'not an integer' to INT32\nLINE 1: SELECT (t)::integer AS t198 FROM public.foo\n                  ^","position":"10"}
```

Is there something specific to do so it doesn't output JSON?

Y--: @Tishj I've pushed that [commit](https://github.com/duckdb/pg_duckdb/pull/251/commits/a91f2e636f08181febf0a621868d26f35fc5bd7d) to support the `pending->ThrowError()`. 
Maybe I missed something from DuckDB API, but I'm a bit surprised that:
* I have to specially handle `duckdb::Exception`
* `ThrowError` (in some cases at least) throws a `std::exception`, not a  `duckdb::Exception` one

Happy to improve if there's a better way?

JelteF (approved):

PR #252, opened 2024-10-07
Get rid of PostgresContextState

JelteF:
Since #247 is merged PostgresContextState is not actually used anymore. So this removes it completely. This will make a bunch of PRs easier to implement, specifically #241 and #206.


Tishj (approved):
Thanks, sounds good to me
It was only used to provide the PlannerInfo before, which we now get through another path / no longer need 👍

PR #253, opened 2024-10-07
Sync after each insert/update/delete of `duckdb.secrets` table

mkaruza:
If duckdb.secrets table is inserted, updated or deleted in same connection we would not sync changes for next query. Introduce sequence that keeps track of each change so we can compare with and sync if required.

### DIFF
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -12,25 +12,34 @@ namespace pgduckdb {
 
 class DuckDBManager {
 public:
-	static inline const DuckDBManager &
+	static inline DuckDBManager &
 	Get() {
 		static DuckDBManager instance;
 		return instance;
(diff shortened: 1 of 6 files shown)


wuputah:
This is tangential but while you are here, it would be nice to add a primary key to this table. I think an auto-increment column would be sorta nice it could probably be `(type, id)`.

This will make it safer / easier when either editing or deleting values from this table to have a WHERE clause that ensures you are only affecting a single row.

JelteF (approved):
Using a sequence this way definitely feels like a bit of a hack, but it should work pretty well. And if we run into issues with it we can always change this.

mkaruza:
> Using a sequence this way definitely feels like a bit of a hack, but it should work pretty well. And if we run into issues with it we can always change this.

Yes, but problem is multiple connections to same database. Modifying secrets table in one of connections should reflect to others as well.

PR #256, opened 2024-10-07
Fix memory leaks in custom executor

Y--:
When one calls `elog(ERROR,...` it `longjmp` out of the current stack leaving no chance for the C++ to be executed.

This PR is a first iteration toward what we want to rationalize (cf. https://github.com/duckdb/pg_duckdb/issues/93 ):
* keep a function with a pure C++ "flavor" (object creation, exceptions, etc.)
* call it in a new wrapper that copies out the error message and uses PG mechanism to throw the error

This fixes https://github.com/duckdb/pg_duckdb/issues/120, even though there still seem to be a small leak with DuckDB's `PrepareQuery` leftover (to be followed-up separately)

### DIFF
--- a/include/pgduckdb/pgduckdb_utils.hpp
+++ b/include/pgduckdb/pgduckdb_utils.hpp
@@ -5,6 +5,7 @@ extern "C" {
 }
 
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/error_data.hpp"
 
 #include <vector>
 #include <string>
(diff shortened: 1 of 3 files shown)


JelteF (approved):

JelteF (changes requested):
Tests are still failing it seems after merging with main.

Y--:
> Tests are still failing it seems after merging with main.

Yeah sorry should have kept this in draft - I'm still working on it; this is a simple revert

JelteF (approved):

PR #257, opened 2024-10-07
cleanup duckdb build

wuputah:
* dont build duckdb shell and other things we aren't using
* use correct flags that are supported by the Makefile
* start using ninja
* specify `OVERRIDE_GIT_DESCRIBE` via `DUCKDB_VERSION`

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -8,7 +8,7 @@ on:
 jobs:
   format:
     runs-on: ubuntu-24.04
-    name: 'Linting and formatting'
+    name: "Linting and formatting"
 
     steps:
       - name: Checkout pg_duckdb extension code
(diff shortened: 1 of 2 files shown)


wuputah:
I tested this locally. CI will not rebuild DuckDB in this case since the DuckDB sha did not change.

The motivation for this is running into an error building the DuckDB shell in a dpkg environment (to build an apt package). There's no particular reason to try to fix that when it can just be disabled.

wuputah:
Making this a draft for the moment until I'm 100% sure this actually fixes the dpkg issue.

Y-- (approved):
👍🏼

wuputah:
do we have opinions about when to use upper- vs lower-case var names in the Makefile?

wuputah:
With ninja and a fully hot ccache, a rebuild can be very fast now:

```
➜ time bash -c 'make clean-all && make -j16'
________________________________________________________
Executed in   21.58 secs    fish           external
   usr time   57.08 secs    0.09 millis   57.08 secs
   sys time    8.53 secs    1.55 millis    8.53 secs
```

But it did cause some issues in low-resource environments, so I needed a way to disable it.

JelteF (approved):
LGTM

> do we have opinions about when to use upper- vs lower-case var names in the Makefile?

UPPER_CASE seems to be the most common so let's do that

PR #259, opened 2024-10-08
Create a single way of creating a connection

JelteF:
Because of the changes made in #231 and #252 `DuckdbCreateConnection` isn't doing anything special anymore. This removes that function and replaces it with a `GetConnection` method on the `DuckDBManager`. This still returns a new connection every time it's called, but that should probably change in the future. Doing that requires a bit more complex changes to some logic, so for now this is just a refactoring.


mkaruza (approved):
Looks reasonable.

PR #260, opened 2024-10-08
Revert "cleanup duckdb build (#257)"

JelteF:
This reverts commit 7489afec0f018c6da90c8047781c6a8b30bceca7.

This commit broke our CI tests because it removed support for JSON. I think removal of JSON support was by accident. Let's revert for now, and re-open a new PR with a fixed build.

PR #261, opened 2024-10-08
Cleanup duckdb build with working extensions

wuputah:
This re-applies #257 and fixes the issue from #260.

* was missing a backslash
* while some vars were being picked up as arguments, some were not. 
  We now instead set them as env vars when calling `make`.


wuputah:
I manually deleted the `...-ninja` cache so this will be a clean rebuild without changing the cache key.

JelteF (approved):

PR #262, opened 2024-10-08
ci: improve gha ccache caching key

wuputah:
Separate the `ccache` cache into separate caches per job (i.e. PG version) to ensure the gha picks up the most recent cache for that version, rather than the overall most recent cache that was written.


wuputah:
So just based on the increased build time of trying a new cache key, it seems the logic of the gha was picking up the previous timestamp. This still litters our cache with old versions, but that's not causing any issues.

This PR now just separates the caching to be per job (i.e. per PG version).

JelteF (approved):

PR #264, opened 2024-10-09
Rewrite COPY ... TO command to match pgduckdb execution

mkaruza:
Currently COPY ... TO utility is broken due changes in maching table names inside duckb execution. Rewrite
complete query and parse manually each part of copy statement.

### DIFF
--- a/src/utility/copy.cpp
+++ b/src/utility/copy.cpp
@@ -23,30 +23,9 @@ extern "C" {
 #include "pgduckdb/vendor/pg_list.hpp"
 #include "pgduckdb/pgduckdb_utils.hpp"
 
-#include <functional>
-#include <iostream>
-#include <numeric>
-#include <string>
-#include <vector>
-
 static constexpr char s3_filename_prefix[] = "s3://";
 static constexpr char gcs_filename_prefix[] = "gs://";
 static constexpr char r2_filename_prefix[] = "r2://";
-
-typedef struct DuckdbCopyOptions {
-	bool csv_mode;
-	struct CSVOptions {
-		char *delimiter;
-		char *null_str;
-		int null_str_len;
-		bool include_header;
-		char *quote;
-		char *escape;
-		List *force_quote;
-		bool force_quote_all;
-		bool *force_quote_flags;
-	} csv_options;
-} DuckdbCopyOptions;
 
 /*
  * Returns the relation of the copy_stmt as a fully qualified DuckDB table reference. This is done
@@ -146,7 +125,6 @@ CreateCopyOptions(CopyStmt *copy_stmt, bool *options_valid) {
 		if (!first) {
 			options_string += ", ";
 		}
-
 		options_string += defel->defname;
 		if (defel->arg) {
 			options_string += " ";
@@ -173,18 +151,13 @@ CreateCopyOptions(CopyStmt *copy_stmt, bool *options_valid) {
 
 		first = false;
 	}
-
 	options_string += ");";
-
 	return options_string;
 }
 
 bool
-DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment *query_env, uint64 *processed,
-           bool *is_copy_to_cloud) {
+DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment *query_env, uint64 *processed) {
 	CopyStmt *copy_stmt = (CopyStmt *)pstmt->utilityStmt;
-
-	*is_copy_to_cloud = false;
 
 	if (!copy_stmt->filename) {
 		return false;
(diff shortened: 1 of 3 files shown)


src/utility/copy.cpp
```diff
@@ -3,30 +3,134 @@
...
+	if (copy_stmt->attlist) {
+		ListCell *lc;
+		relation_copy += "(";
+		bool first = true;
+		foreach (lc, copy_stmt->attlist) {
+			if (!first) {
+				relation_copy += ", ";
+			}
+			first = false;
+			relation_copy += strVal(lfirst(lc));
```
JelteF, line 125:
```suggestion
			relation_copy += quote_identifier(strVal(lfirst(lc)));
```

src/utility/copy.cpp
```diff
@@ -3,30 +3,134 @@
...
+		relation_copy += ") ";
+	}
+
+	*allowed = true;
+	return relation_copy;
+}
+
 bool
 DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment *query_env, uint64 *processed) {
 	CopyStmt *copy_stmt = (CopyStmt *)pstmt->utilityStmt;
```
JelteF, line 136: Not really related to this PR. But we should probably check `copy_stmt->filename` for NULL here.

src/utility/copy.cpp
```diff
@@ -43,8 +147,38 @@ DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment
...
+
+		raw_stmt = makeNode(RawStmt);
+		raw_stmt->stmt = copy_stmt->query;
+		raw_stmt->stmt_location = pstmt->stmt_location;
+		raw_stmt->stmt_len = pstmt->stmt_len;
+
+		rewritten = pg_analyze_and_rewrite_fixedparams(raw_stmt, query_string, NULL, 0, NULL);
+		query = linitial_node(Query, rewritten);
+		rewritten_query_string = duckdb::StringUtil::Format("COPY (%s) %s;", pgduckdb_pg_get_querydef(query, false),
+		                                                    query_string + token_start_offset);
```
JelteF, line 166: How about we include the filename instead of verbatim copy pasting the end of the string starting at TO (which we first have to find with your new FindQueryTokenOffset function)

```suggestion
		rewritten_query_string = duckdb::StringUtil::Format("COPY (%s) TO %s;", pgduckdb_pg_get_querydef(query, false),
		                                                    quote_literal_cstr(copy_stmt->filename);
```

src/utility/copy.cpp
```diff
@@ -3,30 +3,134 @@
...
+			token_offset = yylloc;
+			break;
+		}
+	}
+	scanner_finish(yyscanner);
+	return token_offset;
+}
+
+static std::string
+CreateRelationCopyString(ParseState *pstate, CopyStmt *copy_stmt, bool *allowed) {
```
JelteF, line 58: A little function comment would be helpful here. Something like this:
```suggestion
/* Returns the relation of the copy_stmt as a fully qualified DuckDB table reference. This is done including the column names if provided in the original copy_stmt, e.g. my_table(column1, column2). This also checks permissions on the table to see if the user is allowed to copy the data from this table */
static std::string
CreateRelationCopyString(ParseState *pstate, CopyStmt *copy_stmt, bool *allowed) {
```

JelteF (changes requested):

src/utility/copy.cpp
mkaruza: We can't use lexer on PG16/PG17 so relaying now quoted filename string to give us location of OPTIONS part of query.

src/utility/copy.cpp
```diff
@@ -147,7 +125,15 @@ DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment
 		return false;
 	}
 
-	int token_start_offset = FindQueryTokenOffset(query_string, TO);
+	duckdb::string query_string_copy = std::string(query_string);
+
+	/* Quote filename so we can match it with query string*/
+	auto filename_quoted = quote_literal_cstr(copy_stmt->filename);
+	auto filename_quoted_end_pos = query_string_copy.find(filename_quoted) + strlen(filename_quoted);
```
JelteF, line 132: I'm not a fan of this. e.g. It'll break if if you have a query that contains this quoted filename. It might be okay for the 0.1.0, but I'd be much happier if the options string was built from the `List` of `DefElem`s in `copy_stmt->options`.

JelteF (approved):

src/utility/copy.cpp
```diff
@@ -2,49 +2,465 @@
...
+
+		if (strcmp(defel->defname, "format") == 0) {
+			char *fmt = defGetString(defel);
+
+			if (format_specified) {
+				ReportConflictingCopyOption(defel, options_valid);
+				break;
+			}
+			if (strcmp(fmt, "csv") == 0) {
+				duckdb_copy_options.csv_mode = true;
```
JelteF, line 158: Don't we want to be able to write parquet files too?

mkaruza:
@JelteF with your simplification changes this works as expected.

JelteF (approved):

PR #267, opened 2024-10-09
Add MotherDuck support

JelteF:
This adds initial MotherDuck support. It supports syncing of table and schema metadata from MotherDuck to Postgres. Every second a background worker will check if a MotherDuck catalog has changed to a new version. If there was a catalog version update, then the changes are synced to Postgres by dropping and recreating all MotherDuck shell tables related to that MotherDuck catalog.

The schemas from the default MotherDuck database is merged with schemas with the same name from Postgres. Schemas from other databases are encoded in this way:
1. If the db is `awesomedb` schema is `main`, then the postgres schema name is `ddb$awesomedb`
2. If the db is `awesomedb` schema is `some_other_schema`, then the postgres schema name is `ddb$awesomedb$some_other_schema`  

To enable MotherDuck support all that's needed is adding the following to your Postgres config:
```
shared_preload_libraries = 'pg_duckdb'
duckdb.background_worker_db = '<the db in which you do CREATE EXTENSION>'
duckdb.motherduck_token = '<your token>`
```

This also fixes a problem with temporary duckdb tables that were not explicitely dropped before the backend was terminated. The oids of those tables would stay in the `duckdb.tables` table indefinitely. To fix that this now tracks temporary table oids in backend-local in-memory set. 

TODO:
- [x] Check MotherDuck catalog version to see if Postgres catalog needs to be updated, and do nothing if version is the same.
- [ ] ~~Only update tables that have changed since the last sync (optional?)~~ (creating separate issue for 0.2.0 for this)
- [ ] ~~Create the tables as the duckdb.motherduck_postgres_user~~ (I'm going to make a separate PR for this)
- [x] Change to use eager committing of transactions, so Postgres locks are not held for long periods of time.
- [x] Disallow GRANT on table
- [x] Disallow creation of `ddb$` schemas
- [x] Check that no `ddb$` schemas exist in extension creation script
- [x] Automatically drop all `ddb$` schemas when dropping the extension
- [x] Make the background worker handle create+drop extension sensibly

### DIFF
--- a/include/pgduckdb/pgduckdb_background_worker.hpp
+++ b/include/pgduckdb/pgduckdb_background_worker.hpp
@@ -1,0 +1,21 @@
+#pragma once
+
+#include "duckdb.hpp"
+#include "duckdb/parser/parsed_data/create_info.hpp"
+#include "duckdb/parser/parsed_data/create_info.hpp"
+#include "duckdb/common/unordered_set.hpp"
+#include "duckdb/parser/column_definition.hpp"
+#include "duckdb/parser/constraint.hpp"
+#include "duckdb/parser/statement/select_statement.hpp"
+#include "duckdb/catalog/catalog_entry/column_dependency_manager.hpp"
+#include "duckdb/parser/column_list.hpp"
+#include "duckdb/parser/parsed_data/create_table_info.hpp"
+
+void DuckdbInitBackgroundWorker(void);
+
+namespace pgduckdb {
+
+void SyncMotherDuckCatalogsWithPg(bool drop_with_cascade);
+extern bool doing_motherduck_sync;
+
+} // namespace pgduckdb
--- a/include/pgduckdb/pgduckdb_background_worker.hpp
+++ b/include/pgduckdb/pgduckdb_background_worker.hpp
@@ -11,11 +11,17 @@
 #include "duckdb/parser/column_list.hpp"
 #include "duckdb/parser/parsed_data/create_table_info.hpp"
 
+extern "C" {
+#include "postgres.h"
+}
+
 void DuckdbInitBackgroundWorker(void);
 
 namespace pgduckdb {
 
 void SyncMotherDuckCatalogsWithPg(bool drop_with_cascade);
 extern bool doing_motherduck_sync;
+extern char *current_motherduck_database_name;
+extern char *current_motherduck_catalog_version;
 
 } // namespace pgduckdb
--- a/include/pgduckdb/pgduckdb_background_worker.hpp
+++ b/include/pgduckdb/pgduckdb_background_worker.hpp
@@ -21,7 +21,7 @@ namespace pgduckdb {
 
 void SyncMotherDuckCatalogsWithPg(bool drop_with_cascade);
 extern bool doing_motherduck_sync;
-extern char *current_motherduck_database_name;
+extern char *current_duckdb_database_name;
 extern char *current_motherduck_catalog_version;
 
 } // namespace pgduckdb
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -42,7 +42,7 @@ private:
 
 	int secret_table_num_rows;
 	int secret_table_current_seq;
-	duckdb::unique_ptr<duckdb::DuckDB> database;
+	duckdb::DuckDB *database;
 	std::string default_dbname;
 };
 
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -23,6 +23,11 @@ public:
 
 	static duckdb::unique_ptr<duckdb::Connection> CreateConnection();
 
+	inline const std::string &
+	GetDefaultDBName() const {
+		return default_dbname;
+	}
+
 private:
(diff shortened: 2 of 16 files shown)


src/pgduckdb_ddl.cpp
```diff
@@ -176,65 +275,143 @@ duckdb_drop_table_trigger(PG_FUNCTION_ARGS) {
...
+			connection = pgduckdb::DuckDBManager::CreateConnection();
+			pgduckdb::DuckDBQueryOrThrow(*connection, "BEGIN TRANSACTION");
+		}
+		char *table_name = SPI_getvalue(tuple, SPI_tuptable->tupdesc, 2);
+		pgduckdb::DuckDBQueryOrThrow(*connection,
+		                             std::string("DROP TABLE pg_temp.main.") + quote_identifier(table_name));
+		temporary_duckdb_tables.erase(relid);
 	}
 
+	AtEOXact_GUC(false, save_nestlevel);
```
Y--, line 409: Is that what restore the GUC (search_path/duckdb.execution)? If so we might have an issue when an exception is thrown no? (either C++ or PG)

Y-- (approved):
Looks really good, I've added a bunch of comments - happy to discuss online

include/pgduckdb/pgduckdb_background_worker.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+extern "C" {
+#include "postgres.h"
+}
+
+void DuckdbInitBackgroundWorker(void);
+
+namespace pgduckdb {
+
+void SyncMotherDuckCatalogsWithPg(bool drop_with_cascade);
```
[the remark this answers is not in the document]
JelteF: Since it uses `__md_local_databases_metadata()` and the motherduck specific catolog version stuff, I think for now it makes sense to keep MotherDuck in the name. If/when we change it to support plain DuckDB I think it makes sense to rename it.

src/pgduckdb_types.cpp
```diff
@@ -627,6 +632,20 @@ GetPostgresDuckDBType(duckdb::LogicalType type) {
 	}
 }
 
+int32
+GetPostgresDuckDBTypemod(duckdb::LogicalType type) {
```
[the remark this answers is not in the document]
JelteF: It's only used by the background worker. Right now we don't have tests for the syncing. I'm planning to improve that in the next release.

src/pgduckdb_ddl.cpp
JelteF: At transaction end the postgres transaction management will automatically rollback these changes too, even on failure. This call is only necessary on success, because we want any further code to use the old values for these GUCs again.

src/pgduckdb_background_worker.cpp
```diff
@@ -0,0 +1,583 @@
...
+
+static std::string
+DropPgTableString(const char *postgres_schema_name, const char *table_name, bool with_cascade) {
+	std::string ret = "";
+	ret += "DROP TABLE ";
+	ret += duckdb::KeywordHelper::WriteQuoted(postgres_schema_name, '"');
+	ret += ".";
+	ret += duckdb::KeywordHelper::WriteQuoted(table_name, '"');
+	ret += with_cascade ? " CASCADE; " : "; ";
+	return ret;
```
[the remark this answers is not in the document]
JelteF: I'll leave this as is for now. When deciding on if we're going to write this in C or C++ style we can replace it to the use either `StringInfo` or `std::ostringstream` depending on that decision.

include/pgduckdb/pgduckdb_duckdb.hpp
```diff
@@ -41,7 +48,8 @@ class DuckDBManager {
 
 	int secret_table_num_rows;
 	int secret_table_current_seq;
-	duckdb::unique_ptr<duckdb::DuckDB> database;
+	duckdb::DuckDB *database;
```
Y--, line 51: Can we add a FIXME to restore the smart pointer here?

src/pgduckdb_duckdb.cpp
```diff
@@ -42,7 +53,21 @@ DuckDBManager::Initialize() {
...
+
+	if (pgduckdb::IsMotherDuckEnabled()) {
+		/*
+		 * Workaround for MotherDuck catalog sync that turns off automatically,
+		 * in case of no queries being sent to MotherDuck. Since the background
+		 * worker never sends any query to MotherDuck we need to turn this off.
+		 * So we set the timeout to an arbitrary very large value.
+		 */
+		pgduckdb::DuckDBQueryOrThrow(context,
+		                             "SET motherduck_background_catalog_refresh_inactivity_timeout='99 years'");
```
Y--, line 69: The good old 99y bug 😆

Y-- (approved):
LGTM, tested on my end successfully!

include/pgduckdb/pgduckdb_duckdb.hpp
JelteF: Done

src/pgduckdb_duckdb.cpp
JelteF: Yeah, I'm pretty certain no-one runs their postgres server without restart for that long. And even they do, it's not going to be my problem anymore.

Y-- (approved):

PR #268, opened 2024-10-10
GetConnection should check secret table changes

mkaruza:
We don't need use anymore `GetDatabase()` which contained logic to update secrets based on sequence number. Move thisi logic now in `GetConnection` which is currently used in code.


JelteF (approved):

JelteF:
Merged quickly to fix CI

PR #272, opened 2024-10-11
Factorize simple DuckDB queries

Y--:
* change `DuckDBManager::GetConnection()` member into a static `CreateConnection()`
* introduce 3 flavor of `DuckDBQueryOrThrow` to simplify recurring pattern

### DIFF
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -15,13 +15,19 @@ public:
 	static inline DuckDBManager &
 	Get() {
 		static DuckDBManager instance;
+		if (!instance.database) {
+			instance.Initialize();
+		}
 		return instance;
 	}
 
-	duckdb::unique_ptr<duckdb::Connection> GetConnection();
+	static duckdb::unique_ptr<duckdb::Connection> CreateConnection();
 
 private:
 	DuckDBManager();
+
+	void Initialize();
+
 	void InitializeDatabase();
 	bool CheckSecretsSeq();
 	void LoadSecrets(duckdb::ClientContext &);
--- a/include/pgduckdb/pgduckdb_utils.hpp
+++ b/include/pgduckdb/pgduckdb_utils.hpp
@@ -1,11 +1,12 @@
 #pragma once
 
+#include "duckdb/common/exception.hpp"
+#include "duckdb/common/error_data.hpp"
+#include "pgduckdb/pgduckdb_duckdb.hpp"
+
 extern "C" {
 #include "postgres.h"
 }
-
-#include "duckdb/common/exception.hpp"
-#include "duckdb/common/error_data.hpp"
 
 #include <vector>
 #include <string>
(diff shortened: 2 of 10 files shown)


include/pgduckdb/pgduckdb_duckdb.hpp
```diff
@@ -15,13 +15,24 @@ class DuckDBManager {
 	static inline DuckDBManager &
 	Get() {
 		static DuckDBManager instance;
+		if (!instance.database) {
+			instance.Initialize();
```
Y--, line 19: Perform initialization out of constructor so we don't throw during object creation.

include/pgduckdb/pgduckdb_utils.hpp
```diff
@@ -103,4 +104,10 @@ DuckDBFunctionGuard(FuncType duckdb_function, const char* function_name, FuncArg
 
 std::string CreateOrGetDirectoryPath(std::string directory_name);
 
+duckdb::unique_ptr<duckdb::QueryResult> DuckDBQueryOrThrow(duckdb::ClientContext &context, const std::string &query);
+
+duckdb::unique_ptr<duckdb::QueryResult> DuckDBQueryOrThrow(duckdb::Connection &connection, const std::string &query);
+
+duckdb::unique_ptr<duckdb::QueryResult> DuckDBQueryOrThrow(const std::string &query);
+
```
JelteF, line 112: I think it'd be good to have `const char*` versions of this too. Generally when doing calling such a function we'll be in a Postgres C context.

src/pgduckdb_ddl.cpp
```diff
@@ -160,15 +158,7 @@ duckdb_create_table_trigger(PG_FUNCTION_ARGS) {
...
-	std::string query_string(pgduckdb_get_tabledef(relid));
-
-	auto connection = pgduckdb::DuckDBManager::Get().GetConnection();
-	auto &context = *connection->context;
-	auto result = context.Query(query_string, false);
-	if (result->HasError()) {
-		elog(ERROR, "(PGDuckDB/duckdb_create_table_trigger) Could not create table: %s", result->GetError().c_str());
-	}
-
+	pgduckdb::DuckDBQueryOrThrow(pgduckdb_get_tabledef(relid));
```
JelteF, line 161: It seems incorrect to go from `elog` to `throw` here. This will cause a Postgres crash currently.

include/pgduckdb/pgduckdb_utils.hpp
Y--: Yeah I'm coming with a dedicated PR that rationalize this.

src/pgduckdb_ddl.cpp
Y--: Ah yes - I extracted it from my PR that wraps all functions with try/catch but didn't pay attention here. Will review other places too.

include/pgduckdb/pgduckdb_utils.hpp
Y--: I've actually made so that they always throw PG errors for now (which was the case already)

src/pgduckdb_ddl.cpp
Y--: I've made these function "throw" PG errors.

JelteF (approved):

Issue #209, opened 2024-09-24
`enable_extension` function missing


rakesh-tmdc:
It is mentioned that to `Enable the DuckDB Iceberg extension using SELECT duckdb.enable_extension('iceberg') and read Iceberg files with iceberg_scan.` but when i do this i getting error that is - 

`
postgres=# SELECT duckdb.enable_extension('iceberg');
ERROR:  function duckdb.enable_extension(unknown) does not exist
LINE 1: SELECT duckdb.enable_extension('iceberg');
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
`

This shows that the `enable_extension` function is not available.

JelteF:
It's a mistake in the readme (I created a PR to fix that). You should use `duckdb.install_extension` instead.

rakesh-tmdc:
Hey @JelteF Thanks for the update. I did try that below are the steps - 
 
`
postgres=# DROP EXTENSION IF EXISTS pg_duckdb; CREATE EXTENSION pg_duckdb; SET duckdb.execution TO true;SHOW duckdb.execution;


NOTICE:  extension "pg_duckdb" does not exist, skipping
DROP EXTENSION
WARNING:  To actually execute queries using DuckDB you need to run "SET duckdb.execution TO true;"
CREATE EXTENSION
SET
 duckdb.execution 
------------------
 on
(1 row)
`

`
postgres=# SELECT duckdb.install_extension('iceberg');

 install_extension 
-------------------
 t
(1 row)
`

`
postgres=# SELECT count(*) FROM iceberg_scan('s3a://****/sanity/iceberg/tablea') as (customerid TEXT);
WARNING:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Catalog Error: Table Function with name "iceberg_scan" is not in the catalog, but it exists in the iceberg extension.

Please try installing and loading the iceberg extension:
INSTALL iceberg;
LOAD iceberg;


WARNING:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Catalog Error: Table Function with name "iceberg_scan" is not in the catalog, but it exists in the iceberg extension.

Please try installing and loading the iceberg extension:
INSTALL iceberg;
LOAD iceberg;


ERROR:  Function `iceberg_scan(TEXT)` only works with Duckdb execution.
CONTEXT:  PL/pgSQL function iceberg_scan(text) line 3 at RAISE

`


[...]

PR #277, opened 2024-10-13
Improve error when read_xxx functions fail

### LINKED ISSUE #106: Misleading error when using `read_xxx` functions with invalid URLs
I get ```ERROR:  Function `read_csv(TEXT)` only works with Duckdb execution.``` when using `read_csv` (along with the actual error as a warning):

```
y=# SELECT * from read_csv('https://www.data.gouv.fr/fr/datasets/r/6d186965-f41b-41f3-9b23-88241cc6890c') as (n int);
WARNING:  (DuckDB) HTTP Error: Unable to connect to URL "https://www.data.gouv.fr/fr/datasets/r/6d186965-f41b-41f3-9b23-88241cc6890c": 404 (Not Found)
ERROR:  Function `read_csv(TEXT)` only works with Duckdb execution.
CONTEXT:  PL/pgSQL function read_csv(text) line 3 at RAISE
```

I would expect the `WARNING` to be the `ERROR` and the `ERROR` to not show.

JelteF:
Previously the error would be shown as a WARNING and then the query would be run as a fallback using the Postgres executor, which would fail because the DuckDB `read_csv`/`read_parquet` functions cannot be executed by the Postgres executor. This is quite confusing as many people only look at the last ERROR, and not the warnings before.

This is also relevant for MotherDuck support, since the MotherDuck backed tables can only be read using the DuckDB executor.

Fixes #106
Related to #209 (the error there looks terribly confusing)

### DIFF
--- a/include/pgduckdb/pgduckdb_planner.hpp
+++ b/include/pgduckdb/pgduckdb_planner.hpp
@@ -11,6 +11,6 @@ extern "C" {
 
 extern bool duckdb_explain_analyze;
 
-PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options);
+PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options, int elevel);
 std::tuple<duckdb::unique_ptr<duckdb::PreparedStatement>, duckdb::unique_ptr<duckdb::Connection>>
 DuckdbPrepare(const Query *query);
--- a/include/pgduckdb/pgduckdb_planner.hpp
+++ b/include/pgduckdb/pgduckdb_planner.hpp
@@ -11,6 +11,6 @@ extern "C" {
 
 extern bool duckdb_explain_analyze;
 
-PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options, int elevel);
+PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options, bool throw_error);
 std::tuple<duckdb::unique_ptr<duckdb::PreparedStatement>, duckdb::unique_ptr<duckdb::Connection>>
 DuckdbPrepare(const Query *query);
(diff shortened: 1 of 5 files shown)


include/pgduckdb/pgduckdb_planner.hpp
```diff
@@ -11,6 +11,6 @@ extern "C" {
 
 extern bool duckdb_explain_analyze;
 
-PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options);
+PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options, int elevel);
```
Y--, line 14: nit
```suggestion
PlannedStmt *DuckdbPlanNode(Query *parse, int cursor_options, int elevel_on_error);
```

src/pgduckdb_hooks.cpp
```diff
@@ -166,17 +166,13 @@ IsAllowedStatement(Query *query, bool throw_error = false) {
...
-				return duckdbPlan;
-			}
-		}
-
 		if (NeedsDuckdbExecution(parse)) {
 			IsAllowedStatement(parse, true);
 
-			PlannedStmt *duckdbPlan = DuckdbPlanNode(parse, cursor_options);
+			PlannedStmt *duckdbPlan = DuckdbPlanNode(parse, cursor_options, ERROR);
+			return duckdbPlan;
```
Y--, line 173: nit
```suggestion
			return DuckdbPlanNode(parse, cursor_options, ERROR);
```

Y-- (approved):
LGTM

include/pgduckdb/pgduckdb_planner.hpp
JelteF: I changed this to `bool throw_error` for consistency with `IsAllowedStatement`.

Y-- (approved):

Issue #279, opened 2024-10-14
"EXPLAIN" statement Segmentation fault


Leo-XM-Zeng:
**Current Behavior**
```plpgsql
postgres=# set duckdb.execution = true;
SET
postgres=# create table ta(a int);
CREATE TABLE
postgres=# explain SELECT count(*) FROM ta;
2024-10-14 16:40:30.331 CST [1621] WARNING:  01000: relcache reference leak: relation "ta" not closed
2024-10-14 16:40:30.331 CST [1621] LOCATION:  PrintRelCacheLeakWarning, resowner.c:1203
WARNING:  relcache reference leak: relation "ta" not closed
                         QUERY PLAN                         
------------------------------------------------------------
 Custom Scan (DuckDBScan)  (cost=0.00..0.00 rows=0 width=0)
   DuckDB Execution Plan: 
 
 ┌───────────────────────────┐
 │    UNGROUPED_AGGREGATE    │
 │    ────────────────────   │
 │        Aggregates:        │
 │        count_star()       │
 └─────────────┬─────────────┘
 ┌─────────────┴─────────────┐
 │     POSTGRES_SEQ_SCAN     │
 │    ────────────────────   │
 │         Function:         │
 │     POSTGRES_SEQ_SCAN     │
 │                           │
 │         ~2550 Rows        │
 └───────────────────────────┘
 
 
(19 rows)

postgres=# \q
[postgres@halo-centos8 ~]$ 2024-10-14 16:40:33.891 CST [1607] LOG:  00000: server process (PID 1621) was terminated by signal 11: Segmentation fault
```

Leo-XM-Zeng:
"explain" statement has a resource management problem, but I can't find out how to fix it

JelteF:
Hmm, that's not good. I can reproduce this, although for me it is throwing an assert failure because I compile postgres with asserts enabled. 

The segfault/assert failure is fixed/worked around for me in #267. I presume because we don't call the destructor of DuckDB anymore at process exit, which seems to be the reason why it's crashing. It seems we're calling `RelationClose` too late in this case for some reason.

JelteF:
We should definitely fix the resource leak too though.

Leo-XM-Zeng:
Maybe there's a "relation_close" function call missing somewhere

mkaruza:
[...]

PR #281, opened 2024-10-14
Remove Tx from TxManager  on rollback

Y--:
Fixes https://github.com/duckdb/pg_duckdb/issues/279

Also adds a test that removes `DuckDB` singleton and re-creates it: this is when the crash was happening.

And fully consume the explain output stream from DuckDB

### DIFF
--- a/src/catalog/pgduckdb_transaction.cpp
+++ b/src/catalog/pgduckdb_transaction.cpp
@@ -82,8 +82,10 @@ SchemaItems::GetTable(const string &entry_name) {
 	info.table = entry_name;
 	Cardinality cardinality = 1;
 	if (!PostgresTable::SetTableInfo(info, rel)) {
+		RelationClose(rel);
 		return nullptr;
 	}
+
 	cardinality = PostgresTable::GetTableCardinality(rel);
 	table = make_uniq<PostgresHeapTable>(catalog, *schema, info, rel, cardinality, snapshot);
 	tables[entry_name] = std::move(table);
--- a/src/catalog/pgduckdb_transaction.cpp
+++ b/src/catalog/pgduckdb_transaction.cpp
@@ -90,18 +90,22 @@ SchemaItems::GetTable(const string &entry_name) {
 	return tables[entry_name].get();
 }
 
+optional_ptr<CatalogEntry> SchemaItems::GetSchema() const {
+	return schema.get();
+}
+
 optional_ptr<CatalogEntry>
 PostgresTransaction::GetSchema(const string &name) {
 	auto it = schemas.find(name);
 	if (it != schemas.end()) {
-		return it->second.schema.get();
+		return it->second.GetSchema();
 	}
 
 	CreateSchemaInfo create_schema;
 	create_schema.schema = name;
 	auto pg_schema = make_uniq<PostgresSchema>(catalog, create_schema, snapshot);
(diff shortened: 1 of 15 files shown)


src/catalog/pgduckdb_transaction_manager.cpp
```diff
@@ -26,7 +26,8 @@ PostgresTransactionManager::CommitTransaction(ClientContext &context, Transactio
 
 void
 PostgresTransactionManager::RollbackTransaction(Transaction &transaction) {
-	return;
+	lock_guard<mutex> l(transaction_lock);
+	transactions.erase(transaction);
```
Y--, line 30: This is the fix

src/catalog/pgduckdb_transaction.cpp
```diff
@@ -82,26 +82,32 @@ SchemaItems::GetTable(const string &entry_name) {
 	info.table = entry_name;
 	Cardinality cardinality = 1;
 	if (!PostgresTable::SetTableInfo(info, rel)) {
+		RelationClose(rel);
```
Y--, line 85: Here too we would leak `rel` when returning

src/pgduckdb_hooks.cpp
```diff
@@ -37,14 +37,14 @@ IsCatalogTable(List *tables) {
 				return true;
 			}
 		}
+
 		if (table->relid) {
 			auto rel = RelationIdGetRelation(table->relid);
 			auto namespace_oid = RelationGetNamespace(rel);
+			RelationClose(rel);
```
Y--, line 44: Minor cleanup - move it higher since we were closing it in both branches below

src/catalog/pgduckdb_transaction.cpp
JelteF: Afaict that function can (surprisingly) only return true. So I guess this was not the cause.

Y--: Yeah it wasn't the root cause indeed. And I didn't look at `!PostgresTable::SetTableInfo` if it only returns true, then I'll remove the branch :-)

JelteF: > if it only returns true, then I'll remove the branch :-)

Sounds good let's make it a `void` function indeed.

Y--: Yep, exactly what I did :-)

test/regression/expected/duckdb_recycle.out
```diff
@@ -0,0 +1,57 @@
...
+ │        Aggregates:        │
+ │        count_star()       │
+ └─────────────┬─────────────┘
+ ┌─────────────┴─────────────┐
+ │     POSTGRES_SEQ_SCAN     │
+ │    ────────────────────   │
+ │         Function:         │
+ │     POSTGRES_SEQ_SCAN     │
+ │                           │
+ │         ~2550 Rows        │
```
JelteF, line 21: I'm a bit worried this line with the cardinality is going to flaky. I'd be less worried if you added a `pycheck` test instead in `test_explain.py`.

JelteF (approved):

test/regression/expected/duckdb_recycle.out
Y--: As discussed offline - I wasn't able to reproduce the issue with the pycheck. So keeping these for now

Issue #101, opened 2024-08-07
Extended list type support


Mytherin:
We should add support for more list types (e.g. floating point lists, etc).

JelteF:
@Tishj I assume you didn't start on this yet

mkaruza:
@JelteF we have used this ticket just as indicator and probably at this point doesn't make sense that it is still open. IMO, for any future data type support we should open different tickets.

mkaruza:
Closing this issue. We should probably track new supported types by creating independent tickets.

PR #282, opened 2024-10-14
support extended list types

Reminiscent:
Related to https://github.com/duckdb/pg_duckdb/issues/101

This pull request adds more array types. The changes include:

1. Extended Array Type Support:
  * Added support for additional array types including CHAR, SMALLINT, VARCHAR, TIMESTAMP, FLOAT4, FLOAT8, NUMERIC, UUID and JSON.
  * Implemented support for both single-dimension and two-dimensional arrays for these types.

2. Improved Type Conversion:
  * Refactored the type conversion logic to handle the new array types efficiently.
  * Implemented specialized conversion functions for each supported data type to ensure accurate data representation between PostgreSQL and DuckDB.

3. Enhanced PostgresArrayAppendState:
  * Modified to handle edge cases, such as empty arrays and NULL values, more robustly.

4. Updated Test Suite:
  * Expanded the test suite to cover the newly supported array types and dimensions.

5. Code Refactoring:
  * Introduced a more generic approach to array handling, reducing code duplication and improving maintainability.
Implemented the PostgresOIDMapping structure to streamline type-specific operations.


JelteF:
Thanks a lot! This looks great. We're currently preparing for a 0.1.0 release, and given the size of the PR we probably don't want to squeeze it in before that release (we're going to be mostly working on stabilizing what we have soon). But this looks like a good candidate to merge early for the release cycle after that one.

test/regression/schedule
```diff
@@ -17,4 +17,4 @@ test: query_filter
 test: temporary_tables
 test: altered_tables
 test: transaction_errors
-test: secrets
+test: secrets
```
JelteF, line 20: Can you install https://editorconfig.org/ and save this file and your other files again. Then this diff should disappear, because a newline will be included at the end of the file.

src/pgduckdb_types.cpp
```diff
@@ -30,94 +30,72 @@ extern "C" {
```
JelteF, line 30: The file mode was changed to be executable (i.e. from 644 to 755). This seems by accident. Can you undo that?

Same issue for `array_type_support.sql`.

JelteF (commented):
Some quick cursory review. I'll look at the actual code later.

Tests are failing on PG17. Seemingly because of a changed error message string. If you cannot change the test to create a consistent error message across postgres versions, then it's probably easiest to create a Python test in `test/pycheck` directory. That way you can change the expected error message based on the postgres version.

src/pgduckdb_types.cpp
```diff
@@ -128,7 +106,6 @@ ConvertNumeric(T value, idx_t scale) {
 	result.dscale = scale;
 	auto &weight = result.weight;
 	auto &ndigits = result.ndigits;
-
```
JelteF, line 131: Let's keep these empty new lines. I think they make the code easier to read.

src/pgduckdb_types.cpp
```diff
@@ -30,94 +30,72 @@ extern "C" {
...
-};
+static Datum
+ConvertBoolDatum(const duckdb::Value &value) {
+	return value.GetValue<bool>();
+}
 
-template <int32_t OID>
-struct PostgresIntegerOIDMapping {};
+static Datum
+ConvertCharDatum(const duckdb::Value &value) {
```
JelteF, line 39: I very much like these new functions. I think taking that refactor on its own would make the code already a lot better.

src/pgduckdb_types.cpp
```diff
@@ -173,24 +148,209 @@ ConvertNumeric(T value, idx_t scale) {
...
+	}
+	case duckdb::PhysicalType::INT32: {
+		numeric_var = ConvertNumeric<int32_t>(value.GetValueUnsafe<int32_t>(), scale);
+		break;
+	}
+	case duckdb::PhysicalType::INT64: {
+		numeric_var = ConvertNumeric<int64_t>(value.GetValueUnsafe<int64_t>(), scale);
+		break;
+	}
+	case duckdb::PhysicalType::UINT64: {
```
JelteF, line 191: I know you didn't write this code, just moved it around. But reading this now it makes me wonder. Why is this the only unsigned integer type that we check for the physical type? @mkaruza do you know?

src/pgduckdb_types.cpp
```diff
@@ -236,6 +396,16 @@ struct PostgresArrayAppendState {
 		auto &values = duckdb::ListValue::GetChildren(value);
 		idx_t to_append = values.size();
 
+        if (to_append == 0 && dimension < number_of_dimensions-1) {
```
JelteF, line 399: This shouldn't be using spaces for indentation. I'm surprised our clang-format check is not failing on this... That's something we should probably look into. Installing editorconfig should help with this too. 

There are a bunch more places in the PR where space indentation is used, instead of tabs.

src/pgduckdb_types.cpp
```diff
@@ -173,24 +148,209 @@ ConvertNumeric(T value, idx_t scale) {
...
+	char typalign;
+};
+
+PGTypeInfo
+GetPGTypeInfo(Oid typid) {
+	PGTypeInfo info;
+	HeapTuple tp;
+	Form_pg_type typtup;
+
+	tp = SearchSysCache1(TYPEOID, ObjectIdGetDatum(typid));
```
JelteF, line 239: This is one thing we shouldn't be doing. Because to safely call this function we need to take a global lock, which is way to slow in the hot codepath where you're currently using it. There are two options I think:

1. Make sure we call this `SearchSysCache1` function only once (while taking the lock) and cache its result somewhere ourselves which we can access in a thread-safe way.
2. Hardcode the values for the supported types for now.

src/pgduckdb_types.cpp
JelteF, line 399: This seems to be new code. Could you explain in a comment what edge case this is handling? Because that is not super clear from me from just the code. Also I'm curious why this was not necessary before.

src/pgduckdb_types.cpp
```diff
@@ -317,8 +485,10 @@ ConvertDuckToPostgresArray(TupleTableSlot *slot, duckdb::Value &value, idx_t col
 	auto arr = OP::ConstructArray(datums, nulls, number_of_dimensions, dimensions, lower_bounds);
 
 	// Free allocated memory
-	pfree(datums);
-	pfree(nulls);
+    if (append_state.expected_values > 0) {
+        pfree(datums);
+        pfree(nulls);
+    }
```
JelteF, line 491: I don't think this change should be necessary.

src/pgduckdb_types.cpp
```diff
@@ -488,79 +617,102 @@ numeric_typmod_scale(int32 typmod) {
...
-	case VARCHAROID:
-		return duckdb::LogicalTypeId::VARCHAR;
+	case TEXTARRAYOID:
+	case VARCHARARRAYOID:
+	case VARCHAROID: {
+		base_type = duckdb::LogicalTypeId::VARCHAR;
+		break;
+	}
 	case DATEOID:
 		return duckdb::LogicalTypeId::DATE;
```
JelteF, line 662: This seems to be missing `DATEARRAYOID` by accident.
```suggestion
	case DATEOID:
	case DATEARRAYOID: {
		base_type = duckdb::LogicalTypeId::DATE;
		break;
	}
```

src/pgduckdb_types.cpp
```diff
@@ -488,79 +617,102 @@ numeric_typmod_scale(int32 typmod) {
...
+		break;
+	}
 	case BPCHAROID:
+	case BPCHARARRAYOID:
 	case TEXTOID:
-	case VARCHAROID:
-		return duckdb::LogicalTypeId::VARCHAR;
+	case TEXTARRAYOID:
+	case VARCHARARRAYOID:
+	case VARCHAROID: {
```
JelteF, line 657: nit: for consistency, let's flip these arround:
```suggestion
	case VARCHAROID:
	case VARCHARARRAYOID: {
```

JelteF (changes requested):
Overall this seems like a good change. I quite like the refactoring that is done. The only real problem with this implementation is the current usage of `SearchSysCache1`, which is not thread-safe.

src/pgduckdb_types.cpp
Reminiscent: Fixed. But in `.clang-format`, alignment doesn’t have to use tabs, specifically `UseTab: ForIndentation IndentWidth: 4 TabWidth: 4`, where `UseTab` is not `always`. However, in `.editorconfig`, `[*.{c,h,cpp,hpp}] indent_style = tab`. Are these two configurations conflicting? Which one should be used?

src/pgduckdb_types.cpp
Reminiscent: Yes, I use the second way to solve this problem based on the performance.

src/pgduckdb_types.cpp
Reminiscent: I also think the `pfree` itself can handle this. But when I remove the if condition, and run the tests. I'll get the `CREATE TABLE char_array_2d(a CHAR(1)[][]);
INSERT INTO char_array_2d VALUES
    ('{{"a","b"},{"c","d"}}'),
    ('{{"e","f","g"},{"h","i","j"}}'),
    (NULL),
    ('{{"k","l"},{"m",NULL}}'),
    ('{}');
SELECT * FROM char_array_2d;
connection to server was lost`. That's why I add this.

src/pgduckdb_types.cpp
Reminiscent: Updated. Use another way to solve this issues.
The origin issue is when we insert empty array into multi-dimension array. The `dimension[1]` to `dimension[number_of_dimensions-1]` can not be set because of the logic of `AppendValueAtDimension` function. So we use the `dimension` to call the function `ConstructArray`. It will cause overflow. The reason why it not necessary before is lack of the test cases. When you insert empty array into multi-dimension array. It will happen.

Reminiscent:
@JelteF Thanks for your review! All the comments(except the test failed in PG17) are address. PTAL

JelteF:
Regarding the test failure in PG17, which happens due to a different error message in that PG version. The approach we take so far is to move this test to a Python test in the `test/pycheck` directory.

Reminiscent:
> Regarding the test failure in PG17, which happens due to a different error message in that PG version. The approach we take so far for cases like this is to move this test to a Python test in the `test/pycheck` directory.

I known your meaning. But find the test cases can pass now. Just wonder will you run the regression tests in different postgres version when you develop some new features?

JelteF:
> But find the test cases can pass now.

Ah if it passes now you can disregard my comment.

src/pgduckdb_types.cpp
```diff
@@ -192,6 +169,269 @@ ConvertNumeric(T value, idx_t scale) {
...
+
+// Specializations for each type
+// BOOL type
+template <>
+struct PostgresTypeTraits<BOOLOID> {
+	static constexpr int16_t typlen = 1;
+	static constexpr bool typbyval = true;
+	static constexpr char typalign = 'c';
+
+	static Datum
```
JelteF, line 245: Let's mark all these `ToDatum` functions as `inline`. They are so tiny that there's no need for introducing calling overhead.

JelteF: Let's do the same for some of the smaller `Convert` functions.

Reminiscent: @JelteF Updated. Thanks for your review. PTAL

JelteF (approved):
LGTM now. @Y-- do you have some thoughts?

src/pgduckdb_types.cpp
```diff
@@ -301,7 +540,6 @@ ConvertDuckToPostgresArray(TupleTableSlot *slot, duckdb::Value &value, idx_t col
 	D_ASSERT(value.type().id() == duckdb::LogicalTypeId::LIST);
 	auto &child_type = GetChildTypeRecursive(value.type());
 	auto child_id = child_type.id();
-	D_ASSERT(child_id == OP::ExpectedType());
```
Y--, line 304: Since you've removed the `D_ASSERT`, you can now remove the unused variable `child_id`?

src/pgduckdb_types.cpp
```diff
@@ -192,6 +169,269 @@ ConvertNumeric(T value, idx_t scale) {
 	return result;
 }
 
+static Datum
+ConvertNumericDatum(const duckdb::Value &value) {
+	if (value.type().id() == duckdb::LogicalTypeId::DOUBLE) {
```
Y--, line 174: nit: we could assign `value.type().id()` to a const here and re-use it a lot in this function

Y-- (approved):
LGTM, great refactoring + added list support. 
I can think of a few more cleanup/refactoring we could do in a separated PR like:
* split functions that have big `switch/case` to return in each case and handle the type mapping/dimensionality separately
* use C++ vectors/allocator & raii pattern for allocations rather than `palloc`

Overall a lot of good stuff, thanks for the PR!

Reminiscent:
@Y-- Updated. Thanks for your review. PTAL.

Reminiscent:
> LGTM, great refactoring + added list support. I can think of a few more cleanup/refactoring we could do in a separated PR like:
> 
> * split functions that have big `switch/case` to return in each case and handle the type mapping/dimensionality separately
> * use C++ vectors/allocator & raii pattern for allocations rather than `palloc`
> 
> Overall a lot of good stuff, thanks for the PR!

I think you are right. After this PR has been merged. I'd like to do this. Thanks!

Y-- (approved):
LGTM, thanks!

PR #290, opened 2024-10-15
Small improvements to secret and extension management

### LINKED ISSUE #289: DuckDB secrets table should have a PK and/or name column
At the moment, there's no PK or name alias (as there's in the DuckDB secrets manager) on `duckdb.secrets`, making it hard to delete specific secrets.

This is the error I got when inserting twice a secrets :
```
 (PGDuckDB/DuckDBQuery) Invalid Input Error: Initialization function "httpfs_init" from file "/var/lib/postgresql/17/data/duckdb_extensions/v1.1.1/linux_amd64/httpfs.duckdb_extension" threw an exception: "Attempted to register an already registered secret type: 's3'"
 ```

JelteF:
1. Add a unique name column to the secrets table, so people can reference a unique identifier. Defaults to a random uuid. This makes it easy to delete/update/reference a secret.
2. Rename `id` column in secrets to `key_id` to align with DuckDB naming.
3. Correctly track count `secret_table_num_rows` in case of partial failure.
4. Add a primary key to `duckdb.extensions`, you cannot install the same extension multiple times. Let's enforce that.

Fixes #289

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ The best way to get started is to connect Postgres to a new or existing object s
 	```sql
 	-- Session Token is Optional
 	INSERT INTO duckdb.secrets
-	(type, id, secret, session_token, region)
+	(type, key_id, secret, session_token, region)
 	VALUES ('S3', 'access_key_id', 'secret_access_key', 'session_token', 'us-east-1');
 	```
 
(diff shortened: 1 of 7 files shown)


Y-- (approved):
Nice, thanks!

mkaruza (approved):

Issue #216, opened 2024-09-24
Allow configuring of DuckDB settings


JelteF:
DuckDB has a bunch of settings that users probably want to tune based on their specific usage. At least threads memory usage should be configurable, and probably more. I think it is sensible to use postgres its GUC system for this, which would allow people to put things like the following in their postgresql.conf file or use SET commands:

```
duckdb.threads = 1
duckdb.memory = 256MB
duckdb.disabled_filesystems = 'LocalFilesystem'
duckdb.enable_external_access = true
```

Part of this task is figuring out which duckdb settings make sense for people to configure.

PR #291, opened 2024-10-15
Allow configuring of DuckDB settings

Y--:
Fixes https://github.com/duckdb/pg_duckdb/issues/216

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -20,7 +20,6 @@ char *duckdb_motherduck_postgres_user = strdup("");
 
 int duckdb_maximum_threads = -1;
 char *duckdb_maximum_memory = NULL;
-char *duckdb_disabled_filesystems = NULL;
 bool duckdb_enable_external_access = true;
 bool duckdb_allow_unsigned_extensions = false;
 
@@ -95,10 +94,6 @@ DuckdbInitGUC(void) {
 	                     &duckdb_allow_unsigned_extensions);
 
 	DefineCustomVariable("duckdb.max_memory", "The maximum memory DuckDB can use (e.g., 1GB)", &duckdb_maximum_memory);
-
-	DefineCustomVariable("duckdb.disabled_filesystems",
-	                     "Disable specific file systems preventing access (e.g., LocalFileSystem)",
-	                     &duckdb_disabled_filesystems);
 
 	DefineCustomVariable("duckdb.threads", "Maximum number of DuckDB threads per Postgres backend.",
 	                     &duckdb_maximum_threads, -1, 1024);
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -18,6 +18,12 @@ char *duckdb_background_worker_db = strdup("postgres");
 char *duckdb_motherduck_token = NULL;
 char *duckdb_motherduck_postgres_user = NULL;
 
+int duckdb_maximum_threads = -1;
+char *duckdb_maximum_memory = NULL;
+char *duckdb_disabled_filesystems = NULL;
+bool duckdb_enable_external_access = true;
+bool duckdb_allow_unsigned_extensions = false;
+
 extern "C" {
 PG_MODULE_MAGIC;
 
@@ -27,6 +33,7 @@ _PG_init(void) {
 		ereport(ERROR, (errmsg("pg_duckdb needs to be loaded via shared_preload_libraries"),
 		                errhint("Add pg_duckdb to shared_preload_libraries.")));
 	}
+
 	DuckdbInitGUC();
 	DuckdbInitHooks();
 	DuckdbInitNode();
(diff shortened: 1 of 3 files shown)


src/pgduckdb_duckdb.cpp
```diff
@@ -25,6 +25,14 @@ namespace pgduckdb {
 DuckDBManager::DuckDBManager() {
 }
 
+#define SET_DUCKDB_OPTION(ddb_option_name)                                                                             \
+	config.options.ddb_option_name = duckdb_##ddb_option_name;                                                         \
+	{                                                                                                                  \
+		std::ostringstream oss;                                                                                        \
+		oss << "[PGDuckDB] Set DuckDB option: '" << #ddb_option_name << "'=" << duckdb_##ddb_option_name;              \
+		elog(DEBUG2, "%s", oss.str().c_str());                                                                           \
```
JelteF, line 33: For DEBUG logs it's best to put everything in arguments of the `elog` call, because those are not executed when the log level is lower. So something like this:
```suggestion
		elog(DEBUG2, "[PGDuckDB] Set DuckDB option: '%s'=%s", #ddb_option_name, std::string(oss.str()).c_str());                                                                           \
```

src/pgduckdb.cpp
```diff
@@ -27,52 +33,79 @@ _PG_init(void) {
...
+	                     "Allow DuckDB to load extensions with invalid or missing signatures",
+	                     &duckdb_allow_unsigned_extensions);
+
+	DefineCustomVariable("duckdb.max_memory", "The maximum memory DuckDB can use (e.g., 1GB)", &duckdb_maximum_memory);
+
+	DefineCustomVariable("duckdb.disabled_filesystems",
+	                     "Disable specific file systems preventing access (e.g., LocalFileSystem)",
+	                     &duckdb_disabled_filesystems);
+
+	DefineCustomVariable("duckdb.threads", "DuckDB max number of threads.", &duckdb_maximum_threads, -1, 1024);
```
JelteF, line 100:
```suggestion
	DefineCustomVariable("duckdb.threads", "Maximum number of DuckDB threads per Postgres backend.", &duckdb_maximum_threads, -1, 1024);
```

src/pgduckdb.cpp
```diff
@@ -27,52 +33,79 @@ _PG_init(void) {
...
+
+	DefineCustomVariable("duckdb.max_threads_per_query", "DuckDB max number of thread(s) per query.",
```
JelteF, line 102:
```suggestion
	DefineCustomVariable("duckdb.max_threads_per_query", "Maximum number of DuckDB threads used for a single Postgres scan",
```

src/pgduckdb.cpp
```diff
@@ -27,52 +33,79 @@ _PG_init(void) {
...
+	                     &duckdb_max_threads_per_query, 1, 64);
+
+	DefineCustomVariable("duckdb.background_worker_db", "Which database run the backgorund worker in",
```
JelteF, line 105:
```suggestion
	DefineCustomVariable("duckdb.background_worker_db", "Which database run the background worker in",
```

src/pgduckdb.cpp
```diff
@@ -27,52 +33,79 @@ _PG_init(void) {
...
 	DuckdbInitGUC();
 	DuckdbInitHooks();
 	DuckdbInitNode();
 	DuckdbInitBackgroundWorker();
 }
 } // extern "C"
 
-/* clang-format off */
+static void
+DefineCustomVariable(const char *name, const char *short_desc, bool *var) {
```
JelteF, line 45: Nice helpers! Can you add the context and flags as an optional parameter though? Those are often necessary to set.

JelteF (changes requested):

src/pgduckdb.cpp
```diff
@@ -14,10 +14,16 @@ static void DuckdbInitGUC(void);
 
 bool duckdb_execution = false;
 int duckdb_max_threads_per_query = 1;
-char *duckdb_background_worker_db = NULL;
+char *duckdb_background_worker_db = strdup("postgres");
```
JelteF, line 17: I'm not sure if/how this actually works. It seems scary to allocate with `pstrdup` before `_PG_init` is necessarily called.

JelteF: I'm very surprised this passes CI.

JelteF: Okay I figured it out. `MemoryContextInit` is called in Postgres its `main` function before it does any of the `dlopen` calls for the `shared_preload_libraries`. So yeah I guess this is fine.

JelteF: Yeah, I noticed that C++ cares about const char vs non const char when I created that option. It's indeed reasonable. I'm fine with the strdup approach, now that I understand why it works, and I'm convinced that it will continue working.

src/pgduckdb_duckdb.cpp
Y--: Yeah good point - initally I didn't care too much cause I did use `INFO` level, but tests reminded me it wasn't a good idea :-) 
I've reworked it.

JelteF (approved):

JelteF:
Merged because I'm adding some GUCs too.

PR #293, opened 2024-10-15
Handle custom DateStyle correctly

JelteF:
DuckDB only understands ISO dates when parsing dates. So we need to make sure that when we generate the DuckDB query, any date literals are formatted using ISO. We still want the final date output to be whatever the user requested though. We do this by temporarily changing the DateStyle to ISO when generating the DuckDB query.

Ideally we'd even want a TIMESTAMP casted explicitly to TEXT in DuckDB to be formatted using Postgres its DateStyle. But that seems a bridge to far for now. With this change at least queries are parseable by DuckDB when Postgres its DateStyle is set to anything other than ISO.

### DIFF
--- a/include/pgduckdb/pgduckdb_ruleutils.h
+++ b/include/pgduckdb/pgduckdb_ruleutils.h
@@ -1,7 +1,9 @@
 #include "postgres.h"
+#include "nodes/parsenodes.h"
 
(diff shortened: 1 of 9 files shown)


mkaruza (approved):
LGTM

Issue #278, opened 2024-10-13
Extensions are not synced before reconnect


JelteF:
Basically the same problem as #215, but for exensions instead of secrets. This was reported by @rakesh-tmdc in #209

A workaround for this is to either reconnect, or to run `SELECT * FROM duckdb.recycle_ddb()`

I think we should probably do the same solution as in #253

PR #294, opened 2024-10-15
Add `disabled_filesystems` support

### LINKED ISSUE #105: security: GUC to block access to local filesystem
Discussed this a _long_ time ago... but currently we allow DuckDB to read from the local filesystem. This is a security risk; the CSV reader is particularly easy to use here since it will read just about any plain text file. 

This should instead be controllable via a GUC, default disabled, that can only be enabled by superuser.

Y--:
cf. https://duckdb.org/docs/configuration/overview.html
Adds support for DuckDB's `disabled_filesystems` option

Fixes #105 

Also set `enable_external_access`, `allow_unsigned_extensions`, `max_threads` and `max_memory` as only modifiable by a super user.

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -20,6 +20,7 @@ char *duckdb_motherduck_postgres_user = strdup("");
 
 int duckdb_maximum_threads = -1;
 char *duckdb_maximum_memory = NULL;
+char *duckdb_disabled_filesystems = NULL;
 bool duckdb_enable_external_access = true;
 bool duckdb_allow_unsigned_extensions = false;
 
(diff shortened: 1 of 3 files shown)


JelteF (approved):

src/pgduckdb.cpp
```diff
@@ -95,6 +96,10 @@ DuckdbInitGUC(void) {
 
 	DefineCustomVariable("duckdb.max_memory", "The maximum memory DuckDB can use (e.g., 1GB)", &duckdb_maximum_memory);
 
+	DefineCustomVariable("duckdb.disabled_filesystems",
+	                     "Disable specific file systems preventing access (e.g., LocalFileSystem)",
+	                     &duckdb_disabled_filesystems);
```
JelteF, line 101: Needs to be PGC_SUSET

JelteF (changes requested):
This setting should only be modifiable by the postgres super user (see other comment).

src/pgduckdb.cpp
Y--: That's a good point. And that makes me think that these settings won't be taken into account if a (super)user try to update them at runtime unless they call `reset_ddb` method. Probably not a big deal for now but we might want to track it WDYT?

JelteF: Yeah we should probably create an issue to track that, similar to how we have #278. Below some thoughts on the implementation:

For settings that DuckDB allows changing on the fly using `SET` commands this should be relatively easy/elegant to implement. We can use the GucAssignHook for each of these GUCs to run a SET on the database (if it is initialized), whenever the GUC is changed.

For GUCs that are not possible to change without resetting the DuckDB database (such as `threads` afaik), I think there are two options:
1. We change PGC_SUSET to PGC_SUBACKEND, so that changing these settings after connecting to postgres is disallowed.
2. We have the assignhook log a WARNING/NOTICE (if the DuckDB db is initialized) that the change will only be picked up after calling `reset_ddb`.

I think the choice there is how much we want to advertise `reset_ddb`.

JelteF (approved):

PR #297, opened 2024-10-16
Better MotherDuck configuration

JelteF:
1. Allow taking the `motherduck_token` from an environment variable by setting `duckdb.motherduck_enabled = true`.
2. Include instructions for MotherDuck in README
3. Rename `duckdb.background_worker_db` to `duckdb.motherduck_postgres_datase` and check this GUC before connecting to MotherDuck.
4. Allow dropping orphaned MotherDuck shell tables if MotherDuck is not enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,18 +101,49 @@ The best way to get started is to connect Postgres to a new or existing object s
 	LIMIT 100;
 	```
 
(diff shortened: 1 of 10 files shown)


Y-- (approved):
LGTM!

PR #302, opened 2024-10-16
Make sure ANALYZE does not fail on DuckDB tables

JelteF:
Previously if you had any DuckDB tables and ran `ANALYZE` you would get this error:

```
> ANALYZE;
ERROR:  0A000: duckdb does not implement duckdb_relation_size
LOCATION:  duckdb_relation_size, pgduckdb_table_am.cpp:312
```

That's pretty bad because people run `ANALYZE` regularly when doing benchmarks and to ensure the Postgres planner works correctly after a data load. This commit fixes the above issue by "implementing" just enough of the table AM API to make ANALYZE pass.


Y-- (approved):
👍🏼

PR #303, opened 2024-10-17
Docker compose

wuputah:
Usage:

```shell
git clone https://github.com/duckdb/pg_duckdb
cd pg_duckdb
docker compose up -d
docker exec db psql
```

Alternatively, the compose file and config file `docker` directory can be downloaded and used directly without the rest of the repo.


docker/docker-entrypoint-initdb.d/pg_duckdb.sql
```diff
@@ -0,0 +1 @@
+CREATE EXTENSION IF NOT EXISTS pg_duckdb;
```
JelteF, line 1: I created #305 to bake this into the docker images themselves.

docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+services:
+  pgduckdb:
+    container_name: pgduckdb
+    image: pgduckdb/pgduckdb:16-main
```
JelteF, line 4: Something is wrong with how we publish images. This tag is only available for arm64, while some others are available for amd64. But non are available for both.

docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      pgduckdb:
+        condition: service_healthy
+    stdin_open: true
+    tty: true
+    environment:
+      PGPASSWORD: ${PASSWORD:-duckdb}
+    command: psql -h 127.0.0.1 -d postgres -U postgres -p ${PORT:-5432}
+    restart: always
+    networks:
+      - pgduckdb_net
```
JelteF, line 36: I think we can replace this with telling people to run:
```
docker compose exec pgduckdb psql -U postgres
```

wuputah: yeah, inclined to give up on this approach, though I still don't understand why it didn't work despite trying many different things (what I committed here was just the last version of what I tried).

docker-compose.yml
wuputah: opened #307

wuputah:
I figured out the psql issue. The secret to inter-container networking is you need to use the name of service as the hostname. But compose doesnt really support anything interactive, so best to axe it. We could set up adminer or similar, but I tried adminer and is functional but ugly. Maybe we can add drizzle studio which is a bit slicker, but they don't have an image that would work out of the box.

Anyway, done for now.

JelteF (approved):

PR #304, opened 2024-10-17
Convert BPCHAR type to DuckDB VARCHAR without padding

mkaruza:
When conversion from BPCHAR to DuckDB VARCHAR type needs to be done we need to strip extra padding form postgres value.

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -669,9 +669,9 @@ Append(duckdb::Vector &result, T value, idx_t offset) {
 }
 
 static void
-AppendString(duckdb::Vector &result, Datum value, idx_t offset) {
+AppendString(duckdb::Vector &result, Datum value, idx_t offset, bool is_bpchar) {
 	const char *text = VARDATA_ANY(value);
-	int len = VARSIZE_ANY_EXHDR(value);
+	auto len = is_bpchar ? bpchartruelen(VARDATA_ANY(value), VARSIZE_ANY_EXHDR(value)) : VARSIZE_ANY_EXHDR(value);
 	duckdb::string_t str(text, len);
 
 	auto data = duckdb::FlatVector::GetData<duckdb::string_t>(result);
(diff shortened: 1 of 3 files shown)


src/pgduckdb_types.cpp
```diff
@@ -669,9 +669,9 @@ Append(duckdb::Vector &result, T value, idx_t offset) {
 }
 
 static void
-AppendString(duckdb::Vector &result, Datum value, idx_t offset) {
+AppendString(duckdb::Vector &result, Datum value, idx_t offset, bool is_bpchar) {
 	const char *text = VARDATA_ANY(value);
-	int len = VARSIZE_ANY_EXHDR(value);
+	auto len = is_bpchar ? bpchartruelen(VARDATA_ANY(value), VARSIZE_ANY_EXHDR(value)) : VARSIZE_ANY_EXHDR(value);
```
JelteF, line 674:
```suggestion
    /* Remove the padding of a bpchar. DuckDB expects unpadded value. */
	auto len = is_bpchar ? bpchartruelen(VARDATA_ANY(value), VARSIZE_ANY_EXHDR(value)) : VARSIZE_ANY_EXHDR(value);
```

test/regression/expected/type_support.out
```diff
@@ -57,17 +57,17 @@ SELECT * FROM bool_tbl;
...
-             a             
----------------------------
-                          
- test                     
- this is a long string    
+           a           
+-----------------------
+ 
+ test
+ this is a long string
```
JelteF, line 64: This change seems incorrect. I tried with `duckdb.execution = false` and that contains the trailing whitespace too. We should add the padding again when converting back to a Postgres bpchar. It seems a rather harmless difference though, so let's just create an issue for it (unless it's very simple to fix).

mkaruza: I'll create issue for this.

JelteF (approved):

PR #305, opened 2024-10-17
Pre-configure docker image to use pg_duckdb

JelteF:
This slightly modifies the default Postgres docker image so that when it initializes the database it will automatically configure `pg_duckdb`. It also uses the `postgres` user to start the container, so that `docker exec ... psql` just works without having to specify `-U postgres`

This also uses a more selective COPY. I often have some random stuff in my top repo directory, such as gigabytes of TPCDS csv files. This makes sure those are not copied into the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,3 +62,4 @@ RUN echo "shared_preload_libraries='pg_duckdb'" >> /usr/share/postgresql/postgre
 RUN echo "CREATE EXTENSION IF NOT EXISTS pg_duckdb;" >> /docker-entrypoint-initdb.d/0001-install-pg_duckdb.sql
 
 COPY --from=builder /out /
+USER postgres
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,17 +19,23 @@ WORKDIR /build
 ENV PATH=/usr/lib/ccache:$PATH
 ENV CCACHE_DIR=/ccache
 
-# A more selective copy might be nice, but the git submodules are not cooperative.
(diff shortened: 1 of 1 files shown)


Y-- (approved):
Nice!

Y-- (approved):

Dockerfile
```diff
@@ -50,4 +56,10 @@ RUN make installcheck
 ###
 # this creates a usable postgres image but without the packages needed to build
 FROM base AS output
+
+# Automatically enable pg_duckdb
+RUN echo "shared_preload_libraries='pg_duckdb'" >> /usr/share/postgresql/postgresql.conf.sample
```
wuputah, line 61: ah, I didnt think that you could make changes to the _sample file_ (you know, it being named "sample" and all) and it would get picked up somehow later. TIL. :joy: I think it's still nice to be able to inject a config file for other purposes (like motherduck token) via docker-compose, but if this means that the image works without doing so, then great! 👍

Dockerfile
```diff
@@ -50,4 +56,10 @@ RUN make installcheck
...
+RUN echo "CREATE EXTENSION IF NOT EXISTS pg_duckdb;" >> /docker-entrypoint-initdb.d/0001-install-pg_duckdb.sql
```
wuputah, line 62: This is probably okay to do here. I was only hesitant to do so in case some users want to opt out of this behavior, doing so this way means they need to make their own image which removes this file. But if we get push back it can always be removed.

PR #307, opened 2024-10-17
de-matrix the docker platforms

wuputah:
Docker doesn't like it when you push from separate workers


wuputah:
It would be nice to build only natively (i.e. amd64) when doing docker-related PRs and do full builds only on `main` and tags, but I'll leave that to later improvements.

Y--:
An alternative would be to add the platform in the tag. Less elegant but maybe more efficient when GH comes up with Arm runners for open source projects.

Y-- (approved):

PR #316, opened 2024-10-18
Synchronise extensions before reconnect

Y--:
If duckdb.extensions table is inserted, updated or deleted in same connection we would not sync changes for next query. Introduce sequence that keeps track of each change so we can compare with and sync if required.

Inspired from https://github.com/duckdb/pg_duckdb/pull/253

Fixes https://github.com/duckdb/pg_duckdb/issues/278

### DIFF
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -168,7 +168,7 @@ DuckDBManager::LoadSecrets(duckdb::ClientContext &context) {
 
 void
 DuckDBManager::DropSecrets(duckdb::ClientContext &context) {
-	for (auto secret_id = 0; secret_id < secret_table_num_rows; ++secret_id) {
+	for (auto secret_id = 0; secret_id < secret_table_num_rows; secret_id++) {
 		auto drop_secret_cmd = duckdb::StringUtil::Format("DROP SECRET pgduckb_secret_%d;", secret_id);
 		pgduckdb::DuckDBQueryOrThrow(context, drop_secret_cmd);
 	}
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -124,17 +124,11 @@ DuckDBManager::LoadFunctions(duckdb::ClientContext &context) {
 	context.transaction.Commit();
 }
 
-bool
-DuckDBManager::CheckSecretsSeq() {
+int64
+GetSeqLastValue(const char *seq_name) {
 	Oid duckdb_namespace = get_namespace_oid("duckdb", false);
-	Oid secret_table_seq_oid = get_relname_relid("secrets_table_seq", duckdb_namespace);
-	int64 seq =
-	    PostgresFunctionGuard<int64>(DirectFunctionCall1Coll, pg_sequence_last_value, InvalidOid, secret_table_seq_oid);
-	if (secret_table_current_seq < seq) {
-		secret_table_current_seq = seq;
-		return true;
-	}
-	return false;
+	Oid table_seq_oid = get_relname_relid(seq_name, duckdb_namespace);
+	return PostgresFunctionGuard<int64>(DirectFunctionCall1Coll, pg_sequence_last_value, InvalidOid, table_seq_oid);
 }
 
 void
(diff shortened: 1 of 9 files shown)


test/regression/expected/extensions.out
```diff
@@ -0,0 +1,97 @@
...
+parquet	true	true
+pgduckdb	true	false
+
+
+ raw_query 
+-----------
+ 
+(1 row)
+
+SET duckdb.execution TO true;
```
JelteF, line 97: No need to do this. It's a session-level SET, so it will be gone after the test finishes anyway.
```suggestion
```

test/regression/expected/extensions.out
```diff
@@ -0,0 +1,97 @@
+SET duckdb.execution TO false;
```
JelteF, line 1: Needs to be renamed now since #314  
```suggestion
SET duckdb.force_execution TO false;
```

src/pgduckdb_duckdb.cpp
```diff
@@ -167,17 +161,19 @@ DuckDBManager::LoadSecrets(duckdb::ClientContext &context) {
 		DuckDBQueryOrThrow(context, secret_key->data);
 
 		pfree(secret_key->data);
-		secret_id++;
-		secret_table_num_rows = secret_id;
+		++secret_id;
 	}
+
+	secret_table_num_rows = secret_id;
```
JelteF, line 167: I think this was better in the previous place. That way even if `LoadSecrets` fails midway through when adding e.g. the 3rd secret, the 2 previously created secrets would be cleaned up by `DropSecrets`. Ofcourse we should probably handle this with transactions, but we currently don't. So it seems better to keep this line in the same place, and probably add a comment that it makes sense to have it there for now.

JelteF (changes requested):

src/pgduckdb_duckdb.cpp
Y--: Yeah - I actually wanted to make it so if one fails it doesn't arbitrarily fails all the following ones but didn't finish. I might do that in a different PR, but will revert in the meantime.

src/pgduckdb_duckdb.cpp
```diff
@@ -167,17 +161,19 @@ DuckDBManager::LoadSecrets(duckdb::ClientContext &context) {
 		DuckDBQueryOrThrow(context, secret_key->data);
 
 		pfree(secret_key->data);
-		secret_id++;
+		++secret_id;
```
JelteF, line 164: nit: why swap this around? It has no functional difference here right? I checked to see if this is common duckdb style, and generally it seems postfix is preferred:
```
duckdb on  main via △ v3.28.3
❯ rg '\+\+\w+' | wc -l
3766

duckdb on  main via △ v3.28.3
❯ rg '\w+\+\+' | wc -l
10928
```

Same for postgres:
```
postgres-17 on  REL_17_STABLE [$]
❯ rg '\+\+\w+' | wc -l
1014

postgres-17 on  REL_17_STABLE [$]
❯ rg '\w+\+\+' | wc -l
11167
```

Unless there's a reason that I'm not seeing for doing it this way, I'd prefer to keep it this way. Partially to keep the diff small (to simplify reviews and reduce merge conflicts). And partially because that's what I'm used to.

JelteF: Same for a few lines below

src/pgduckdb_duckdb.cpp
```diff
@@ -202,6 +198,8 @@ DuckDBManager::LoadExtensions(duckdb::ClientContext &context) {
 		if (extension.name == "httpfs") {
 			continue;
 		}
+
+		DuckDBQueryOrThrow(context, "INSTALL " + extension.name);
```
JelteF, line 202: Interesting, was this not needed before?

Y--: In the test we insert rows directly in `extensions` table, so we don't run the `install_extension` function. I hesitated using the function or making sure that no matter what we install the extension if needed as it makes the UX nicer, though I don't have strong opinion on it.

JelteF: Is `INSTALL` an actual no-op if the extension is already installed? Or does it do something semi-expensive like network requests, which would be good to avoid on every DB initialization.

Y--: Yep it is a no-op, cf. [here](https://github.com/duckdb/duckdb/blob/52b43b166091c82b3f04bf8af15f0ace18207a64/src/main/extension/extension_install.cpp#L538-L548)

JelteF (approved):

Issue #323, opened 2024-10-20
Make InsertTupleIntoChunk & HeapTupleFetchNextColumnDatum easier to understand


JelteF:
### Description

While working on #322 it took me a lot of time to grasp how InsertTupleIntoChunk & HeapTupleFetchNextColumnDatum worked. Partially it's because they are doing something relatively complex/clever that requires both DuckDB and Postgres understanding, which I think some comments would help with. I also think some refactoring could make a bunch of stuff a bit clearer. 

To call out some of the things that confused me.
1. The conversion from `DuckDB` column indexes, which are 0 based to Postgres column indexes which are 1 based.
2. The 4 different `idx` variables: `columnIdx`, `valueIdx`, `idx` & `output_column_idx`
3. `att_num` and `attnum` variable in the same function, which are not the same
4. `m_read_column_ids` and `m_output_column_ids` are very easy to confuse.

I'm planning to refactor/comment this code a bit, because it's a very important piece of code, and as #322 showed it was incorrect in a (to me) hard to discover way.

Y--:
Great points. I actually had started re-working the code in here https://github.com/duckdb/pg_duckdb/compare/main...yl/custom-allocator as I had some of the same concerns. (note that this branch was doing refactoring + changing allocators, which I would have split in two then).

Another thing I noticed when working on this code (and slightly unrelated to this issue) is that we could actually only allocate once (store it in local state) at the beginning of the execution and re-use it for every call.

JelteF:
You mean allocate the `values` and `nulls` arrays once? Yeah that seems like an easy perf improvement.

Y--:
Yeah exactly.

JelteF:
Just for reference, there was another bug because `m_read_columns_ids` and `m_output_columns_ids` ideas are easy to confuse: [`ba2f7b3` (#328)](https://github.com/duckdb/pg_duckdb/pull/328/commits/ba2f7b351ce9cadec668415cc5b8052553fe0c86)

Y--:
> Just for reference, there was another bug because `m_read_columns_ids` and `m_output_columns_ids` ideas are easy to confuse: [`ba2f7b3` (#328)](https://github.com/duckdb/pg_duckdb/pull/328/commits/ba2f7b351ce9cadec668415cc5b8052553fe0c86)

That being said, those are pretty explicit, in this case, it was more a classical PEBCAK ;-)

Y--:
[...]

PR #324, opened 2024-10-21
Allocate buffer once per local state

Y--:
As discussed in https://github.com/duckdb/pg_duckdb/issues/323: only allocate `values` and `nulls` buffers once per local state (i.e 1/ thread participating to the query) rather than once per call (tuple)

### DIFF
--- a/include/pgduckdb/scan/postgres_scan.hpp
+++ b/include/pgduckdb/scan/postgres_scan.hpp
@@ -33,12 +33,33 @@ public:
 
 class PostgresScanLocalState {
 public:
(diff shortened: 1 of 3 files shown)


mkaruza (approved):

JelteF (approved):
I'm not super happy about merging perf improvements so close to the release. But this seems simple enough that I think it's probably worth doing it anyway. @mkaruza had a branch with some more improvements, and I think we'd better wait with those until after the release.

This actually seems to resolve a possible memory leak too, because now we always call `duckdb_free`. Previously we'd forget to do this in case of exceptions.

PR #328, opened 2024-10-21
Allocate buffer once per local state

Y--:
As discussed in https://github.com/duckdb/pg_duckdb/issues/323: only allocate values and nulls buffers once per local state (i.e 1/ thread participating to the query) rather than once per call (tuple)


JelteF (approved):

JelteF:
Seems this improves perf by ~10% on most TPCDS queries when using `duckdb.force_execution = true`

PR #329, opened 2024-10-21
Map DuckDB main schema to public

JelteF:
After having some people play with the MotherDuck support we realized we should be mapping the `main` schema to `public` to make it easy to access the tables that are in there.

### DIFF
--- a/src/pgduckdb_background_worker.cpp
+++ b/src/pgduckdb_background_worker.cpp
@@ -148,6 +148,14 @@ char *current_motherduck_catalog_version;
 static std::string
 PgSchemaName(const std::string &db_name, const std::string &schema_name, bool is_default_db) {
 	if (is_default_db) {
+		/*
+		 * We map the "main" DuckDB schema of the default database to the
+		 * "public" Postgres schema, because they are pretty much equivalent
+		 * from a user perspective even though they are named differently.
+		 */
+		if (schema_name == "main") {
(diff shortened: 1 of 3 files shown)


src/pgduckdb_background_worker.cpp
```diff
@@ -399,12 +407,12 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
 	return true;
 }
 
-void
-SyncMotherDuckCatalogsWithPg_Unsafe(bool drop_with_cascade);
+void SyncMotherDuckCatalogsWithPg_Unsafe(bool drop_with_cascade);
```
Y--, line 410: Was this change triggered by `make format`? It feels like it goes the other way on my end :-) Maybe I'm not up to date?

Y-- (approved):
Code LGTM. From an UX perspective I wonder if we won't want a flag to turn it off as it might be confusing for some power users. Though we can totally add later.

src/pgduckdb_background_worker.cpp
JelteF: Hmm, it seems I had a too new `clang-format` that did a bit more. Downgraded now for consistency to the version that is in our `dev_requirements.txt`. It doesn't seem to be formatted back with this older version though. So I'll leave it as is.

PR #332, opened 2024-10-21
Custom vector allocator

Y--:
Introduces a custom DuckDB allocator to avoid using `duckdb_malloc` and `duckdb_free` which might lead to memory leaks.


src/scan/postgres_scan.cpp
```diff
@@ -52,7 +52,6 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
 	for (auto const &[attr_id, column_idx] : ordered_input_columns) {
 		m_input_columns.emplace_back(attr_id, column_idx);
 
-		duckdb::TableFilter *column_filter = nullptr;
```
Y--, line 55: Forgotten in a previous cleanup

mkaruza (approved):

Issue #271, opened 2024-10-10
Test out and fix non-superuser behaviour


JelteF:
Right now only a Postgres superuser can use the extension. Any other user will get the following error:
```
ERROR:  XX000: (PGDuckDB/CreatePlan) Executor Error: permission denied for sequence secrets_table_seq
LOCATION:  DuckDBFunctionGuard, pgduckdb_utils.hpp:98
```

This is obviously by accident. And can be easily fixed using a command like:
```
GRANT ALL ON SEQUENCE duckdb.secrets_table_seq TO PUBLIC; -- Probably not use ALL though
```

But doing so is currently extremely bad security-wise. Because now, any user can simply read from all tables in any of schemas they have access to by simply enabling `duckdb.execution`:

```
> select * from main.foo4;
ERROR:  42501: permission denied for table foo4
> set duckdb.execution = TRUE;
SET
> select * from main.foo4;
 a
───
 1
(1 rows)
```

We should be checking ACLs of tables *and columns* before executing the query to see if the current user has access to them. A very crude, but possibly easy way of doing that is by invoking the Postgres planner on the query again like we did before #247, i.e. we don't actually use the resulting plan, we just have the postgres planner check the permissions on the query.

Also we should probably just bail out in `IsAllowedStatement` if there's row level security set up on any of the tables.

wuputah:
Good catch on the sequence. I think that's a slightly tangential and certainly unintentional issue.

> A very crude, but possibly easy way of doing that is by invoking the Postgres planner on the query again like we did before #247, i.e. we don't actually use the resulting plan, we just have the postgres planner check the permissions on the query.

+1 to this approach as it's basically what I assumed is/was/should be happening. Not sure I would classify it as crude 😂 but instead "leveraging Postgres internals", but that's just marketing.

RLS should definitely be a feature for later... I can see some cool stuff here if we can get it right. For now if we can bail when we detect it (or fallback to Postgres execution) that sounds fine to me.

PR #338, opened 2024-10-22
Update default settings

### LINKED ISSUE #217: Decide on default DuckDB configuration that makes sense for use in postgres
- [ ] how many threads
- [ ] how much memory
- [ ] disabled_filesystems = 'LocalFilesystem' (fixes #105)
- [ ] more?

Related to but separate from #216

JelteF:
Fixes #217 

This changes the defaults for `duckdb.max_memory` and `duckdb.disabled_filesystems` to the agreed upon settings. While doing that I found two issues that this also fixes:
1. `duckdb.disabled_filesystems = 'LocalFilesystem'` was basically a way to make DuckDB execution unusable, due to us doing file system access when initializing the database/connection. So now we enable the setting later. Also we now don't call `INSTALL {extension}` during connection initialization anymore.
2. #316 didn't fix the loading of extensions when `duckdb.install_extension()` was called. Only for direct inserts to the `duckdb.extensions` table. That's fixed by not using the low-level `CatalogTupleInsert`, and instead switching to SPI.
3. Postgres superusers should not be restricted by `duckdb.disabled_filesystems`, they should be able to do whatever they want. So now this setting is not actually configured for superusers anymore.

### DIFF
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -193,7 +193,6 @@ DuckDBManager::LoadExtensions(duckdb::ClientContext &context) {
 			continue;
 		}
 
-		DuckDBQueryOrThrow(context, "INSTALL " + extension.name);
 		DuckDBQueryOrThrow(context, "LOAD " + extension.name);
 	}
 }
@@ -207,16 +206,6 @@ DuckDBManager::CreateConnection() {
 	auto &instance = Get();
 	auto connection = duckdb::make_uniq<duckdb::Connection>(*instance.database);
 	auto &context = *connection->context;
-	if (!superuser() && duckdb_disabled_filesystems != NULL) {
-		pgduckdb::DuckDBQueryOrThrow(context,
-		                             "SET disabled_filesystems='" + std::string(duckdb_disabled_filesystems) + "'");
-		instance.disabled_filesystems_is_set = true;
-	}
-
-	else if (superuser() && instance.disabled_filesystems_is_set) {
-		pgduckdb::DuckDBQueryOrThrow(context, "SET disabled_filesystems=''");
-		instance.disabled_filesystems_is_set = false;
-	}
 
 	const auto secret_table_last_seq = GetSeqLastValue("secrets_table_seq");
 	if (instance.IsSecretSeqLessThan(secret_table_last_seq)) {
@@ -234,6 +223,21 @@ DuckDBManager::CreateConnection() {
 	auto http_file_cache_set_dir_query =
 	    duckdb::StringUtil::Format("SET http_file_cache_dir TO '%s';", CreateOrGetDirectoryPath("duckdb_cache"));
 	context.Query(http_file_cache_set_dir_query, false);
+
+	if (!superuser() && duckdb_disabled_filesystems != NULL) {
+		/*
+		 * DuckDB does not allow us to disable this setting on the
+		 * database after the DuckDB connection is created for a non
+		 * superuser, any further connections will inherit this
+		 * restriction. This means that once a non-superuser used a
+		 * DuckDB connection in aside a Postgres backend, any loter
+		 * connection will inherit these same filesystem restrictions.
+		 * This shouldn't be a problem in practice.
+		 */
+		pgduckdb::DuckDBQueryOrThrow(context,
+		                             "SET disabled_filesystems='" + std::string(duckdb_disabled_filesystems) + "'");
+		instance.disabled_filesystems_is_set = true;
+	}
 
 	return connection;
 }
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -104,11 +104,6 @@ DuckDBManager::Initialize() {
 
 	LoadFunctions(context);
 	LoadExtensions(context);
-
-	if (duckdb_disabled_filesystems != NULL) {
-		pgduckdb::DuckDBQueryOrThrow(context,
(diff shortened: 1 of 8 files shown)


wuputah (approved):

JelteF:
@Y-- can you look into these test failures related to `disabled_filesystems='LocalFilesystem'` to see if they are fixable? We might have to only set `disabled_filesystems` while a user is executing a query or something.

JelteF:
One other probably good change is to never set `disabled_filesystems` in duckdb for queries done by the postgres superuser, but that wouldn't apply to the changes I'm doing for: https://github.com/duckdb/pg_duckdb/issues/271

So that's not really a real solution, just a thing that we probably want to do anyway.

wuputah (dismissed):

test/regression/regression.conf
```diff
@@ -3,3 +3,4 @@
 shared_preload_libraries = 'pg_duckdb'
 duckdb.force_execution = true
 log_temp_files = -1
+duckdb.disabled_filesystems = ''
```
JelteF, line 6: This should not be necessary. If we need to do this in tests it means this is a bad default (at the moment). Having a default  that breaks half of pg_duckdb half of the functionality completely, even for the superuser

I think there's two things we need to do:
1. temporarily change `duckdb_disabled_filesystems` to `''` when we know we want to interact with the filesystem. Like installing duckdb extensions, or when caching a parquet file.
2. Make sure that for postgres superusers this setting is always set to `''`. They can already access files on the filesystem in various ways, so there's no point in blocking them when executing queries in DuckDB. Also they could just change the `duckdb.disabled_filesystems` setting.

Those two should probably build on top of #342

src/pgduckdb_options.cpp
```diff
@@ -115,36 +116,25 @@ ReadDuckdbExtensions() {
...
-	}
-
-	bool nulls[Natts_duckdb_extension] = {0};
-	Datum values[Natts_duckdb_extension] = {0};
-
-	values[Anum_duckdb_extension_name - 1] = name;
-	values[Anum_duckdb_extension_enable - 1] = 1;
-
-	/* create heap tuple and insert into catalog table */
-	Relation duckdb_extensions_relation = relation_open(ExtensionsTableRelationId(), RowExclusiveLock);
```
Y--, line 139: Is `ExtensionsTableRelationId` still used somewhere now?

src/pgduckdb_options.cpp
```diff
@@ -115,36 +116,25 @@ ReadDuckdbExtensions() {
...
-	TupleDesc tuple_descr = RelationGetDescr(duckdb_extensions_relation);
-
-	/* inserting extension record */
-	HeapTuple new_tuple = heap_form_tuple(tuple_descr, values, nulls);
-	CatalogTupleInsert(duckdb_extensions_relation, new_tuple);
-
-	CommandCounterIncrement();
-	relation_close(duckdb_extensions_relation, RowExclusiveLock);
+	pgduckdb::DuckDBQueryOrThrow(install_extension_command);
```
Y--, line 123: We need to wrap the call to `DuckdbInstallExtension` in `install_extension` now o/w `duckdb.install_extension('foo')` will crash psql :-)

src/pgduckdb_duckdb.cpp
```diff
@@ -230,6 +224,21 @@ DuckDBManager::CreateConnection() {
 	    duckdb::StringUtil::Format("SET http_file_cache_dir TO '%s';", CreateOrGetDirectoryPath("duckdb_cache"));
 	context.Query(http_file_cache_set_dir_query, false);
 
+	if (!superuser() && duckdb_disabled_filesystems != NULL) {
```
Y--, line 227: nit- looks like `superuser()` can be non trivial, so:
```suggestion
	if (duckdb_disabled_filesystems != NULL && !superuser()) {
```

src/pgduckdb_options.cpp
JelteF: In `ReadDuckdbExtensions`

src/pgduckdb_options.cpp
JelteF: Nope `DuckDBQueryOrThrow` currently throws an `elog` not an exception. We probably want to change that in the next release, and we'll need to be careful with all these, but for now this seems totally fine imo.

src/pgduckdb_options.cpp
```diff
@@ -115,36 +116,25 @@ ReadDuckdbExtensions() {
...
-	values[Anum_duckdb_extension_name - 1] = name;
-	values[Anum_duckdb_extension_enable - 1] = 1;
-
-	/* create heap tuple and insert into catalog table */
-	Relation duckdb_extensions_relation = relation_open(ExtensionsTableRelationId(), RowExclusiveLock);
-	TupleDesc tuple_descr = RelationGetDescr(duckdb_extensions_relation);
-
-	/* inserting extension record */
-	HeapTuple new_tuple = heap_form_tuple(tuple_descr, values, nulls);
-	CatalogTupleInsert(duckdb_extensions_relation, new_tuple);
```
JelteF, line 144: This low level API was not causing the trigger to be fired. So this PR actually also fixes a bug where the `duckdb.install_extension()` call would not update the sequence, and thus not existing cause connections to load the extension.

src/pgduckdb_options.cpp
Y--: Ah that's right - thanks!

Y-- (commented):
LGTM!

Y-- (approved):
LGTM!

Y-- (approved):

Issue #336, opened 2024-10-22
Server crashes If extension drops inside transaction block


saygoodbyye:
### What happens?

Server crashes when executing the script below
PostgreSQL build:
```
CFLAGS='-Og' ./configure --enable-tap-tests --with-openssl --enable-debug --enable-cassert --with-icu --with-lz4 --with-libxml
```

### To Reproduce

test.sql:
```sql
CREATE EXTENSION pg_duckdb;
SET duckdb.execution = true;
BEGIN;
DROP EXTENSION pg_duckdb;
SELECT COUNT(*) FROM t WHERE a % 10 = 0;
END;
```

backtrace:
```c
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x000077fb2744526e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x000077fb274288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x0000621fdc1b64fc in ExceptionalCondition (conditionName=conditionName@entry=0x621fdc2c3bfa "IsTransactionState()", fileName=fileName@entry=0x621fdc2d1337 "relcache.c",
    lineNumber=lineNumber@entry=2068) at assert.c:66
#6  0x0000621fdc1ae094 in RelationIdGetRelation (relationId=<optimized out>, relationId@entry=3079) at relcache.c:2068
#7  0x0000621fdbc8bb23 in relation_open (relationId=relationId@entry=3079, lockmode=lockmode@entry=1) at relation.c:58
#8  0x0000621fdbd0de91 in table_open (relationId=relationId@entry=3079, lockmode=lockmode@entry=1) at table.c:44
#9  0x0000621fdbe0e1f9 in get_extension_oid (extname=0x77fb27fbda68 "pg_duckdb", missing_ok=true) at extension.c:153
#10 0x000077fb27f8a17f in pgduckdb::IsExtensionRegistered () at src/pgduckdb_metadata_cache.cpp:159
[...]

PR #339, opened 2024-10-22
Don't check if extension is registered in aborted tx

Y--:
Fixes https://github.com/duckdb/pg_duckdb/issues/336


JelteF (approved):
Let's open an issue to investigate why we're calling this function in aborted transaction. That seems not on purpose.

Y--:
> Let's open an issue to investigate why we're calling this function in aborted transaction. That seems not on purpose.

Filed https://github.com/duckdb/pg_duckdb/issues/349

PR #341, opened 2024-10-22
Allow hooks to be set using our DefineCustomVariable wrapper

### LINKED ISSUE #296: DuckDB GUCs cannot be changed after the first DuckDB query
Settings such as `duckdb.disabled_filesystems` are set when the DuckDB database is first created, but the value is not reflected anymore when changing the GUCs later in that session using a `SET` command. For settings that DuckDB allows changing on the fly using `SET` commands this should be relatively easy/elegant to implement. We can use the GucAssignHook for each of these GUCs to run a SET on the database (if it is initialized), whenever the GUC is changed.

For GUCs that are not possible to change without resetting the DuckDB database (such as `threads` afaik), I think there are two options:
1. We change PGC_SUSET to PGC_SUBACKEND, so that changing these settings after connecting to postgres is disallowed.
2. We have the assignhook log a WARNING/NOTICE (if the DuckDB db is initialized) that the change will only be picked up after calling `reset_ddb`.

I think the choice there is how much we want to advertise `reset_ddb` (i.e. is it only meant for debugging, or is it a really supported function, see also #295)

JelteF:
This doesn't actually start using these hooks, but it's likely that we'll want to do so to fix #296.

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -44,27 +44,33 @@ _PG_init(void) {
 
 static void
 DefineCustomVariable(const char *name, const char *short_desc, bool *var, GucContext context = PGC_USERSET,
-                     int flags = 0) {
-	DefineCustomBoolVariable(name, gettext_noop(short_desc), NULL, var, *var, context, flags, NULL, NULL, NULL);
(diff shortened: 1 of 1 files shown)


Y--:
Should we do that only when working on https://github.com/duckdb/pg_duckdb/issues/296 then?

Y-- (approved):

PR #342, opened 2024-10-22
Support non-superuser usage of pg_duckdb

JelteF:
DuckDB is not really designed for use in a multi-user system, and Postgres is a multi-user system. This is visible through at least two ways:

1. An important security concern here is that secrets are usable by all users, and the same concern applies to MotherDuck server-side secrets. We can possibly work around these issues.
2. DuckDB doesn't enforce any permissions when accessing tables, but Postgres applies permissions on tables.

So security wise it's not straight forward to allow non-superusers in Postgres to use DuckDB execution, without allowing them to escalate privileges somehow. Allowing regular users to run arbitrary SQL in DuckDB is a huge security no-go, since they can bypass any permission checks from any table in Postgres.

This concern is addressed in five ways:

1. By default only allow superuser to run DuckDB queries. Security wise this would be enough on its own, but since this makes usage on production systems near-impossible we add some more options.
2. Allow configuring a `duckdb.postgres_role` setting which contains a role name that can also use DuckDB execution. A role in Postgres can be used as a group and can be assigned to multiple users.
3. Have Postgres check permissions on tables used in a query before forwarding them DuckDB.
4. Don't support queries where Row Level Security (RLS) needs to be enforced when reading tables.
5. Don't allow non-superusers to run any functions in the `duckdb`
   schema that trigger manually constructed queries. So no `duckdb.raw_query()`, and no `duckdb.install_extension()`. I haven't deeply investigated these for possible SQL injection yet, but it's `raw_query` is SQL injection by design, and `LoadExtensions` is currently not escaping the extension name yet before passing it to queries. So for now these are super user only to decrease attack surface.

NOTE: This means we have to be extra careful to avoid SQL injection when constructing DuckDB queries, because that would allow a non-superuser with the `duckdb.postgres_role` to execute arbitrary SQL in DuckDB, and thus read arbitrary Postgres tables. Luckily even if this happens currently, it's limited to reading arbitrary data, not modifying, because we don't support writes to Postgres tables from DuckDB yet.

A small follow up change I will implement on this tomorrow is to create MotherDuck tables as the `duckdb.postgres_role`, that way users with the `duckdb.postgres_role` can actually read these tables (now they cannot due to the new permission checks).


wuputah:
so 

```
duckdb.postgres_role = 'duckdb';
```

then I can

```
GRANT duckdb TO some_users_or_groups;
```

to enable usage.

sql/pg_duckdb--0.0.1.sql
```diff
@@ -232,12 +257,15 @@ REVOKE ALL ON tables FROM PUBLIC;
...
     LANGUAGE C AS 'MODULE_PATHNAME', 'install_extension';
+REVOKE ALL ON FUNCTION install_extension(TEXT) FROM PUBLIC;
 
 CREATE OR REPLACE FUNCTION raw_query(query TEXT) RETURNS void
     LANGUAGE C AS 'MODULE_PATHNAME', 'pgduckdb_raw_query';
+REVOKE ALL ON FUNCTION raw_query(TEXT) FROM PUBLIC;
 
 CREATE OR REPLACE FUNCTION cache(object_path TEXT, type TEXT) RETURNS bool
     LANGUAGE C AS 'MODULE_PATHNAME', 'cache';
+REVOKE ALL ON FUNCTION cache(TEXT, TEXT) FROM PUBLIC;
```
wuputah, line 268: is this being blocked because its parameters do not get escaped?

wuputah:
A nice idea is standard Postgres permissions to `duckdb.secrets` could be enforced, ideally with RLS allowed, based on who is running the query. But this probably messes with the idea that secrets are cached between queries and only reloaded if the table data changes.

sql/pg_duckdb--0.0.1.sql
JelteF: Yeah. This definitely doesn't seem safe:

```
	auto cache_object_query =
	    duckdb::StringUtil::Format("SELECT 1 FROM %s('%s');", object_type_fun, object_path.c_str());
```

JelteF:
> A nice idea is standard Postgres permissions to duckdb.secrets could be enforced, ideally with RLS allowed, based on who is running the query. But this probably messes with the idea that secrets are cached between queries and only reloaded if the table data changes.

Yeah, I think we can definitely design secrets differently (at least the non motherduck ones), so that different PG users have access to different sets of secrets. That seemed way to complex for me to write on the plane to PGConf.eu though. And probably need a bunch of design discussions. So for now I went with the most simple thing I could think of that allows non superusers to do at least a bunch of things with DuckDB in a safe manner.

sql/pg_duckdb--0.0.1.sql
wuputah: 😬

JelteF:
> so
> 
> ```
> duckdb.postgres_role = 'duckdb';
> ```
> 
> then I can
> 
> ```
> GRANT duckdb TO some_users_or_other_roles;
> ```
> 
> to enable usage.

Correct. Although that's assuming you run `CREATE EXTENSION pg_duckdb` after configuring `duckdb.postgres_role = 'duckdb'`. If you create the extension before, and then configure, then you have to manually create the role.

> This seems sane enough to be the default behavior / setting, if we are OK with creating a role by default.

Yeah, I did consider that. But having pg_duckdb create a role by default has one annoying implication:
Roles do not become owned by an extension like pretty much all other objects created by an extension. This is because they are objects global to the postgres server, i.e. they don't live in a database. So that means that if you do `CREATE EXTENSION pg_duckdb; DROP EXTENSION pg_duckdb;` on a fresh server, that you now suddenly have a leftover `duckdb` role there that you need to clean up manually.

And there's also the security side to this. If there already exists a `duckdb` user on the server that has nothing to do with pg_duckdb, then that user suddenly can access the.

It seemed a bit late to decide that now. And it seemed easier to decide we want `duckdb` as the default for 0.2.0, then reverting `duckdb` as the default to no default.

sql/pg_duckdb--0.0.1.sql
JelteF: To be clear, I think we can probably get these functions SQL injection safe fairly easily. I just didn't want to batch even more into this already fairly big and complex PR.

src/pgduckdb_ruleutils.cpp
```diff
@@ -128,7 +129,28 @@ pgduckdb_relation_name(Oid relation_oid) {
 	Form_pg_class relation = (Form_pg_class)GETSTRUCT(tp);
 	const char *relname = NameStr(relation->relname);
 	const char *postgres_schema_name = get_namespace_name_or_temp(relation->relnamespace);
-	const char *db_and_schema = pgduckdb_db_and_schema_string(postgres_schema_name, pgduckdb::IsDuckdbTable(relation));
+	bool is_duckdb_table = pgduckdb::IsDuckdbTable(relation);
+	const char *db_and_schema = pgduckdb_db_and_schema_string(postgres_schema_name, is_duckdb_table);
```
Y--, line 133: nit - we could move that call after the `if`

src/utility/copy.cpp
```diff
@@ -154,6 +166,37 @@ CreateCopyOptions(CopyStmt *copy_stmt, bool *options_valid) {
 	return options_string;
 }
 
+/*
+ * Throws an error if a rewritten raw statement returns an unexpected number of
+ * queries (i.e. not just a single query). This is taken from Postgres its
+ * BeginCopyTo function.
+ */
+void
+check_rewritten(List *rewritten) {
```
Y--, line 175: nit
```suggestion
CheckRewritten(List *rewritten) {
```
(to keep consistency w/ other functions in this file - unless it's to keep the Postgres original casing?)

Also it seems this function isn't used?

src/utility/copy.cpp
```diff
@@ -185,28 +228,26 @@ DuckdbCopy(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment
...
 		raw_stmt->stmt = copy_stmt->query;
 		raw_stmt->stmt_location = pstmt->stmt_location;
 		raw_stmt->stmt_len = pstmt->stmt_len;
 
-		rewritten = pg_analyze_and_rewrite_fixedparams(raw_stmt, query_string, NULL, 0, NULL);
-		query = linitial_node(Query, rewritten);
+		List *rewritten = pg_analyze_and_rewrite_fixedparams(raw_stmt, query_string, NULL, 0, NULL);
+		if (list_length(rewritten) != 1) {
+			return false;
+		}
```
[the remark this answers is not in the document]
JelteF: nah, the other way around. Fixed now.

Y-- (approved):

PR #343, opened 2024-10-22
Resolve security issues reported by pgspot

### LINKED ISSUE #250: Run pgspot tool to find security issues
It's really easy to make critical security mistakes in the SQL scripts of an extension. Luckily [pgspot](https://github.com/timescale/pgspot) can help finding those.

JelteF:
It's really easy to make critical security mistakes in the SQL scripts of an extension. Luckily [pgspot][1] can help finding those. There were a few false positives because the tool did not realize that the search_path `duckdb, pg_catalog, pg_temp` was also safe. I'll probably make a PR to pgspot to allow specifying additional safe schemas. For now I just followed its advice to get to 0 errors, and I think the script is still pretty readable.

I also changed the script to use `@extschema@` whenever we were previously storing a function in the `public` schema. This way we can more easily choose to make the extension relocatable in the future. Or at least allow people to store the these objects in a schema of their choice by doing `CREATE EXTENSION pg_duckdb WITH SCHEMA some_other_schema`. For now I forced the schema to `public` in the control file. Because we don't have tests for this multi schema stuff and we might not think its worth the effort in the long run.

[1]: https://github.com/timescale/pgspot

Fixes #250

### DIFF
--- a/sql/pg_duckdb--0.0.1.sql
+++ b/sql/pg_duckdb--0.0.1.sql
@@ -336,9 +336,9 @@ CREATE EVENT TRIGGER duckdb_alter_table_trigger ON ddl_command_end
     WHEN tag IN ('ALTER TABLE')
     EXECUTE FUNCTION duckdb_alter_table_trigger();
 
--- We explicitely don't set the search_path here in the functinon definitian.
+-- We explicitly don't set the search_path here in the function definition.
 -- Because we actually need the original search_path that was active during the
--- GRANT to reslove the RangeVar using RangeVarGetRelid. We don't need this for
+-- GRANT to resolve the RangeVar using RangeVarGetRelid. We don't need this for
 -- any of the other triggers since those don't manually resolve RangeVars, at
 -- least not yet. So for those we might as well err on the side of caution and
 -- force a safe search_path.
--- a/sql/pg_duckdb--0.0.1.sql
+++ b/sql/pg_duckdb--0.0.1.sql
@@ -1,40 +1,42 @@
 LOAD 'pg_duckdb';
 
--- If duckdb.postgres_role is configured let's make sure it's actually created.
--- If people change this setting after installing the extension they are
--- responsible for creating the user themselves. This is especially useful for
--- demo purposes and our pg_regress testing.
-DO $$
-DECLARE
-    role_name text;
-BEGIN
-    SELECT current_setting('duckdb.postgres_role') INTO role_name;
-    IF role_name != '' AND NOT EXISTS (
-      SELECT FROM pg_catalog.pg_roles
-      WHERE  rolname = role_name) THEN
-        EXECUTE 'CREATE ROLE ' || quote_ident(current_setting('duckdb.postgres_role'));
-    END IF;
-END
-$$;
+-- We create a duckdb schema to store most of our things. We explicitely
+-- don't use CREATE IF EXISTS or the schema key in the control file, so we know
+-- for sure that the extension will own the schema and thus non superusers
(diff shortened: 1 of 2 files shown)


sql/pg_duckdb--0.0.1.sql
```diff
@@ -296,17 +336,25 @@ CREATE EVENT TRIGGER duckdb_alter_table_trigger ON ddl_command_end
     WHEN tag IN ('ALTER TABLE')
     EXECUTE FUNCTION duckdb_alter_table_trigger();
 
+-- We explicitely don't set the search_path here in the functinon definitian.
+-- Because we actually need the original search_path that was active during the
+-- GRANT to reslove the RangeVar using RangeVarGetRelid. We don't need this for
+-- any of the other triggers since those don't manually resolve RangeVars, at
+-- least not yet. So for those we might as well err on the side of caution and
+-- force a safe search_path.
```
wuputah, line 344:
```suggestion
-- We explicitly don't set the search_path here in the function definition.
-- Because we actually need the original search_path that was active during the
-- GRANT to resolve the RangeVar using RangeVarGetRelid. We don't need this for
-- any of the other triggers since those don't manually resolve RangeVars, at
-- least not yet. So for those we might as well err on the side of caution and
-- force a safe search_path.
```

wuputah (dismissed):
just noticed some typos in comments

sql/pg_duckdb--0.0.1.sql
```diff
@@ -1,40 +1,42 @@
...
+-- for sure that the extension will own the schema and thus non superusers
+-- cannot put random things in it, so we can assume it's safe. A few functions
+-- we'll put into @extschema@ so that in normal usage they get put into the
+-- public schema and are thus more easily usable. This is the case for the
+-- read_csv, read_parquet and iceberg functions. It would be sad for usability
+-- if people would have to prefix those with duckdb.read_csv
+CREATE SCHEMA duckdb;
+-- Allow users to see the objects in the duckdb schema. We'll manually revoke rights
+-- for the dangerous ones.
+GRANT USAGE ON SCHEMA duckdb to PUBLIC;
```
Y--, line 14: We can't use `@extschema@` here?

sql/pg_duckdb--0.0.1.sql
```diff
@@ -117,20 +123,22 @@ $func$;
...
+SET search_path = pg_catalog, pg_temp
+AS
 $func$
 BEGIN
     RAISE EXCEPTION 'Function `iceberg_scan(TEXT)` only works with Duckdb execution.';
 END;
 $func$;
 
-CREATE TYPE iceberg_metadata_record AS (
+CREATE TYPE duckdb.iceberg_metadata_record AS (
```
Y--, line 141: Same question here (and below where we have `duckdb.xx`

JelteF: We could, but it pollutes the `public` namespace with stuff that does not need to be there. We mainly put a few functions in the public schema to allow accessing them without the `duckdb.` prefix. These types we no-one really needs to reference, so there's no reason for them to be in the `public` schema.

sql/pg_duckdb--0.0.1.sql
JelteF: So it's kinda nice to group most of the `duckdb` functions in the `duckdb`, nice for auto complete and listing them basically.

Y--: Ah sorry - I thought `@extschema@` was an alias for `duckdb`. Nevermind then!

Y-- (approved):

wuputah (approved):

PR #344, opened 2024-10-22
Correctly set missing_ok to false in the GRANT trigger

JelteF:
We expect this RangeVarGetRelid call to actually find the relid, since the event trigger only fires when the grant was successfully done by Postgres. But we were passing it `missing_ok = false` as the last argument. In practice this didn't matter, because it always found it. But in an earlier version of #343 I broke that, which would then cause a crash later on. This passing `missing_ok = true` instead so that next someone breaks something at least it will fail with a nice error instead of a crash.

### DIFF
--- a/src/pgduckdb_ddl.cpp
+++ b/src/pgduckdb_ddl.cpp
@@ -643,7 +643,7 @@ duckdb_grant_trigger(PG_FUNCTION_ARGS) {
 	}
 
 	foreach_node(RangeVar, object, stmt->objects) {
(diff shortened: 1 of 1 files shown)


Y--:
I assume it's not trivial, but could we add a test?

JelteF:
> I assume it's not trivial, but could we add a test?

I have no clue how. I think it's only possible to hit this by creating mistakes somewhere else. To clarify I hit this because I was forcing the search_path to `pg_catalog, pg_temp` in the GRANT trigger

mkaruza (approved):

PR #351, opened 2024-10-23
Don't get page if buffer was released

mkaruza:
There is edge case when we relase buffer and also have complete output vector filled. In that case we would call `BufferGetPage` with InvalidBuffer as parameter. Current check is only to check block which can be valid. We should check do we need to read next buffer or not instead.


Y-- (approved):

Y--:
Can we add a test?

Y-- (approved):

PR #352, opened 2024-10-23
Create MotherDuck tables as duckdb.postgres_user

JelteF:
Before we were always creating tables as the bootstrap superuser, this means that the `duckdb.postgres_user` was not able to access them. This changes that and also protects against a type of attack where someone with access to the MotherDuck account, but not full access to the Postgres server can create tables in any Postgres schema by creating them in MotherDuck and waiting for them to be synced. Now by setting `duckdb.postgres_user` only schemas that the `duckdb.postgres_user` can access will allow tables to be created in it by the background sync.

### DIFF
--- a/src/pgduckdb_background_worker.cpp
+++ b/src/pgduckdb_background_worker.cpp
@@ -32,9 +32,11 @@ extern "C" {
 #include "storage/shmem.h"
 #include "utils/builtins.h"
 #include "catalog/dependency.h"
+#include "catalog/pg_authid.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_extension.h"
+#include "utils/acl.h"
 #include "utils/guc.h"
 #include "utils/memutils.h"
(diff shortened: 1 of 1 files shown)


src/pgduckdb_background_worker.cpp
```diff
@@ -329,18 +333,59 @@ CreateTable(const char *postgres_schema_name, const char *table_name, const char
...
+		}
+
+		did_delete_table = true;
 	}
 
-	if (!SPI_run_utility_command(create_table_query)) {
+	Oid saved_userid;
+	int sec_context;
+	GetUserIdAndSecContext(&saved_userid, &sec_context);
+	SetUserIdAndSecContext(MotherDuckPostgresUser(), sec_context | SECURITY_LOCAL_USERID_CHANGE);
```
Y--, line 364: I assume there's no problem if `saved_userid==MotherDuckPostgresUser()` and there's no problem doing it because it's a trivial call?

Y-- (approved):
LGTM

PR #354, opened 2024-10-23
docs v0

wuputah:


docs/functions.md
```diff
@@ -0,0 +1,154 @@
...
+| [`duckdb.cache`](#cache) | Caches a Parquet or CSV file to disk |
+| [`duckdb.install_extension`](#install_extension) | Installs a DuckDB extension |
+| [`duckdb.raw_query`](#raw_query) | Runs a query directly against DuckDB |
+| [`duckdb.recycle_ddb`](#recycle_ddb) | TODO |
+
+## Motherduck Functions
+
+| Name | Description |
+| :--- | :---------- |
+| [`duckdb.force_motherduck_sync`](#force_motherduck_sync) | Forces a full resync of Motherduck databases and schemas to Postgres |
```
JelteF, line 30:
```suggestion
| [`duckdb.force_motherduck_sync`](#force_motherduck_sync) | Forces a full resync of Motherduck databases and schemas to Postgres (meant for debugging) |
```

docs/compilation.md
```diff
@@ -0,0 +1,6 @@
+TODO:
```
JelteF, line 1: I think it would be nice to add a README.md this directory with a table of contents, simply linking to the other files. Mainly so we can put a link to that file from the real README in the repo root, with the title "Detailed docs" or something.

wuputah:
Even more ideas:

* Add changelog / version history, though this could be in the project root (CHANGELOG.md)

* Everything in the README should probably be reiterated in further detail, e.g.:

  * Installation
  * Overview
  * Features / usage (possibly by use case / category)

JelteF:
> Add changelog / version history, though this could be in the project root (CHANGELOG.md)

I totally agree that we should add a changelog when we do the next release. But for 0.1.0 I think it could simply say something like: "The first release :tada:"

> Everything in the README should probably be reiterated in further detail, e.g.:

Agreed but seems the least critical to work on of the TODOs

wuputah:
this still has a ton of `TODO`s but it _could_ be merged if desired.

PR #356, opened 2024-10-23
ci: ignore markdown and all docs

wuputah:


wuputah:
oh nevermind, if we do this it wont merge because of the must-pass-CI rules.

PR #366, opened 2024-10-29
Heap reading optimization and fixes

mkaruza:
PR contains two commits that are supposed to be committed separately.

First commit improves performances of reading heap tuples and writing them back to duckdb output vector by converting map lookups into vector structures that are read sequentially for each row.

Second commit addresses problem of creating BufferAccessStrategy for each page read and creates single object for complete duration of scan execution.

### DIFF
--- a/include/pgduckdb/scan/postgres_scan.hpp
+++ b/include/pgduckdb/scan/postgres_scan.hpp
@@ -24,36 +24,30 @@ public:
 	TupleDesc m_tuple_desc;
 	std::mutex m_lock; // Lock for one replacement scan
 	bool m_count_tuples_only;
-	duckdb::map<duckdb::idx_t, duckdb::column_t> m_read_columns_ids;
-	duckdb::map<duckdb::idx_t, duckdb::column_t> m_output_columns_ids;
-	duckdb::TableFilterSet *m_filters = nullptr;
+	/* Postgres column id to duckdb output vector idx */
+	std::vector<duckdb::pair<duckdb::column_t, duckdb::idx_t>> m_input_columns;
+	std::vector<duckdb::TableFilter *> m_column_filters;
+	/* Duckdb output vector idx with information about postgres column id */
+	duckdb::vector<duckdb::pair<duckdb::idx_t, duckdb::column_t>> m_output_columns;
 	std::atomic<std::uint32_t> m_total_row_count;
 	duckdb::map<int, Datum> m_relation_missing_attrs;
 };
 
 class PostgresScanLocalState {
 public:
-	PostgresScanLocalState(const PostgresScanGlobalState *psgs) : m_output_vector_size(0), m_exhausted_scan(false) {
-		if (psgs->m_count_tuples_only) {
-			values = nullptr;
-			nulls = nullptr;
-		} else {
+	PostgresScanLocalState(const PostgresScanGlobalState *psgs)
+	    : m_output_vector_size(0), m_exhausted_scan(false), values(nullptr), nulls(nullptr) {
+		if (!psgs->m_count_tuples_only) {
 			/* FIXME: all calls to duckdb_malloc/duckdb_free should be changed in future */
-			const auto s = psgs->m_read_columns_ids.size();
+			const auto s = psgs->m_input_columns.size();
 			values = (Datum *)duckdb_malloc(sizeof(Datum) * s);
 			nulls = (bool *)duckdb_malloc(sizeof(bool) * s);
 		}
 	}
 
 	~PostgresScanLocalState() {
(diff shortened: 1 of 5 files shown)


include/pgduckdb/scan/postgres_scan.hpp
```diff
@@ -24,36 +24,30 @@ class PostgresScanGlobalState {
...
 	~PostgresScanLocalState() {
-		if (values) {
-			duckdb_free(values);
-			values = nullptr;
-		}
-		if (nulls) {
-			duckdb_free(nulls);
-			nulls = nullptr;
-		}
+		duckdb_free(values);
```
Y--, line 49: Won't that be a problem when `values` isn't allocated? (when `m_count_tuples_only`)

src/scan/postgres_scan.cpp
```diff
@@ -39,9 +39,25 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
...
+
+	for (auto const &[column_id, output_idx] : ordered_input_columns) {
+		duckdb::TableFilter *column_filter = nullptr;
+		if (table_filters) {
+			auto column_filter_it = table_filters->filters.find(output_idx);
+			if (column_filter_it != table_filters->filters.end()) {
+				column_filter = column_filter_it->second.get();
+			}
+		}
+		m_input_columns.push_back(duckdb::make_pair(column_id, output_idx));
```
Y--, line 59: nit:
```suggestion
		m_input_columns.emplace_back(column_id, output_idx);
```
(same below) for `m_output_columns`

src/pgduckdb_types.cpp
```diff
@@ -1088,35 +1088,33 @@ InsertTupleIntoChunk(duckdb::DataChunk &output, duckdb::shared_ptr<PostgresScanG
...
-	for (auto const &[columnIdx, valueIdx] : scan_global_state->m_read_columns_ids) {
-		values[valueIdx] =
-		    HeapTupleFetchNextColumnDatum(scan_global_state->m_tuple_desc, tuple, heap_tuple_read_state, columnIdx + 1,
-		                                  &nulls[valueIdx], scan_global_state->m_relation_missing_attrs);
-		if (scan_global_state->m_filters &&
-		    (scan_global_state->m_filters->filters.find(valueIdx) != scan_global_state->m_filters->filters.end())) {
-			auto &filter = scan_global_state->m_filters->filters[valueIdx];
-			const auto valid_tuple = ApplyValueFilter(*filter, values[valueIdx], nulls[valueIdx],
-			                               scan_global_state->m_tuple_desc->attrs[columnIdx].atttypid);
+	for (auto const &[column_id, output_column_idx] : scan_global_state->m_input_columns) {
```
Y--, line 1091: Should we also rename `column_id` to `attr_id` or `input_column_id`?

src/pgduckdb_types.cpp
```diff
@@ -1088,35 +1088,33 @@ InsertTupleIntoChunk(duckdb::DataChunk &output, duckdb::shared_ptr<PostgresScanG
...
 		}
-
 	}
 
 	/* Write tuple columns in output vector. */
-	for (idx_t idx = 0; idx < scan_global_state->m_output_columns_ids.size(); idx++) {
-		auto &result = output.data[idx];
-		idx_t output_column_idx = scan_global_state->m_read_columns_ids[scan_global_state->m_output_columns_ids[idx]];
+	int i = 0;
+	for (auto const &[output_column_idx, column_id] : scan_global_state->m_output_columns) {
```
Y--, line 1107: Same here for `column_id`

include/pgduckdb/scan/postgres_scan.hpp
mkaruza: AFACT, free should do nothing if variable is null (we initialize this in constructor to null value)

JelteF (approved):
These look like good changes to me. I'm still planning to look at refactoring this code a bit more to make the different indexes and arrays easier to understand. But for that I need to try a few things and play around with the code. The current PR is already much easier to follow than before.

JelteF:
Let's add PR numbers to the commit though for easy reference if you're not doing a squash merge.

include/pgduckdb/scan/postgres_scan.hpp
Y--: With the stdlib implementation of `free` (which is what `duckdb_free` is by default AFAIK) yeah, but I don't know if that's a hard requirement. Anyway, not a big deal, we'll see quickly enough if that happen to change :-)

Y-- (approved):

PR #377, opened 2024-10-30
Improve C++ exception handling

Y--:
Forked off https://github.com/duckdb/pg_duckdb/pull/254

This PR reworks `DuckDBFunctionGuard` to make invocation easier:
* no need to specify type in template invocation
* no need to provide the function name

It also introduces a new `DECLARE_PG_FUNCTION(func_name)` macro which wraps a given function and make it a safe `PG_FUNCTION` that catches any C++ exceptions. This works by expecting a `func_name_cpp` (note the `_cpp` suffix) function to be implemented.

It also renames existing `_Unsafe` functions into `_Cpp` for consistency.

### DIFF
--- a/src/pgduckdb_options.cpp
+++ b/src/pgduckdb_options.cpp
@@ -178,36 +178,36 @@ DuckdbCacheObject(Datum object, Datum type) {
 
 extern "C" {
 
-PG_FUNCTION_INFO_V1(install_extension);
+DECLARE_PG_FUNCTION(install_extension);
+DECLARE_PG_FUNCTION(pgduckdb_raw_query);
+DECLARE_PG_FUNCTION(cache);
+DECLARE_PG_FUNCTION(pgduckdb_recycle_ddb);
+
 Datum
-install_extension(PG_FUNCTION_ARGS) {
+install_extension_cpp(PG_FUNCTION_ARGS) {
 	Datum extension_name = PG_GETARG_DATUM(0);
 	bool result = pgduckdb::DuckdbInstallExtension(extension_name);
 	PG_RETURN_BOOL(result);
 }
 
-PG_FUNCTION_INFO_V1(pgduckdb_raw_query);
 Datum
-pgduckdb_raw_query(PG_FUNCTION_ARGS) {
+pgduckdb_raw_query_cpp(PG_FUNCTION_ARGS) {
 	const char *query = text_to_cstring(PG_GETARG_TEXT_PP(0));
-	typedef duckdb::unique_ptr<duckdb::QueryResult> (*DuckDBQueryOrThrow)(const std::string &);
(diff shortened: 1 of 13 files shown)


src/pgduckdb_options.cpp
```diff
@@ -178,36 +178,36 @@ DuckdbCacheObject(Datum object, Datum type) {
 
 extern "C" {
 
-PG_FUNCTION_INFO_V1(install_extension);
+DECLARE_PG_FUNCTION(install_extension);
+DECLARE_PG_FUNCTION(pgduckdb_raw_query);
+DECLARE_PG_FUNCTION(cache);
+DECLARE_PG_FUNCTION(pgduckdb_recycle_ddb);
```
JelteF, line 184: nit: it seems nice to keep these close to the declaration like they were before

JelteF: It might even be nice to take the macro one step further end have the definition be like the below instead. Because it is a bit confusing that here we declare the function with the `_cpp` suffix but then the actual function uses the `_cpp` suffix. I think it would be nice to hide that implementation detail:

```c
DECLARE_PG_FUNCTION(install_extension) {
	Datum extension_name = PG_GETARG_DATUM(0);
	bool result = pgduckdb::DuckdbInstallExtension(extension_name);
	PG_RETURN_BOOL(result);
}
```

src/pgduckdb_ruleutils.cpp
```diff
@@ -27,14 +28,19 @@ extern "C" {
...
 		return nullptr;
 	}
+
 	auto func_name = get_func_name(function_oid);
 	return psprintf("system.main.%s", quote_identifier(func_name));
 }
 
+char *
+pgduckdb_function_name(Oid function_oid) {
+	return InvokeCPPFunc(pgduckdb_function_name_cpp, function_oid);
```
JelteF, line 42: Why is this change needed? it seems like that the function doesn't actually run any C++ code.

src/pgduckdb_utils.cpp
```diff
@@ -17,68 +18,56 @@ extern "C" {
...
 DuckDBQueryOrThrow(duckdb::ClientContext &context, const std::string &query) {
-	const char *error_message = nullptr;
-	{
-		auto res = context.Query(query, false);
-		if (!res->HasError()) {
-			return res;
-		}
-
-		error_message = pstrdup(res->GetError().c_str());
+	auto res = context.Query(query, false);
```
JelteF, line 66: This seems a big enough error in behaviour that I think it might be good to it in a separate PR so we can do some additional testing.

src/pgduckdb_ruleutils.cpp
Y--: Ah right - it calls `pgduckdb::IsDuckdbOnlyFunction` which in itself isn't a C++-exception throwing function

JelteF (approved):
LGTM now

PR #381, opened 2024-11-01
Simplify `DuckDBQueryOrThrow`

Y--:
Now that we correctly handle C++ exceptions, simplify the `DuckDBQueryOrThrow` helper.

### DIFF
--- a/src/pgduckdb_utils.cpp
+++ b/src/pgduckdb_utils.cpp
@@ -63,21 +63,11 @@ CreateOrGetDirectoryPath(const char* directory_name) {
 
 duckdb::unique_ptr<duckdb::QueryResult>
 DuckDBQueryOrThrow(duckdb::ClientContext &context, const std::string &query) {
-	const char *error_message = nullptr;
-	{
-		auto res = context.Query(query, false);
-		if (!res->HasError()) {
-			return res;
-		}
-
-		error_message = pstrdup(res->GetError().c_str());
+	auto res = context.Query(query, false);
+	if (res->HasError()) {
+		res->ThrowError();
 	}
-
-	if (error_message) {
-		elog(ERROR, "(PGDuckDB/DuckDBQuery) %s", error_message);
-	}
-
-	return nullptr; // unreachable
+	return res;
 }
 
 duckdb::unique_ptr<duckdb::QueryResult>


src/pgduckdb_utils.cpp
```diff
@@ -63,21 +63,11 @@ CreateOrGetDirectoryPath(const char* directory_name) {
...
 DuckDBQueryOrThrow(duckdb::ClientContext &context, const std::string &query) {
-	const char *error_message = nullptr;
-	{
-		auto res = context.Query(query, false);
-		if (!res->HasError()) {
-			return res;
-		}
-
-		error_message = pstrdup(res->GetError().c_str());
+	auto res = context.Query(query, false);
```
JelteF, line 66: So what I'm thinking: I think there are going to be places where we use Postgres style code, but do need to fire off one duckdb query. So can we have functions for both use cases: `DuckDBQueryOrThrow(std::string)` and `DuckDBQueryOrElog(const char*)`

Y--: I was actually hoping to never have the mix of C++ & C code and always have a pure "C++" way. For example, this is what I did in this PR https://github.com/duckdb/pg_duckdb/pull/383

JelteF: That makes sense, but are you saying C code should never run any queries in DuckDB? I was more thinking that `DuckDBQueryOrElog` would be on of our "glue" functions. i.e. just an small wrapper function around `DuckDBQueryOrThrow`, which calls `InvokeCPPFunc` for you and converts the `const char *` to a `std::string`. So then you have a C way to run a query in DuckDB.

Y--: Yeah that make sense indeed.
For now, I prefer to throw deep in the stack more than "elog" cause the latter can lead to memory leak (while the former should be fully handled at top level). But I can indeed add `DuckDBQueryOrElog` as soon as it make sense!

JelteF (approved):
Alright let's merge this. I didn't do extensive testing if this breaks existing stuff. But it shouldn't given you added all the InvokeCPP calls. And if it doesn't we'll automatically find out during development. Seems better to merge this soon, so it has time to bake.

PR #383, opened 2024-11-01
Rewrite copy helper in C form

Y--:
Copy utilities was mostly using Postgres functions, so this PR removes any C++ function calls.
It also moves out the DuckDB query execution off the function itself, which only construct the query string to be executed.

### DIFF
--- a/src/utility/copy.cpp
+++ b/src/utility/copy.cpp
@@ -1,6 +1,6 @@
-#include "duckdb.hpp"
-#include "pgduckdb/pgduckdb_utils.hpp"
 #include "pgduckdb/utility/copy.hpp"
+
+#include <inttypes.h>
 
 extern "C" {
 #include "postgres.h"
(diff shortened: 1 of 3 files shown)


src/utility/copy.cpp
```diff
@@ -73,22 +60,25 @@ CreateRelationCopyString(ParseState *pstate, CopyStmt *copy_stmt) {
...
-			}
+	appendStringInfoString(info, pgduckdb_relation_name(relid));
+	if (!copy_stmt->attlist) {
+		return;
+	}
+
+	ListCell *lc;
+	appendStringInfo(info, "(");
+	bool first = true;
+	foreach (lc, copy_stmt->attlist) {
```
JelteF, line 71: Let's change this to use the modern `foreach` macros instead while we're at it.
```suggestion
	foreach_node (String, attr, copy_stmt->attlist) {
```

src/utility/copy.cpp
```diff
@@ -118,48 +108,50 @@ CheckQueryPermissions(Query *query, const char *query_string) {
...
-				options_stream << NameListToQuotedString((List *)defel->arg);
+				appendStringInfoString(info, NameListToQuotedString((List *)defel->arg));
 				break;
 			case T_A_Star:
-				options_stream << "*";
+				appendStringInfo(info, "*");
 				break;
 			default:
-				throw std::runtime_error("unexpected node type in COPY: " + std::to_string((int)nodeTag(defel->arg)));
+				// elog(ERROR, "Expected single table to be created, but found %" PRIu64, static_cast<uint64_t>(SPI_processed));
```
JelteF, line 148: Is this commented out line here on purpose?

src/utility/copy.cpp
```diff
@@ -18,11 +19,6 @@ extern "C" {
 #include "pgduckdb/pgduckdb_ruleutils.h"
 }
 
-#include "pgduckdb/utility/copy.hpp"
-#include "pgduckdb/pgduckdb_duckdb.hpp"
-#include "pgduckdb/vendor/pg_list.hpp"
```
JelteF, line 23: Based on the test failures. I think we need to continue to include this one.

Y--: Ah I should have read this earlier, it took me a minute to figure it out 😅

src/utility/copy.cpp
Y--: Nope, a leftover... Thanks!

JelteF (approved):

PR #384, opened 2024-11-01
Remove unused headers

Y--:
In preparation for partitioning the code between C/PG & C++/DuckDB, removes unused headers.

### DIFF
--- a/src/pgduckdb_utils.cpp
+++ b/src/pgduckdb_utils.cpp
@@ -1,19 +1,13 @@
 #include "pgduckdb/pgduckdb_utils.hpp"
 
-extern "C" {
-#include "postgres.h"
-#include "miscadmin.h"
-#include "lib/stringinfo.h"
-#include "storage/fd.h"
-#include "executor/spi.h"
-}
-
 #include <string>
 
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
-#include <errno.h>
+
+// FIXME - this file should not depend on PG, rather DataDir should be provided
+extern PGDLLIMPORT char *DataDir;
 
 namespace pgduckdb {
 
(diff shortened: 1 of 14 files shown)


JelteF (commented):
Compilation is failing on PG15 and PG16, I guess some headers were needed for those.

src/pgduckdb_utils.cpp
```diff
@@ -1,19 +1,13 @@
...
-
 #include <string>
 
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
-#include <errno.h>
+
+// FIXME - this file should not depend on PG, rather DataDir should be provided
+extern PGDLLIMPORT char *DataDir;
```
JelteF, line 10: Is this to avoid including any postgres headers here?

JelteF: I'm surprised `PGDLLIMPORT` is available without importing postgres headers. I guess that's because `pgduckdb_utils.hpp` is still importing `postgres.h`.

src/pgduckdb_utils.cpp
```diff
@@ -50,7 +44,7 @@ CreateOrGetDirectoryPath(const char* directory_name) {
 	const auto duckdb_data_directory = oss.str();
 
 	if (!CheckDirectory(duckdb_data_directory)) {
-		if (MakePGDirectory(duckdb_data_directory.c_str()) == -1) {
+		if (mkdir(duckdb_data_directory.c_str(), S_IRWXU) == -1) {
```
JelteF, line 47: Should use the `pg_dir_create_mode` instead. That global is only set during startup, so accessing it is fine even when using multiple threads. A Postgres database directory can be configured such that the group also has permissions:

```
/*
 * Set create modes and mask to use when writing to PGDATA based on the data
 * directory mode passed.  If group read/execute are present in the mode, then
 * create modes and mask will be relaxed to allow group read/execute on all
 * newly created files and directories.
 */
void
SetDataDirectoryCreatePerm(int dataDirMode)
```

JelteF (changes requested):
Overall looks good, only the `mode` for `mkdir` is something that we should change.

Y--:
> Overall looks good, only the `mode` for `mkdir` is something that we should change.

Yeah I'm not a big fan of it either, but that's the only PG dependency left. I've copied it over from the source. The other approach would be to thread it through but that feel overkill. Wdyt?

JelteF:
What if you simply add `extern int pg_dir_create_mode;`

JelteF:
Or include `common/file_perm.h` which is pretty small.

JelteF:
> Or include common/file_perm.h which is pretty small.

Actually I think this is a good solution. Importing headers from `common` should be safe. These ones are used by client binaries like psql, so they never call `elog`.

JelteF (approved):

PR #386, opened 2024-11-04
Remove unused headers

Y--:
Follow-up of https://github.com/duckdb/pg_duckdb/pull/384 
But this one introduces some forward declarations of PG, so wanted to separate the discussion.


include/pgduckdb/vendor/pg_fwd.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
```
JelteF, line 1: *moving comment here from #387*

Can we rename this this file to something else? Like postgres_declarations. pg_fwd is a bit confusing to me because pg_fdw is a very commonly known thing.

Also, seems like this should be .hpp instead of .h.

Also, I think it would be better not to put this in the vendor directory. That directory is (imho) meant for files that we copy pasted for code that we copied for more than ~90% from Postgres. That's also why we explicitly disable clang-format, which would be fine here.

include/pgduckdb/vendor/pg_fwd.h
```diff
@@ -0,0 +1,15 @@
...
+
+// Postgres forward declarations
```
JelteF, line 3: Could use a bit more explanation why we have this file, and when we should add new things to it.

JelteF (approved):
Overall seems like a good change. Some small comments about comments/naming of the new forward declaration file.

include/pgduckdb/vendor/pg_fwd.h
Y--: Whole thread here https://github.com/duckdb/pg_duckdb/pull/387#discussion_r1827624201

JelteF (approved):

PR #395, opened 2024-11-05
Improve variable naming in the Postgres scan code

JelteF:
This mostly makes naming of variables clearer, so that it's harder to get
confused about the different indexes. It also changes a two small things:

1. Use `AttrNumber` to store postgres column indexes, these are always 1-based
2. Simplify some of the assignments `HeapTupleFetchNextColumnDatum`


Y-- (commented):
LGTM, I'll let @mkaruza stamp as he worked most on this part of code

mkaruza (approved):

PR #397, opened 2024-11-06
Support postgresql 14

Leo-XM-Zeng:
Support postgresql 14

### DIFF
--- a/src/utility/copy.cpp
+++ b/src/utility/copy.cpp
@@ -72,7 +72,12 @@ appendCreateRelationCopyString(StringInfo info, ParseState *pstate, CopyStmt *co
 
 	appendStringInfo(info, "(");
 	bool first = true;
+
+#if PG_VERSION_NUM >= 150000
 	foreach_node(String, attr, copy_stmt->attlist) {
+#else
+	foreach_ptr(Value, attr, copy_stmt->attlist) {
+#endif
 		if (first) {
 			first = false;
 		} else {
@@ -135,7 +140,9 @@ appendCreateCopyOptions(StringInfo info, CopyStmt *copy_stmt) {
 			switch (nodeTag(defel->arg)) {
 			case T_Integer:
 			case T_Float:
+#if PG_VERSION_NUM >= 150000
 			case T_Boolean:
+#endif
 				appendStringInfoString(info, defGetString(defel));
 				break;
 			case T_String:
@@ -220,7 +227,11 @@ MakeDuckdbCopyQuery(PlannedStmt *pstmt, const char *query_string, struct QueryEn
 		raw_stmt->stmt_location = pstmt->stmt_location;
 		raw_stmt->stmt_len = pstmt->stmt_len;
 
+#if PG_VERSION_NUM >= 150000
 		List *rewritten = pg_analyze_and_rewrite_fixedparams(raw_stmt, query_string, NULL, 0, NULL);
+#else
+		List *rewritten = pg_analyze_and_rewrite(raw_stmt, query_string, NULL, 0, NULL);
+#endif
 		CheckRewritten(rewritten);
 
 		Query *query = linitial_node(Query, rewritten);
--- a/test/pycheck/temporary_table_test.py
+++ b/test/pycheck/temporary_table_test.py
@@ -1,0 +1,24 @@
+from .utils import Cursor, PG_MAJOR_VERSION
+
+import pytest
+import psycopg.errors
+
+
+def test_temporary_table_alter_table(cur: Cursor):
+    cur.sql("CREATE TEMP TABLE t(a int) USING duckdb")
+    with pytest.raises(psycopg.errors.InternalError_):
+        cur.sql("ALTER TABLE t ADD COLUMN b int")
+
+    if PG_MAJOR_VERSION >= 15:
+        with pytest.raises(psycopg.errors.FeatureNotSupported):
+            cur.sql("ALTER TABLE t SET ACCESS METHOD heap")
+
+    with pytest.raises(psycopg.errors.FeatureNotSupported):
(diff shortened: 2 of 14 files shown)


src/utility/copy.cpp
```diff
@@ -81,6 +82,18 @@ appendCreateRelationCopyString(StringInfo info, ParseState *pstate, CopyStmt *co
 
 		appendStringInfoString(info, quote_identifier(strVal(attr)));
 	}
+#else
```
JelteF, line 85: This `#if/#else` shouldn't be necessary afaict. These pieces of code should have the same behaviour.

test/pycheck/temporary_table_test.py
```diff
@@ -0,0 +1,22 @@
+from .utils import Cursor, PG_MAJOR_VERSION, eprint
+
+
+def test_temporary_table_set_access_method(cur: Cursor):
+    if PG_MAJOR_VERSION >= 15:
+        cur.sql("CREATE TEMP TABLE t(a int)")
+        cur.sql("ALTER TABLE t ADD COLUMN b int")
+
+        try:
```
JelteF, line 9: Instead of the `try` this should use `with pytest.raises(...):` that way the test will fail if the intended exception is not thrown.

test/pycheck/temporary_table_test.py
```diff
@@ -0,0 +1,22 @@
+from .utils import Cursor, PG_MAJOR_VERSION, eprint
+
+
+def test_temporary_table_set_access_method(cur: Cursor):
+    if PG_MAJOR_VERSION >= 15:
```
JelteF, line 5: Can you use this instead to skip for lower PG versions?
```suggestion
@pytest.mark.skipif("PG_MAJOR_VERSION < 15", reason="SET ACCESS METHOD was introduced in PG15")
def test_temporary_table_set_access_method(cur: Cursor):
```

src/utility/copy.cpp
Leo-XM-Zeng: PG14 does not have a String structure, so foreach_node cannot be used.
[https://github.com/postgres/postgres/blob/REL_14_STABLE/src/include/nodes/value.h](https://github.com/postgres/postgres/blob/REL_14_STABLE/src/include/nodes/value.h)
[https://github.com/postgres/postgres/blob/REL_15_STABLE/src/include/nodes/value.h](https://github.com/postgres/postgres/blob/REL_15_STABLE/src/include/nodes/value.h)

test/pycheck/temporary_table_test.py
Leo-XM-Zeng: Ok, this seems to be OK, this is my first time using python, thanks for the guidance.

src/utility/copy.cpp
JelteF: Hmm okay, now I understand. Can you then change it to something like this:

```c
#if PG_VERSION_NUM >= 150000
	foreach_node(String, attr, copy_stmt->attlist) {
# else
	foreach_ptr(Value, attr, copy_stmt->attlist) {
#endif
```

Leo-XM-Zeng: Wow, that's a good idea.

test/pycheck/temporary_table_test.py
```diff
@@ -0,0 +1,21 @@
...
+
+import pytest
+import psycopg.errors
+
+
+@pytest.mark.skipif(
+    "PG_MAJOR_VERSION < 15", reason="SET ACCESS METHOD was introduced in PG15"
+)
+def test_temporary_table_set_access_method(cur: Cursor):
+    cur.sql("CREATE TEMP TABLE t(a int)")
```
JelteF, line 11: In the previous test the duckdb access method was the default. In this test it is not. Which is why the first `SET ACCESS METHOD` call is passing.
```suggestion
    cur.sql("CREATE TEMP TABLE t(a int) USING duckdb")
```

Leo-XM-Zeng: Ok, that's done

test/pycheck/temporary_table_test.py
```diff
@@ -0,0 +1,27 @@
+from .utils import Cursor, PG_MAJOR_VERSION
+
+import pytest
+import psycopg.errors
+
+
+@pytest.mark.skipif(
+    "PG_MAJOR_VERSION < 15", reason="SET ACCESS METHOD was introduced in PG15"
+)
+def test_temporary_table_set_access_method(cur: Cursor):
```
JelteF, line 10: Sorry for the back and forth here. But given your latest changes (which are good changes), we should actually execute this test for PG14, just not the `SET ACCESS METHOD` part. And the name of the function also doesn't reflect what this is testing anymore now.
```suggestion
def test_temporary_table_alter_table(cur: Cursor):
```

Leo-XM-Zeng: It's okay. Under your guidance, I've learned a lot of better things

src/utility/copy.cpp
```diff
@@ -135,7 +140,7 @@ appendCreateCopyOptions(StringInfo info, CopyStmt *copy_stmt) {
 			switch (nodeTag(defel->arg)) {
 			case T_Integer:
 			case T_Float:
-			case T_Boolean:
```
JelteF, line 138: This change is incorrect, `T_Boolean` and `T_BooleanTest` are very different nodes. The `Boolean` node type did not exist in PG14, it was using an integer or a string before so we can simply skip it there.

Leo-XM-Zeng: It's my fault.  Sorry, I didn't notice a usage error.  It was very careless of me.  Thank you for your correction.

JelteF:
Thanks @Leo-XM-Zeng I pushed this now as two separate commits to main.

PR #401, opened 2024-11-07
Fix glibc assertion

### LINKED ISSUE #398: Latest version aborts on simple join.
### What happens?

Postgres process terminates with signal 6 upon a simple join.
See script below.

Built from fb271ec0495b612497a9d6e23e1cbfedf1453e52 today.

Sorry I do not have a postgres debug build.

### To Reproduce

psql session:
```
> SET duckdb.force_execution TO true;
SET

> create table t2(pk integer not null);
CREATE TABLE

> create table t1(pk integer not null);
CREATE TABLE

> insert into t1 values(1);
INSERT 0 1

> explain select count(r2.pk) from t1 r1 join t2 r2 on r2.pk = r1.pk where r1.pk<= 90;
QUERY PLAN
------------------------------------------------------------
Custom Scan (DuckDBScan)  (cost=0.00..0.00 rows=0 width=0)
DuckDB Execution Plan:

┌───────────────────────────┐
│    UNGROUPED_AGGREGATE    │
│    ────────────────────   │
│   Aggregates: count(#0)   │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             pk            │
│                           │
│         ~102 Rows         │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         HASH_JOIN         │
│    ────────────────────   │
│      Join Type: INNER     │
│    Conditions: pk = pk    ├──────────────┐
│                           │              │
│         ~102 Rows         │              │
└─────────────┬─────────────┘              │
┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│     POSTGRES_SEQ_SCAN     ││     POSTGRES_SEQ_SCAN     │
│    ────────────────────   ││    ────────────────────   │
│         Function:         ││         Function:         │
│     POSTGRES_SEQ_SCAN     ││     POSTGRES_SEQ_SCAN     │
│                           ││                           │
│      Projections: pk      ││      Projections: pk      │
│                           ││                           │
│          Filters:         ││          Filters:         │
│ pk<=90 AND pk IS NOT NULL ││ pk<=90 AND pk IS NOT NULL │
│                           ││                           │
│         ~510 Rows         ││         ~510 Rows         │
└───────────────────────────┘└───────────────────────────┘


(38 rows)

> explain analyze select count(r2.pk) from t1 r1 join t2 r2 on r2.pk = r1.pk;
!?>
```

From the postgres log:
```
/usr/include/c++/14/bits/stl_vector.h:1130: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = long unsigned int; _Alloc = pgduckdb::DuckDBMallocator<long unsigned int>; reference = long unsigned int&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
2024-11-06 11:37:36.479 PST [12403] LOG:  server process (PID 14403) was terminated by signal 6: Aborted
2024-11-06 11:37:36.479 PST [12403] DETAIL:  Failed process was running: explain analyze select count(r2.pk) from t1 r1 join t2 r2 on r2.pk = r1.pk where r1.pk <= 90;
2024-11-06 11:37:36.479 PST [12403] LOG:  terminating any other active server processes
2024-11-06 11:37:36.480 PST [14582] FATAL:  the database system is in recovery mode
```

### OS:

Linux

### pg_duckdb Version:

1.1.3

### Postgres Version:

16.3

### Hardware:

_No response_

### Full Name:

Lars Hofhansl

### Affiliation:

private

Yes

- [X] Yes, I have

- [X] Yes, I have

JelteF:
We were using `reserve` instead of `resize` to size our vector `values` and `nulls` vectors. This meant that we were reading out of bounds of the vector its size, but still within its capacity.

We cannot use `push_back()` to insert items because we need to insert them in order that they appear in the tuple, not in the order that DuckDB expects them: https://github.com/duckdb/pg_duckdb/blob/9513d2a083d11436c7c270d6339f88388600dafa/src/pgduckdb_types.cpp#L1083-L1084

This also enables glibc assertions whenever they are available in Debug builds because that is how this issue was caused.

Fixes #398


Y-- (approved):
Thanks for the fix!

PR #408, opened 2024-11-08
Add logger wrapper

Y--:
In an effort to further separate C++ related code and PG related one, we want to have a logger that doesn't bring PG includes with it.
This PR creates a simple wrapper around PG's `elog` macro and re-declare the constant levels.


src/logger.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wbuiltin-macro-redefined"
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wbuiltin-macro-redefined"
+
+// Redefine them to match the pgduckdb logger so it uses the provided filename and lineno
+#define __FILE__ filename
+#define __LINE__ lineno
```
Y--, line 17: This is not ideal but the only way I could find to wrap the `elog` macro while preserving original file name & line.

include/pgduckdb/logger.hpp
```diff
@@ -0,0 +1,27 @@
...
+#define WARNING_CLIENT_ONLY	20
+#define ERROR		21
+#define PGERROR		21
+#define FATAL		22
+#define PANIC		23
+
+void pgduckdb_logger(int log_level, const char *filename, int lineno, const char *pattern, ...);
+
+#define pdlog(level, ...) \
+	pgduckdb_logger(level, __FILE__, __LINE__, __VA_ARGS__);
```
JelteF, line 25: There's an important difference between this macro and the postgres elog macro: The postgres elog macro takes core to not evaluate the arguments if the is not sufficiently high to show a log. It does that this way:
```c
		if (errstart(elevel_, domain)) \
			__VA_ARGS__, errfinish(__FILE__, __LINE__, __func__); \
```

include/pgduckdb/logger.hpp
```diff
@@ -0,0 +1,27 @@
...
+#define LOG			15
+#define INFO		17
+#define NOTICE		18
+#define WARNING		19
+#define PGWARNING	19
+#define WARNING_CLIENT_ONLY	20
+#define ERROR		21
+#define PGERROR		21
+#define FATAL		22
+#define PANIC		23
```
JelteF, line 20: Shouldn't we comment these out/remove them. We don't want people to use ERROR and higher with this function right?

JelteF: Maybe even assert somewhere that `WARNING_CLIENT_ONLY` is the highest used code.

include/pgduckdb/logger.hpp
```diff
@@ -0,0 +1,48 @@
...
+		static_assert(elevel >= DEBUG5 && elevel <= WARNING_CLIENT_ONLY, "Invalid error level"); \
+		if (errstart(elevel, domain)) \
+			__VA_ARGS__, errfinish(__FILE__, __LINE__, __func__); \
+	} while(0)
+
+
+#define pd_ereport(elevel, ...)	\
+	pd_ereport_domain(elevel, TEXTDOMAIN, __VA_ARGS__)
+
+#define pdlog(elevel, ...)  \
```
JelteF, line 45: `pd` stands for **p**g_**d**uckdb?

I don't think we use that acronym anywhere else. I do agree that we want something short, so maybe it's the right time to start using it.

let's stay consistent then with the `_` at least.

include/pgduckdb/logger.hpp
```diff
@@ -0,0 +1,48 @@
...
+#else
+#define pd_prevent_errno_in_scope()
+#endif
+
+#define pd_ereport_domain(elevel, domain, ...)	\
+	do { \
+		pd_prevent_errno_in_scope(); \
+		static_assert(elevel >= DEBUG5 && elevel <= WARNING_CLIENT_ONLY, "Invalid error level"); \
+		if (errstart(elevel, domain)) \
+			__VA_ARGS__, errfinish(__FILE__, __LINE__, __func__); \
```
JelteF, line 38: One thing I'm now realizing: I think we cannot call `errstart` safely without the global lock. It internally calls `get_error_stack_entry` which increments a global counter... (which ofcourse is not an atomic increment). Also `errfinish` is using `appendStringInfo`, which does allocations, and the postgres allocator is not thread safe. So basically what I'm saying: We cannot safely log without a global lock. 

We could safely check the `elevel` though without the lock, using `message_level_is_interesting()`. And only take the lock when the that function returns true.

JelteF (changes requested):

include/pgduckdb/logger.hpp
Y--: Yeah exactly - I wanted something as concise as possible (also why I omitted the `_` since it will be so pervasive. But agree it's probably better to stay consistent.

include/pgduckdb/logger.hpp
Y--: That's a great point. I didn't pay attention to that yet. I'll rework it to make it thread-safe.

JelteF (approved):

Issue #409, opened 2024-11-08
Server crashes when scanning partitioned table


dpxcc:
### What happens?

Server crashes when executing the script below to scan partitioned table

### To Reproduce

```sql
SET duckdb.force_execution = ON;
CREATE TABLE t (a INT) PARTITION BY LIST (a);
CREATE TABLE s (b INT);
SELECT * FROM t JOIN s ON a = b; -- crash
```
backtrace
```cpp
#0  __pthread_kill_implementation (threadid=281473108209920, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x0000ffffa30cf254 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x0000ffffa308a67c in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x0000ffffa3077130 in __GI_abort () at ./stdlib/abort.c:79
#4  0x0000aaaabfa25098 in ExceptionalCondition (conditionName=0xaaaabfc5ec70 "false", fileName=0xaaaabfc5e270 "bufmgr.c", lineNumber=3933) at assert.c:66
#5  0x0000aaaabf7bedc0 in RelationGetNumberOfBlocksInFork (relation=0xffffa3037fe0, forkNum=MAIN_FORKNUM) at bufmgr.c:3933
#6  0x0000ffffa361bc5c in pgduckdb::HeapReaderGlobalState::HeapReaderGlobalState (rel=0xffffa3037fe0, this=0xffff70005a30) at include/pgduckdb/scan/heap_reader.hpp:18
#7  __gnu_cxx::new_allocator<pgduckdb::HeapReaderGlobalState>::construct<pgduckdb::HeapReaderGlobalState, RelationData*&> (__p=0xffff70005a30, this=<optimized out>) at /usr/include/c++/11/ext/new_allocator.h:162
#8  std::allocator_traits<std::allocator<pgduckdb::HeapReaderGlobalState> >::construct<pgduckdb::HeapReaderGlobalState, RelationData*&> (__p=0xffff70005a30, __a=...) at /usr/include/c++/11/bits/alloc_traits.h:516
[...]

Issue #19, opened 2024-04-24
support scanning partitioned tables


wuputah:
When obtaining metadata for a partitioned table, several tables are returned - the parent and then its children. Support scanning each table in turn (or in parallel).

For now, we can assume that each partition is a heap table, but in the future, each table might have a different storage method.

ritwizsinha:
@JelteF  I would like to start working to support this feature, if that is okay

JelteF:
I think this should already be supported, now that #477 is merged. If that's indeed the case all that's really needed is some tests.

JelteF:
Yeah #477 included tests for partitioned tables already: https://github.com/duckdb/pg_duckdb/blob/de3735a963704a9bebd6cacd52f03e2ede572219/test/regression/expected/scan_postgres_tables.out#L162-L253

So closing this.

PR #412, opened 2024-11-09
Handle exceptions caused by partitioned tables

Leo-XM-Zeng:
Because we do not yet have good support for partitioned tables, the following code triggers assertions. #409

```c
/*
 * RelationGetNumberOfBlocksInFork
 *		Determines the current number of pages in the specified relation fork.
 *
 * Note that the accuracy of the result will depend on the details of the
 * relation's storage. For builtin AMs it'll be accurate, but for external AMs
 * it might not be.
 */
BlockNumber
RelationGetNumberOfBlocksInFork(Relation relation, ForkNumber forkNum)
{
	if (RELKIND_HAS_TABLE_AM(relation->rd_rel->relkind))
	{
		/*
		 * Not every table AM uses BLCKSZ wide fixed size blocks. Therefore
		 * tableam returns the size in bytes - but for the purpose of this
		 * routine, we want the number of blocks. Therefore divide, rounding
		 * up.
		 */
		uint64		szbytes;

		szbytes = table_relation_size(relation, forkNum);

		return (szbytes + (BLCKSZ - 1)) / BLCKSZ;
	}
	else if (RELKIND_HAS_STORAGE(relation->rd_rel->relkind))
	{
		return smgrnblocks(RelationGetSmgr(relation), forkNum);
	}
	else
		Assert(false);

	return 0;					/* keep compiler quiet */
}
``` 
**What does this PR do?**

When accessing the partition table, we temporarily let PG handle it instead of DuckDB.
This temporarily avoids the unavailability of the system due to an assertion.


Leo-XM-Zeng:
Of course we're going to have to deal with #19 eventually.

src/pgduckdb_hooks.cpp
```diff
@@ -52,6 +52,23 @@ IsCatalogTable(List *tables) {
 	return false;
 }
 
+static bool
+IsPartitionedTable(List *tables) {
```
JelteF, line 56: I think `Is` is the wrong prefix here and the argument name `tables` is also confusing. We pass it a list of tables RTEs not a single table. I realize the problem is the case for `IsCatalogTable`, but we should change this there too (not in this PR though).
```suggestion
ContainsPartitionedTable(List *rtes) {
```

src/pgduckdb_hooks.cpp
```diff
@@ -52,6 +52,23 @@ IsCatalogTable(List *tables) {
...
+	foreach_node(RangeTblEntry, table, tables) {
+		if (table->rtekind == RTE_SUBQUERY) {
+			/* Check whether the table in the subquery is a partitioned table */
```
JelteF, line 59:
```suggestion
			/* Check whether any table in the subquery is a partitioned table */
```

JelteF (commented):
Thank you for the fix. Some small naming suggestions, but other than that this looks good.

JelteF (approved):

src/pgduckdb_hooks.cpp
Leo-XM-Zeng: Updated the IsCatalogTable function information #417

PR #413, opened 2024-11-11
Cache DuckDB connection object and hook into PG transaction life cycle

JelteF:
Instead of creating a new `duckdb::Connection` over and over again, this starts caching the DuckDB connection. For this to work correctly in case of errors we need to link DuckDB and Postgres its transaction life cycle together. Otherwise DuckDB state from a previous PG query might accidentally survive until the next PG. This is also in preparation of supporting multi-statement transactions.

There was one more problem to solve: This PR changes the lifecycle of the `duckdb::Transaction`. 
Before the lifecycle of the different components was as follows, from long-lived to short-lived:
1. Postgres transaction
2. Postgres query
3. DuckDB transaction
4. DuckDB query

Now 2 and 3 are swapped:
1. Postgres transaction
2. DuckDB transaction
3. Postgres query
4. DuckDB query

This causes problems, because the `Relation` from the Postgres relcache should always be closed before the end of the postgres query. Previously we bound `SchemaItems` cache to the lifecycle of the DuckDB transaction, which now outlives the postgres query. To solve this we change its lifecycle to be bound the the DuckDB query.

### DIFF
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -18,7 +18,6 @@ public:
 	}
 
 	static duckdb::unique_ptr<duckdb::Connection> CreateConnection();
-
 	static duckdb::Connection *GetConnection();
 	static duckdb::Connection *GetConnectionUnsafe();
 
--- a/include/pgduckdb/pgduckdb_duckdb.hpp
+++ b/include/pgduckdb/pgduckdb_duckdb.hpp
@@ -2,6 +2,12 @@
 
 #include "duckdb.hpp"
 
+// FIXME: Make sure we don't need this
+extern "C" {
+#include "postgres.h"
+#include "access/xact.h"
+}
+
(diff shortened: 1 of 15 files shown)


include/pgduckdb/pgduckdb_duckdb.hpp
```diff
@@ -2,6 +2,12 @@
 
 #include "duckdb.hpp"
 
+// FIXME: Make sure we don't need this
+extern "C" {
+#include "postgres.h"
+#include "access/xact.h"
+}
```
Y--, line 9: Seems we need this for `XactEvent`. Sadly from what I can tell, this is an C enum so we can't forward declare it. It might work forward-declaring it as an `int` but it's a bit dangerous.

Since the callbacks are `static`, can we move them to a dedicated file/compilation unit? Would make splitting easier :-)

Y-- (commented):
Looks reasonable, though it mixes a C++ and PG code a bit much :-) Would be great to untangle it.

PR #414, opened 2024-11-11
Add helper to ensure we're not including PG files

Y--:
Introduce a new header file, which when included in a given source file will generate a compilation header if other PG files have been declared before.

It works by raising a compilation error if any of the macro used in PG to protect against double inclusion of header file is found to be defined already.
To do that we use a script to extract all these macros at compilation time (since it is PG-dependent)


scripts/generate-pg-include-guard.sh
```diff
@@ -0,0 +1,28 @@
...
+
+mkdir -p "${OUTPUT_DIR}"
+
+echo "Will generate: $(pwd)/${CPP_ONLY_FILE}"
+
+echo "// Auto-generated file, run 'make update_cpp_guard' to update" > "${CPP_ONLY_FILE}"
+echo "#if \\" >> "${CPP_ONLY_FILE}"
+
+
+grep -R '_H$' "${PG_INCLUDE_DIR}" | grep define | cut -d: -f 2 | cut -d \  -f 2 | sort | while read -r m;
```
JelteF, line 18: I think this is a overkill. I think failing on `POSTGRES_H` being defined enough for all practical cases that we'll run into.

Y--: Well I went there cause I found a few files that didn't include `postgres.h`, so I was concerned for false positives. But yeah only checking `POSTGRES_H` would certainly be much simpler. Let's chat about it tomorrow.

Y--: Ok, after sleeping on it I think I agree that the small risk is worth the simplicity. I'm cleaned up the PR.

JelteF (approved):

PR #415, opened 2024-11-11
Remove PG code from pgduckdb_tables

Y--:
Convert `pgduckdb_tables` file to a pure C++ file.

### DIFF
--- a/include/pgduckdb/pg/relations.hpp
+++ b/include/pgduckdb/pg/relations.hpp
@@ -1,0 +1,23 @@
+#pragma once
+
+#include "pgduckdb/pg/declarations.hpp"
+
+namespace pgduckdb {
+
+TupleDesc PDRelationGetDescr(Relation relation);
+
+// Not thread-safe. Must be called under a lock.
+Relation OpenRelation(Oid relationId);
+
+// Not thread-safe. Must be called under a lock.
+void CloseRelation(Relation relation);
+
+int GetTupleDescNatts(const TupleDesc tupleDesc);
+
+const char *GetAttName(const Form_pg_attribute);
+
+Form_pg_attribute GetAttr(const TupleDesc tupleDesc, int i);
+
+void EstimateRelSize(Relation rel, int32_t *attr_widths, BlockNumber *pages, double *tuples, double *allvisfrac);
+
+}
--- a/include/pgduckdb/pg/relations.hpp
+++ b/include/pgduckdb/pg/relations.hpp
@@ -4,7 +4,7 @@
 
 namespace pgduckdb {
 
-TupleDesc PDRelationGetDescr(Relation relation);
+TupleDesc RelationGetDescr(Relation relation);
 
 // Not thread-safe. Must be called under a lock.
 Relation OpenRelation(Oid relationId);
--- a/include/pgduckdb/pgduckdb_types.hpp
+++ b/include/pgduckdb/pgduckdb_types.hpp
@@ -6,18 +6,18 @@ namespace pgduckdb {
 
 class PostgresScanGlobalState;
 class PostgresScanLocalState;
-typedef uint64_t idx_t; // From duckdb.h
 
 // DuckDB has date starting from 1/1/1970 while PG starts from 1/1/2000
 constexpr int32_t PGDUCKDB_DUCK_DATE_OFFSET = 10957;
-constexpr int64_t PGDUCKDB_DUCK_TIMESTAMP_OFFSET = static_cast<int64_t>(10957) * static_cast<int64_t>(86400000000) /* USECS_PER_DAY */;
+constexpr int64_t PGDUCKDB_DUCK_TIMESTAMP_OFFSET =
+    static_cast<int64_t>(PGDUCKDB_DUCK_DATE_OFFSET) * static_cast<int64_t>(86400000000) /* USECS_PER_DAY */;
 
 duckdb::LogicalType ConvertPostgresToDuckColumnType(Form_pg_attribute &attribute);
 Oid GetPostgresDuckDBType(duckdb::LogicalType type);
 int32_t GetPostgresDuckDBTypemod(duckdb::LogicalType type);
 duckdb::Value ConvertPostgresParameterToDuckValue(Datum value, Oid postgres_type);
-void ConvertPostgresToDuckValue(Oid attr_type, Datum value, duckdb::Vector &result, idx_t offset);
-bool ConvertDuckToPostgresValue(TupleTableSlot *slot, duckdb::Value &value, idx_t col);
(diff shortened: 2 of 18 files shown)


include/pgduckdb/pg/relations.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "pgduckdb/pg/declarations.hpp"
+
+namespace pgduckdb {
+
+// FIXME - this is a macro in Postgres, would be great to inline it to avoid the overhead of a function call
```
JelteF, line 7: Do we actually care about this? We don't call this for every row right?

Y--: Yeah true, wanted to flag it though. I've removed the comment for now.

include/pgduckdb/pg/relations.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "pgduckdb/pg/declarations.hpp"
+
+namespace pgduckdb {
+
+TupleDesc PDRelationGetDescr(Relation relation);
+
+// Not thread-safe. Must be called under a lock.
```
JelteF, line 9: It would be nice to actually assert that this lock is held in these functions.

src/pg/relations.cpp
```diff
@@ -0,0 +1,63 @@
...
+	ResourceOwner saveResourceOwner = CurrentResourceOwner;
+	CurrentResourceOwner = TopTransactionResourceOwner;
+	PostgresFunctionGuard(relation_close, rel, NoLock);
+
+	CurrentResourceOwner = saveResourceOwner;
+}
+
+void EstimateRelSize(Relation rel, int32_t *attr_widths, BlockNumber *pages, double *tuples, double *allvisfrac) {
+	::estimate_rel_size(rel, attr_widths, pages, tuples, allvisfrac);
+}
```
JelteF, line 61: What's the idea behind this wrapper function? Why not declare the original definition in the header? If this would contain a lock-held assert then it would make sense to me (see other comment).

include/pgduckdb/pg/relations.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "pgduckdb/pg/declarations.hpp"
+
+namespace pgduckdb {
+
+TupleDesc PDRelationGetDescr(Relation relation);
```
JelteF, line 7: Why does only this one need tho PD prefix? does `::RelationGetDescr` not work for macros?

include/pgduckdb/pgduckdb_types.hpp
```diff
@@ -1,25 +1,20 @@
...
 namespace pgduckdb {
 
+class PostgresScanGlobalState;
+class PostgresScanLocalState;
+typedef uint64_t idx_t; // From duckdb.h
+
 // DuckDB has date starting from 1/1/1970 while PG starts from 1/1/2000
 constexpr int32_t PGDUCKDB_DUCK_DATE_OFFSET = 10957;
-constexpr int64_t PGDUCKDB_DUCK_TIMESTAMP_OFFSET = INT64CONST(10957) * USECS_PER_DAY;
+constexpr int64_t PGDUCKDB_DUCK_TIMESTAMP_OFFSET = static_cast<int64_t>(10957) * static_cast<int64_t>(86400000000) /* USECS_PER_DAY */;
```
JelteF, line 13: Seems it would be nice to reuse our constants
```suggestion
constexpr int64_t PGDUCKDB_DUCK_TIMESTAMP_OFFSET = static_cast<int64_t>(PGDUCKDB_DUCK_DATE_OFFSET) * static_cast<int64_t>(86400000000) /* USECS_PER_DAY */;
```

include/pgduckdb/pgduckdb_types.hpp
```diff
@@ -1,25 +1,20 @@
...
-}
-
-#include "pgduckdb/scan/postgres_scan.hpp"
+#include "pgduckdb/pg/declarations.hpp"
 
 namespace pgduckdb {
 
+class PostgresScanGlobalState;
+class PostgresScanLocalState;
+typedef uint64_t idx_t; // From duckdb.h
```
JelteF, line 9: Why do we need to declare these here. Especially the `idx_t` seems strange to declare. Why not include `duckdb.hpp` anymore?

I guess PostgresScanGlobalState and PostgresScanLocalState forward declarations are useful because `include/pgduckdb/scan/postgres_scan.hpp` is including `postgres.h`. Can we change that instead?

src/pg/relations.cpp
Y--: I don't want to bring them in the header because it would bring PG's headers with it. The point of wrapping them is to decouple PG code from places where we use C++.

include/pgduckdb/pg/relations.hpp
Y--: No, macros are basically inline (as text) directly in the source code.

So if you have something like this:
```c++
#define foo(a, b) something_foo(a, b)
```

And you try to declare a function of the same name:
```c++
// In the header
void foo(int a, int b);

// In the source file
#include "my_foo_macro.h"

void foo(int a, int b) {
  ::foo(a, b);
}
```

Then the pre-processor will transform the source file to:
```c++
void foo(int a, int b) {
  something_foo(a, b)(a, b);
}
```

Which sadly doesn't compile.

We should be able to do this:
```C++
void foo(int a, int b) {
  foo
}
```

As long as the parameter match, but that is super hacky and not very readable IMHO

include/pgduckdb/pg/relations.hpp
Y--: This would mean to bring C++ code in this file, which I'm a bit reluctant to do ;-)

include/pgduckdb/pgduckdb_types.hpp
Y--: Because `pgduckdb_types.hpp` is included in a bunch of file I wanted to remain very conservative with adding headers, especially `duckdb.h`. Instead of declaring `idx_t` I've updated the interface to remain consistent, that's even better I think.

> because include/pgduckdb/scan/postgres_scan.hpp is including postgres.h. Can we change that instead?

yep, it's coming :-)

src/pg/relations.cpp
Y--: Ah sorry I misunderstood. Yeah absolutely, though I actually meant to wrap the call with a PG guard since it's not trivial.

include/pgduckdb/pg/relations.hpp
Y--: As discussed offline - will attempt in a separated PR

src/pg/relations.cpp
JelteF: Ah alright, yeah then having the wrapper makes total sense.

JelteF (approved):

Issue #402, opened 2024-11-07
Make `InsertTupleIntoChunk` more efficient


JelteF:
### Description

While renaming variables I think we should be able to simplify the `InsertTupleIntoChunk` code a bit. In at least two ways:
1. the `values` array only needs to contain the values that we actually need to output, now it also contains values that we only need to filter on
2. We now detoast variables twice if they are both part of a filter and one of the output variables. Doing it once should be enough.

JelteF:
Closing this in favor of #477

PR #416, opened 2024-11-11
Make `InsertTupleIntoChunk` more efficient

Reminiscent:
Fix https://github.com/duckdb/pg_duckdb/issues/402

1. Cache the detoast variables if they are both part of a filter and one of the output variables.
2. Only contain the values that we actually need to output in the value.

### DIFF
--- a/include/pgduckdb/scan/postgres_scan.hpp
+++ b/include/pgduckdb/scan/postgres_scan.hpp
@@ -34,6 +34,8 @@ public:
 	std::vector<duckdb::TableFilter *> m_column_filters;
 	/* Duckdb output vector idx with information about postgres column id */
 	duckdb::vector<duckdb::pair<duckdb::idx_t, AttrNumber>> m_output_columns;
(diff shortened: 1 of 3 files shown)


Reminiscent:
I have two questions:
1. Are there any easy way to add a bench test to show the improvement? Or we just add some complex test cases in the regression test and see the result.
2. Now we still use `vector` for the `values` array, we can replace it to the map. But such a change may not be necessary, and it's unclear how much of an improvement it would bring.

JelteF (changes requested):
Sorry for the delay in reviewing this. I left a few comments. Maybe don't spend time on implementing them yet though, because we're currently working on a big change to the reading and filtering. So it's a possibility that we actually end up removing this code.

JelteF:
Okay, so chances are very high that we'll remove this code completely in 0.3.0. So I don't think it makes sense to continue with this approach. Thanks anyway for your efforts.

JelteF:
For reference #477 is the PR that removes this code.

PR #421, opened 2024-11-12
Remove PG code from pgduckdb_transaction

Y--:


src/catalog/pgduckdb_transaction.cpp
```diff
@@ -40,30 +38,19 @@ SchemaItems::GetTable(const duckdb::string &entry_name) {
...
-		return nullptr;
+	Oid rel_oid = GetRelidFromSchemaAndTable(name.c_str(), entry_name.c_str());
+	if (!IsValidOid(rel_oid)) {
+		return nullptr; // Table could not be found
 	}
 
 	// Check if the Relation is a VIEW
-	auto tuple = SearchSysCache1(RELOID, ObjectIdGetDatum(rel_oid));
-	if (!HeapTupleIsValid(tuple)) {
+	auto tuple = SearchSysCacheForRel(rel_oid);
```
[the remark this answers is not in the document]
Y--: Great point, updated. Thanks!

src/pg/relations.cpp
```diff
@@ -62,4 +70,38 @@ void EstimateRelSize(Relation rel, int32_t *attr_widths, BlockNumber *pages, dou
...
+
+HeapTuple
+SearchSysCacheForRel(Oid oid) {
+	return SearchSysCache1(RELOID, ObjectIdGetDatum(oid));
+}
+
+bool
+IsValidHeapTuple(HeapTuple tuple) {
+	return HeapTupleIsValid(tuple);
+}
```
JelteF, line 95: You refactored the code so these functions aren't needed anymore, but you haven't removed them yet. (same for `ReleaseSysCache`.

Y--: Ugh good catch.

src/pg/relations.cpp
```diff
@@ -62,4 +70,23 @@ void EstimateRelSize(Relation rel, int32_t *attr_widths, BlockNumber *pages, dou
 	PostgresFunctionGuard(estimate_rel_size, rel, attr_widths, pages, tuples, allvisfrac);
 }
 
+Oid
+GetRelidFromSchemaAndTable(const char *schema_name, const char *entry_name) {
+	List *name_list = NIL;
+	name_list = lappend(name_list, makeString(pstrdup(schema_name)));
+	name_list = lappend(name_list, makeString(pstrdup(entry_name)));
+	RangeVar *table_range_var = makeRangeVarFromNameList(name_list);
+	return RangeVarGetRelid(table_range_var, AccessShareLock, true);
```
JelteF, line 79: We should probably wrap this whole function in a PostgresFunctionGuard, or maybe introduce a new template/macro to be able to guard a block instead of a function call, or guard a function definition instead of a function call.

Y--: > guard a function definition instead of a function call.

Yeah I have that somewhere, I'll dig it up and update it. In the meantime I've wrapped the whole function call.

JelteF (approved):

PR #422, opened 2024-11-12
Remove PG code from pgduckdb_seq_scan

Y--:


include/pgduckdb/pg/declarations.hpp
```diff
@@ -30,8 +36,13 @@ typedef HeapTupleData *HeapTuple;
 
 struct Node;
 
+typedef uint16_t OffsetNumber;
+
 typedef unsigned int Oid;
 
+typedef char *Pointer;
```
JelteF, line 43: Maybe it makes sense to include `c.h`, which includes a bunch of these definitions.

src/scan/postgres_seq_scan.cpp
```diff
@@ -98,7 +100,7 @@ PostgresSeqScanFunction::PostgresSeqScanFunc(duckdb::ClientContext &context, duc
 
 	auto hasTuple = local_state.m_heap_table_reader->ReadPageTuples(output);
 
-	if (!hasTuple || local_state.m_heap_table_reader->GetCurrentBlockNumber() == InvalidBlockNumber) {
+	if (!hasTuple || IsValidBlockNumber(local_state.m_heap_table_reader->GetCurrentBlockNumber())) {
```
JelteF, line 103: Probably the reason for the failing tests:
```suggestion
	if (!hasTuple || !IsValidBlockNumber(local_state.m_heap_table_reader->GetCurrentBlockNumber())) {
```

Y--: Ah yeah that would help :-) - I'm waiting for #421 to land to take care of this one

include/pgduckdb/pg/declarations.hpp
Y--: You think it's preferable rather than only bringing here what we need?

JelteF (approved):

include/pgduckdb/pg/declarations.hpp
JelteF: To be clear, I think it should be completely safe to include. But it does pollute the global namespace with a bunch of things that we don't need.

PR #426, opened 2024-11-12
Make a few headers pg-code free

Y--:


include/pgduckdb/pgduckdb_planner.hpp
```diff
@@ -2,13 +2,11 @@
...
 
-extern "C" {
-#include "postgres.h"
-#include "optimizer/planner.h"
-}
-
+#include "pgduckdb/pg/declarations.hpp"
 #include "pgduckdb/pgduckdb_duckdb.hpp"
 
+#include "pgduckdb/utility/cpp_only_file.hpp" // Must be last include.
```
JelteF, line 8: Do we really want to include this in headers? That means that you cannot include it after postgres includes in a `cpp` file I guess. Do we really want that?

src/pgduckdb_planner.cpp
```diff
@@ -1,6 +1,8 @@
+#include "pgduckdb/pgduckdb_planner.hpp"
```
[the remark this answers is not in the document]
Y--: Well here `pgduckdb_planner.hpp` is the header of `pgduckdb_planner.cpp` so it make sense to me that it would be first.

include/pgduckdb/pgduckdb_planner.hpp
Y--: I think it is quite helpful to make sure we don't include PG headers in the current header file and indeed because it is a header it must be `#included` before the PG header inclusion in cpp files.

I find it quite useful to iterate for now, but I'm happy to remove it from all headers when our code is in a stable place.

JelteF: Okay makes sense.

PR #431, opened 2024-11-13
Fix MotherDuck table sync

JelteF:
I accidentally broke syncing of MotherDuck table metadata in #413. This fixes that by not using the cached connection in the background worker.


Y-- (commented):
Can we add a test? (even if one we run manually?)

JelteF:
> Can we add a test? (even if one we run manually?)

Automated testing of motherduck support is separate item. To reproduce this issue, it was enough to simply enable motherduck support. The background worker would error continuously.

Y-- (approved):

PR #432, opened 2024-11-13
Refactor decimal code

Y--:
* move all vendor-ed code to a dedicated file
* merge all numeric related code to one single compilation unit - to be split in a following PR


src/pgduckdb_types.cpp
```diff
@@ -1378,4 +1461,92 @@ InsertTupleIntoChunk(duckdb::DataChunk &output, duckdb::shared_ptr<PostgresScanG
 	scan_global_state->m_total_row_count++;
 }
 
+
+Numeric
+CreateNumeric(const NumericVar &var) {
```
JelteF, line 1466: I think having this in its own file (preferably together with some other numeric code) would be nice. Or `pgduckdb_types.cpp` is getting quite big with a lot of random things in it. This function also seems pretty much entirely copied over from PG source code of the `make_result_opt_error` function. Would be nice to add that as a comment on the function.

src/pgduckdb_types.cpp
```diff
@@ -695,113 +816,130 @@ numeric_typmod_scale(int32 typmod) {
 }
 
 duckdb::LogicalType
-ConvertPostgresToDuckColumnType(Form_pg_attribute &attribute) {
```
JelteF, line 698: Overall I like these general refactorings, but they are quite big and seem to have little to do with the actual decimal changes. I would have preferred if this was a separate PR, but I'm okay with keeping here this time.

Y--: ++ forked off there: https://github.com/duckdb/pg_duckdb/pull/473

src/pgduckdb_types.cpp
Y--: Discussed live:
* we'll vendor in `make_result_opt_error`
* we've checked that this function is called only directly from PG, never from a DuckDB thread

JelteF (approved):

PR #433, opened 2024-11-14
Support multi-statement transactions

JelteF:
Only supports multi-statement transactions for non-DDL queries.


include/pgduckdb/pgduckdb_table_am.hpp
```diff
@@ -1,7 +1,4 @@
-extern "C" {
-#include "postgres.h"
-#include "access/tableam.h"
-}
+#include "pgduckdb/pg/declarations.hpp"
```
JelteF, line 1: This was needed because we cannot define the `XactEvent` and `SubXactEvent` twice, and `tableam.h` was including `xact.h` from Postgres. Which would then cause compilation errors if `pgduckdb_table_am.hpp` was included in the same file as `pg/transactions.hpp`

test/regression/sql/transactions.sql
```diff
@@ -0,0 +1,118 @@
...
+
+TRUNCATE t_ddb;
+INSERT INTO t_ddb VALUES (1);
+
+BEGIN;
+DECLARE c SCROLL CURSOR FOR SELECT a FROM t_ddb;
+FETCH NEXT FROM c;
+FETCH NEXT FROM c;
+-- FIXME: Fix the crash that occurs when uncommenting the next line
+-- FETCH PRIOR FROM c;
```
JelteF, line 115: Important fixme, at the very least we should simply error here.

test/regression/expected/transaction_errors.out
```diff
@@ -1,6 +1,8 @@
 CREATE TABLE foo AS SELECT 'bar'::text AS t;
 BEGIN; SET duckdb.force_execution = true; SELECT t::integer AS t1 FROM foo; ROLLBACK;
-ERROR:  invalid input syntax for type integer: "bar"
+ERROR:  (PGDuckDB/Duckdb_ExecCustomScan) Conversion Error: Could not convert string 'bar' to INT32
```
JelteF, line 3: Previously this was being executed in Postgres, now it's executed in duckdb again. We still don't crash though.

test/regression/expected/duckdb_recycle.out
```diff
@@ -54,4 +49,24 @@ EXPLAIN SELECT count(*) FROM ta;
...
+ERROR:  duckdb.recycle_ddb() cannot run inside a transaction block
+END;
+-- Nor in a function
+CREATE OR REPLACE FUNCTION f() RETURNS void
+    LANGUAGE plpgsql
+    RETURNS NULL ON NULL INPUT
+    AS
+$$
+BEGIN
+    CALL duckdb.recycle_ddb();
```
JelteF, line 64: I changed this to use the `CALL` syntax, so we can easily detect when `recycle_ddb` is not run at the top level (like shown here).

test/pycheck/prepared_test.py
```diff
@@ -125,8 +125,8 @@ def test_prepared_pipeline(conn: Connection):
         # the duckdb table that should fail because the insert into the heap
         # table opens an implicit transaction.
         with pytest.raises(
-            psycopg.errors.ActiveSqlTransaction,
-            match="DuckDB queries cannot be executed within a pipeline",
+            psycopg.errors.InternalError,
```
JelteF, line 128: It would be nice if we would do some errorcode conversions between DuckDB and Postgres errors. That seems like a task for a separate PR though.

src/pgduckdb_duckdb.cpp
```diff
@@ -156,6 +178,18 @@ DuckDBManager::Initialize() {
 
 	LoadFunctions(context);
 	LoadExtensions(context);
+
+	RegisterDuckdbXactCallback();
```
JelteF, line 182: TODO: This should be moved to PG_init (or backend start). Now that we use this to reset the `top_level_statement` variable it should always be enabled and never unregistered.

src/pgduckdb_ddl.cpp
```diff
@@ -101,6 +96,66 @@ DuckdbHandleDDL(Node *parsetree) {
 	}
 }
 
+static void
+DuckdbUtilityHook_Cpp(PlannedStmt *pstmt, const char *query_string, bool read_only_tree, ProcessUtilityContext context,
+                      ParamListInfo params, struct QueryEnvironment *query_env, DestReceiver *dest,
+                      QueryCompletion *qc) {
+	Node *parsetree = pstmt->utilityStmt;
+	if (pgduckdb::IsExtensionRegistered() && IsA(parsetree, CopyStmt)) {
```
Y--, line 104: Should we just test for `pgduckdb::IsExtensionRegistered()` at the top of the function once and for all (and return early after calling prev hook)? Is there any case where we might want to do something when this is false?

src/pgduckdb_duckdb.cpp
```diff
@@ -301,28 +323,30 @@ DuckDBManager::CreateConnection() {
...
 	}
 
 	auto &instance = Get();
 	auto &context = *instance.connection->context;
 
-	if (!transaction_handler_configured) {
-		RegisterDuckdbXactCallback();
+	if (!context.transaction.HasActiveTransaction()) {
+		if (IsSubTransaction()) {
+			elog(ERROR, "SAVEPOINT is not supported in DuckDB");
```
Y--, line 338: We probably want to throw in this file no? (I see that L330 we're already `elog(ERROR` though, so maybe something to change later)

Also regarding the error message, is it what PG user would expect? More than something less technical like "Nested transactions are not supported in DuckDB"?

src/pgduckdb_duckdb.cpp
```diff
@@ -301,28 +323,30 @@ DuckDBManager::CreateConnection() {
...
+		}
+		if (IsInTransactionBlock() || force_transaction) {
```
Y--, line 340: nit - 
```suggestion
		if (force_transaction || IsInTransactionBlock()) {
```

src/pgduckdb_duckdb.cpp
```diff
@@ -301,28 +323,30 @@ DuckDBManager::CreateConnection() {
...
+			/*
+			 * We only want to open a new DuckDB transaction if we're already
+			 * in a Postgres transaction block. Always opening a transaction
+			 * incurs a performance penalty when connecting to
```
Y--, line 344:
```suggestion
			 * incurs a performance penalty when connecting
```
or
```suggestion
			 * incurs a performance penalty when connecting to DuckDB
```

src/pgduckdb_hooks.cpp
```diff
@@ -207,8 +203,21 @@ DuckdbPlannerHook_Cpp(Query *parse, const char *query_string, int cursor_options
 			}
 			/* If we can't create a plan, we'll fall back to Postgres */
 		}
+		if (parse->commandType != CMD_SELECT && pgduckdb::DuckdbDidWrites() && pgduckdb::IsInTransactionBlock(true)) {
```
Y--, line 206: Reading statements like this, I feel it might be useful to either nest PG APIs in a sub-namespace `pg` or to include PG in their name like `PGIsInTransactionBlock`.
Since we're bridging two databases, it is helpful to be extra-pedentic when dealing with their APIs.

What do you think?

src/pgduckdb_planner.cpp
```diff
@@ -120,6 +120,14 @@ DuckdbPlanNode(Query *parse, const char *query_string, int cursor_options, Param
 		return nullptr;
 	}
 
+	/*
+	 * If creating a plan for a scrollable cursor, make sure it can run
+	 * backwards on demand.  Add a Material node at the top at need.
```
Y--, line 125:
```suggestion
	 * backwards on demand. Add a Material node at the top as needed.
```

src/pgduckdb_xact.cpp
```diff
@@ -1,39 +1,172 @@
...
+static int64_t duckdb_command_id = -1;
+static bool top_level_statement = true;
+
+/*
+ * Returns if we're currently in a transaction block. To determine if we are in
+ * a function or not, this uses the tracked top_level_statement variable.
+ */
+bool
+IsInTransactionBlock() {
+	return IsInTransactionBlock(top_level_statement);
```
Y--, line 17: I assume this is safe to be called without a lock on `top_level_statement` because this will only be called in PG hooks, not in DDB threads?
(same for `duckdb_command_id` and potential `elog(ERROR` below?)

Y-- (commented):
LGTM, tests show that it is doing what we expect. Added a few comments :-)

src/pgduckdb_xact.cpp
JelteF: I introduced a `pgduckdb::pg` and `pgduckdb::ddb` namespace now. I moved this to the `pgduckdb::pg` namespace. I think we should write down assumptions about using functions from the `pgduckdb::pg` namespace. The two main assumptions are imo:

1. They should be called from a Postgres hook, or while holding the global PG lock from a duckdb thread.
2. They do not throw elog errors, only exceptions

src/pgduckdb_duckdb.cpp
JelteF: added a lot more context here, I think I gave up writing this comment halfway through.

src/pgduckdb_planner.cpp
JelteF: This comment came straight from the postgres source code. Reworded it, because it actually didn't reflect some of the changes I made to the code.

src/pgduckdb_hooks.cpp
JelteF: Agreed, I was thinking the same. Done now, let me know what you think. We should probably update the other code in another PR for consistency.

Y--: This looks really good, much clearer. And agreed to update the rest of the repo to follow the same philosophy.

src/pgduckdb_xact.cpp
Y--: Yeah I think that makes total sense.

Y-- (approved):
LGTM! 🚀

JelteF:
I merged this a bit quicker than I originally intended, mainly to fix the broken build on the main branch due to this PG change: https://www.postgresql.org/message-id/flat/CAGECzQQS1Oun40BssQDmuHc86teXqr%2BLOndy5NSVdFce9evZvQ%40mail.gmail.com#cf1398dd28d67be9fce7e75096343f07

PR #434, opened 2024-11-14
Cached file management

mkaruza:
For each cached file we will now create additional metadata file that contains simple metadata. Added funcitons `duckdb.cache_info()` and `duckdb.cache_delete(cache_key TEXT)` that are used for managing cached files.

### DIFF
--- a/sql/pg_duckdb--0.1.0--0.2.0.sql
+++ b/sql/pg_duckdb--0.1.0--0.2.0.sql
@@ -49,9 +49,10 @@ END;
 $func$;
 
 CREATE TYPE duckdb.cache_info AS (
-  cache_key TEXT,
   remote_path TEXT,
-  file_size TEXT
+  cache_key TEXT,
+  cache_file_size BIGINT,
+  cache_file_timestamp TIMESTAMPTZ
 );
 
 CREATE FUNCTION duckdb.cache_info()
--- a/sql/pg_duckdb--0.1.0--0.2.0.sql
+++ b/sql/pg_duckdb--0.1.0--0.2.0.sql
@@ -47,3 +47,21 @@ BEGIN
     RAISE EXCEPTION 'Function `read_json(TEXT[])` only works with Duckdb execution.';
 END;
 $func$;
+
+CREATE TYPE duckdb.cache_info AS (
+  cache_key TEXT,
+  remote_path TEXT,
+  file_size TEXT
+);
+
+CREATE FUNCTION duckdb.cache_info()
+    RETURNS SETOF duckdb.cache_info
+    SET search_path = pg_catalog, pg_temp
+    LANGUAGE C AS 'MODULE_PATHNAME', 'cache_info';
+REVOKE ALL ON FUNCTION duckdb.cache_info() FROM PUBLIC;
+
+CREATE FUNCTION duckdb.cache_delete(cache_key TEXT)
+    RETURNS bool
+    SET search_path = pg_catalog, pg_temp
+    LANGUAGE C AS 'MODULE_PATHNAME', 'cache_delete';
+REVOKE ALL ON FUNCTION duckdb.cache_delete(cache_key TEXT) FROM PUBLIC;
--- a/src/pgduckdb_options.cpp
+++ b/src/pgduckdb_options.cpp
@@ -181,17 +181,17 @@ DuckdbCacheObject(Datum object, Datum type) {
 	return true;
 }
 
-typedef struct CacheInfo {
+typedef struct CacheFileInfo {
 	std::string cache_key;
 	std::string remote_path;
 	int64_t file_size;
 	Timestamp file_timestamp;
-} CacheInfo;
+} CacheFileInfo;
 
-static std::vector<CacheInfo>
-DuckdbGetCachedFile() {
+static std::vector<CacheFileInfo>
+DuckdbGetCachedFilesInfos() {
 	std::string ext(".meta");
-	std::vector<CacheInfo> cache_info;
+	std::vector<CacheFileInfo> cache_info;
 	for (auto &p : std::filesystem::recursive_directory_iterator(CreateOrGetDirectoryPath("duckdb_cache"))) {
 		if (p.path().extension() == ext) {
 			std::ifstream cache_file_metadata(p.path());
@@ -199,9 +199,14 @@ DuckdbGetCachedFile() {
 			std::vector<std::string> metadata_tokens;
 			while (std::getline(cache_file_metadata, metadata, ',')) {
 				metadata_tokens.push_back(metadata);
-			};
-			cache_info.push_back(CacheInfo {metadata_tokens[0], metadata_tokens[1], std::stoi(metadata_tokens[2]),
-			                                std::stoi(metadata_tokens[3])});
+			}
+			if (metadata_tokens.size() != 4) {
+				elog(WARNING, "(PGDuckDB/DuckdbGetCachedFilesInfos) Invalid '%s' cache metadata file",
+				     p.path().c_str());
+					 break;
+			}
+			cache_info.push_back(CacheFileInfo {metadata_tokens[0], metadata_tokens[1], std::stoi(metadata_tokens[2]),
+			                                    std::stoi(metadata_tokens[3])});
 		}
 	}
 	return cache_info;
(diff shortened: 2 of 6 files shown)


wuputah (commented):
Would be great to add tests to this.

The code looks pretty straightforward but if @JelteF or @Y-- could have a quick look for particulars it would be appreciated.

src/pgduckdb_options.cpp
```diff
@@ -175,6 +180,39 @@ DuckdbCacheObject(Datum object, Datum type) {
...
+	std::string ext(".meta");
+	std::vector<CacheInfo> cache_info;
+	for (auto &p : std::filesystem::recursive_directory_iterator(CreateOrGetDirectoryPath("duckdb_cache"))) {
+		if (p.path().extension() == ext) {
+			std::ifstream cache_file_metadata(p.path());
+			std::string metadata;
+			std::vector<std::string> metadata_tokens;
+			while (std::getline(cache_file_metadata, metadata, ',')) {
+				metadata_tokens.push_back(metadata);
+			};
```
Y--, line 200:
```suggestion
			}
```

src/pgduckdb_options.cpp
```diff
@@ -175,6 +180,39 @@ DuckdbCacheObject(Datum object, Datum type) {
...
+			cache_info.push_back(CacheInfo {metadata_tokens[0], metadata_tokens[1], metadata_tokens[2]});
```
Y--, line 201: We should check that `metadata_tokens.size() == 3` or throw if not

src/pgduckdb_options.cpp
```diff
@@ -175,6 +180,39 @@ DuckdbCacheObject(Datum object, Datum type) {
...
+DuckdbGetCachedFile() {
+	std::string ext(".meta");
+	std::vector<CacheInfo> cache_info;
+	for (auto &p : std::filesystem::recursive_directory_iterator(CreateOrGetDirectoryPath("duckdb_cache"))) {
+		if (p.path().extension() == ext) {
+			std::ifstream cache_file_metadata(p.path());
+			std::string metadata;
+			std::vector<std::string> metadata_tokens;
+			while (std::getline(cache_file_metadata, metadata, ',')) {
+				metadata_tokens.push_back(metadata);
```
Y--, line 199: Should we throw if `metadata_tokens.size() > 3` to avoid reading a potentially invalid file?

src/pgduckdb_options.cpp
```diff
@@ -175,6 +180,39 @@ DuckdbCacheObject(Datum object, Datum type) {
...
+			};
+			cache_info.push_back(CacheInfo {metadata_tokens[0], metadata_tokens[1], metadata_tokens[2]});
+		}
+	}
+	return cache_info;
+}
+
+static bool
+DuckdbCacheDelete(Datum cache_key_datum) {
+	auto cache_key = DatumToString(cache_key_datum);
```
Y--, line 209: We probably want to throw if `cache_key` is empty?

src/pgduckdb_options.cpp
```diff
@@ -199,6 +237,69 @@ DECLARE_PG_FUNCTION(cache) {
...
+	MemoryContext oldcontext = MemoryContextSwitchTo(rsinfo->econtext->ecxt_per_query_memory);
+	attinmeta = TupleDescGetAttInMetadata(cache_info_tuple_desc);
+	tuple_store = tuplestore_begin_heap(rsinfo->allowedModes & SFRM_Materialize_Random, false, work_mem);
+	MemoryContextSwitchTo(oldcontext);
+
+	char **values = (char **)palloc0((3) * sizeof(char *));
+
+	for (int i = 0; i < result.size(); i++) {
+
+		auto &cache_info = result[i];
```
Y--, line 263: nit
```suggestion
	for (auto &cache_info: result) {
```

src/pgduckdb_options.cpp
```diff
@@ -175,6 +180,39 @@ DuckdbCacheObject(Datum object, Datum type) {
...
 }
 
+typedef struct CacheInfo {
+	std::string cache_key;
+	std::string remote_path;
+	std::string file_size;
+} CacheInfo;
+
+static std::vector<CacheInfo>
+DuckdbGetCachedFile() {
```
Y--, line 190:
```suggestion
DuckdbGetCachedFilesInfos() {
```

src/pgduckdb_options.cpp
```diff
@@ -175,6 +180,39 @@ DuckdbCacheObject(Datum object, Datum type) {
 	return true;
 }
 
+typedef struct CacheInfo {
```
Y--, line 183:
```suggestion
typedef struct CacheFileInfo {
```

src/pgduckdb_options.cpp
```diff
@@ -199,6 +237,69 @@ DECLARE_PG_FUNCTION(cache) {
 	PG_RETURN_BOOL(result);
 }
 
+DECLARE_PG_FUNCTION(cache_info) {
+	ReturnSetInfo *rsinfo = (ReturnSetInfo *)fcinfo->resultinfo;
+	Tuplestorestate *tuple_store;
+	TupleDesc cache_info_tuple_desc;
+	AttInMetadata *attinmeta;
+
+	auto result = pgduckdb::DuckdbGetCachedFile();
```
Y--, line 246:
```suggestion
	auto files_infos = pgduckdb:: DuckdbGetCachedFilesInfos();
```

src/pgduckdb_options.cpp
```diff
@@ -199,6 +237,69 @@ DECLARE_PG_FUNCTION(cache) {
...
+		values[2] = file_size;
+
+		HeapTuple tuple;
+		tuple = BuildTupleFromCStrings(attinmeta, values);
+		tuplestore_puttuple(tuple_store, tuple);
+		heap_freetuple(tuple);
+
+		pfree(file_size);
+		pfree(remote_path);
+		pfree(cache_key);
```
Y--, line 287: It's too bad that PG doesn't allow to create tuple with no copy/realloc :(

third_party/cached_httpfs/http_file_cache.cpp
```diff
@@ -48,6 +48,17 @@ CachedFileHandle::CachedFileHandle(shared_ptr<CachedFile> &file_p) {
 	file = file_p;
 }
 
+void CachedFileHandle::WriteMetadata(const string &cache_key, const string &remote_path, idx_t total_size) {
+	D_ASSERT(!file->initialized);
+	string metadata_file_name = file->cache_directory + "/" + cache_key + ".meta";
+	FileOpenFlags flags = FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE | FileLockType::WRITE_LOCK;
+	auto handle = file->fs.OpenFile(metadata_file_name, flags);
+	string metadata_info = cache_key + "," + remote_path + "," + std::to_string(total_size);
+	handle->Write((void *)metadata_info.c_str(), metadata_info.length(), 0);
```
Y--, line 57: Doesn't `Write` take a `const char*`?
```suggestion
	handle->Write(cache_key.c_str(), cache_key.length(), 0);
	handle->Write(",", 1, 0);
	handle->Write(remote_path.c_str(), remote_path.length(), 0);
	handle->Write(",", 1, 0);
	auto s = std::to_string(total_size);
	handle->Write(s.c_str(), s.length(), 0);
```

I'm also surprised we don't need a new line after?

mkaruza: Write complains unless it is void * so we need to cast. Also, writing commas will also fail.
IMO, constructing single metadata line is simpler.

src/pgduckdb_options.cpp
mkaruza: Changes so we don't need to pmalloc/pfree for intermediate results.

src/pgduckdb_options.cpp
mkaruza: We want to report back to user which metadata are invalid and print back ones that are good.

sql/pg_duckdb--0.1.0--0.2.0.sql
```diff
@@ -47,3 +47,21 @@ BEGIN
     RAISE EXCEPTION 'Function `read_json(TEXT[])` only works with Duckdb execution.';
 END;
 $func$;
+
+CREATE TYPE duckdb.cache_info AS (
+  cache_key TEXT,
+  remote_path TEXT,
+  file_size TEXT
```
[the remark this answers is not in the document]
mkaruza: Added timestamp cache creation time.

third_party/cached_httpfs/http_file_cache.cpp
Y--: Ok, fair enough

Y-- (approved):
LGTM, thanks for the extra iteration!

Issue #435, opened 2024-11-14
Server crashes when execute query "WITH RECURSIVE"


saygoodbyye:
### What happens?

Server crashes if execute script below
PostgreSQL build:
```
CFLAGS='-Og' ./configure --enable-tap-tests --with-openssl --enable-debug --enable-cassert --with-icu --with-lz4 --with-libxml
``` 

### To Reproduce

test.sql:
```sql
CREATE EXTENSION pg_duckdb;
WITH RECURSIVE outermost(x) AS (
 SELECT 1
 UNION (WITH innermost1 AS (
  SELECT 2
  UNION (WITH innermost2 AS (
   SELECT 3
   UNION (WITH innermost3 AS (
    SELECT 4
    UNION (WITH innermost4 AS (
     SELECT 5
     UNION (WITH innermost5 AS (
      SELECT 6
      UNION (WITH innermost6 AS
       (SELECT 7)
       SELECT * FROM innermost6))
      SELECT * FROM innermost5))
     SELECT * FROM innermost4))
    SELECT * FROM innermost3))
   SELECT * FROM innermost2))
  SELECT * FROM outermost
  UNION SELECT * FROM innermost1)
 )
 SELECT * FROM outermost ORDER BY 1;
```
backtrace:
```c
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007e6fd904526e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007e6fd90288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x000062c065354bda in ExceptionalCondition (conditionName=conditionName@entry=0x7e6fd97e8818 "subquery->setOperations == NULL", fileName=fileName@entry=0x7e6fd97e9edd "src/vendor/pg_ruleutils_17.c", lineNumber=lineNumber@entry=6178) at assert.c:66
[...]

PR #436, opened 2024-11-14
Handle "WITH RECURSIVE" abnormal crashes

Leo-XM-Zeng:
This seems to be a PG bug. When a CTE expression with multiple unions is run, this line assertion will necessarily be triggered, causing the program to terminate.

### DIFF
--- a/test/regression/sql/cte.sql
+++ b/test/regression/sql/cte.sql
@@ -5,8 +5,6 @@ WITH modifying_cte AS (
     INSERT INTO t VALUES (1) RETURNING *
 ) select * from modifying_cte;
 
-DROP TABLE t;
-
 WITH RECURSIVE outermost(x) AS (
  SELECT 1
  UNION (WITH innermost1 AS (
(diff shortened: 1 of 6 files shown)


Leo-XM-Zeng:
#436

Leo-XM-Zeng:
This seems to be a design flaw in pg_get_querydef, which does not adequately account for these scenarios

> This seems to be a PG bug. When a CTE expression with multiple unions is run, this line assertion will necessarily be triggered, causing the program to terminate.

Leo-XM-Zeng:
#435

test/regression/sql/cte.sql
```diff
@@ -6,3 +6,27 @@ WITH modifying_cte AS (
 ) select * from modifying_cte;
 
 DROP TABLE t;
+
+WITH RECURSIVE outermost(x) AS (
```
JelteF, line 10: nit: can you move this query before the `DROP TABLE` command. I think it's nice for consistency between tests to keep all the table drops at the end of the sql file.

JelteF (approved):
LGTM, nice that you also created a patch for the upstream issue in Postgres: https://commitfest.postgresql.org/51/5383/

test/regression/sql/cte.sql
Leo-XM-Zeng: Ok, that's done

JelteF (approved):

PR #439, opened 2024-11-15
Fail compilation on warnings

Y--:


src/pgduckdb_metadata_cache.cpp
```diff
@@ -75,7 +75,7 @@ uint32 schema_hash_value;
  * IsExtensionRegistered for details).
  */
 static void
-InvalidateCaches(Datum arg, int cache_id, uint32 hash_value) {
+InvalidateCaches(Datum, int, uint32 hash_value) {
```
JelteF, line 78: I don't think I like this type of change. If we explicitly don't use arguments provided by a hook, I'd like to know what those arguments are used for. Both for reviewing code, but also for when wanting to change behavior of a hook in the future.

src/pgduckdb_ddl.cpp
```diff
@@ -51,7 +51,7 @@ DuckdbTruncateTable(Oid relation_oid) {
 }
 
 void
-DuckdbHandleDDL(Node *parsetree, const char *queryString) {
+DuckdbHandleDDL(Node *parsetree, const char *) {
```
JelteF, line 54: I think here we can just remove this argument, since we control the callsite.

src/pgduckdb_options.cpp
```diff
@@ -200,6 +200,7 @@ DECLARE_PG_FUNCTION(cache) {
 }
 
 DECLARE_PG_FUNCTION(pgduckdb_recycle_ddb) {
+	(void)fcinfo;
```
JelteF, line 203: This seems a bit silly.

src/pgduckdb_ddl.cpp
Y--: Great catch - and that's a good example of why I think this warning is useful :-)

src/pgduckdb_options.cpp
Y--: Indeed - I've reworked it.

src/catalog/pgduckdb_schema.cpp
```diff
@@ -7,86 +7,83 @@
...
+PostgresSchema::CreateType(duckdb::CatalogTransaction, duckdb::CreateTypeInfo &) {
 	throw duckdb::NotImplementedException("CreateType not supported yet");
 }
 
 duckdb::optional_ptr<duckdb::CatalogEntry>
-PostgresSchema::GetEntry(duckdb::CatalogTransaction transaction, duckdb::CatalogType type,
-                         const duckdb::string &entry_name) {
-	auto &pg_transaction = transaction.transaction->Cast<PostgresTransaction>();
-	return pg_transaction.GetCatalogEntry(type, name, entry_name);
+PostgresSchema::GetEntry(duckdb::CatalogTransaction _catalog_transaction, duckdb::CatalogType _type,
```
JelteF, line 74: Why are these underscores necessary? So they don't collide with class members? Seems unnecessary for the
`_catalog_transaction` one.

Y--: > Why are these underscores necessary? So they don't collide with class members?

Yes exactly, while technically C++ let you shadow class members with function parameters, it can be source of bugs to allow it.

And even though it was not necessary for `_catalog_transaction` I preferred to be consistent for all parameters.

src/pgduckdb_detoast.cpp
```diff
@@ -110,7 +110,10 @@ ToastFetchDatum(struct varlena *attr) {
 
 	struct varlena *result = (struct varlena *)duckdb_malloc(attrsize + VARHDRSZ);
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-compare"
```
JelteF, line 114: Would be nice if this had add a comment why it's necessary?

src/pgduckdb_background_worker.cpp
```diff
@@ -58,7 +58,7 @@ static uint64 initial_cache_version = 0;
 
 extern "C" {
 PGDLLEXPORT void
-pgduckdb_background_worker_main(Datum main_arg) {
+pgduckdb_background_worker_main(Datum) {
```
JelteF, line 61:
```suggestion
pgduckdb_background_worker_main(Datum /*main_arg*/) {
```

Makefile
```diff
@@ -35,8 +35,8 @@ endif
 DUCKDB_LIB = libduckdb$(DLSUFFIX)
 FULL_DUCKDB_LIB = third_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src/$(DUCKDB_LIB)
 
-override PG_CPPFLAGS += -Iinclude -Ithird_party/duckdb/src/include -Ithird_party/duckdb/third_party/re2
-override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register ${DUCKDB_BUILD_CXX_FLAGS}
+override PG_CPPFLAGS += -Iinclude -isystem third_party/duckdb/src/include -isystem third_party/duckdb/third_party/re2
+override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register -Wshadow -Wswitch -Wunused-parameter -Wunreachable-code -Wno-unknown-pragmas ${DUCKDB_BUILD_CXX_FLAGS} -Wall -Wextra -Werror
```
JelteF, line 39:
```suggestion
override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register -Wshadow -Wswitch -Wunused-parameter -Wunreachable-code -Wno-unknown-pragmas ${DUCKDB_BUILD_CXX_FLAGS} -Wall -Wextra
```

`-Werror` should be opt-in (and turned on in CI). It's way to common for different compilers to have some extra warnings, and there's no need to fail compilation for someone building the pg_duckdb on a different system than CI is using.

Makefile
```diff
@@ -35,8 +35,8 @@ endif
 DUCKDB_LIB = libduckdb$(DLSUFFIX)
 FULL_DUCKDB_LIB = third_party/duckdb/build/$(DUCKDB_BUILD_TYPE)/src/$(DUCKDB_LIB)
 
-override PG_CPPFLAGS += -Iinclude -Ithird_party/duckdb/src/include -Ithird_party/duckdb/third_party/re2
-override PG_CXXFLAGS += -std=c++17 -Wno-sign-compare -Wno-register ${DUCKDB_BUILD_CXX_FLAGS}
+override PG_CPPFLAGS += -Iinclude -isystem third_party/duckdb/src/include -isystem third_party/duckdb/third_party/re2
```
JelteF, line 38: I think we should probably turn on some warnings here too.

src/pgduckdb_node.cpp
```diff
@@ -71,7 +71,7 @@ Duckdb_CreateCustomScanState(CustomScan *cscan) {
 }
 
 void
-Duckdb_BeginCustomScan_Cpp(CustomScanState *cscanstate, EState *estate, int eflags) {
+Duckdb_BeginCustomScan_Cpp(CustomScanState *cscanstate, EState *estate, int) {
```
JelteF, line 74:
```suggestion
Duckdb_BeginCustomScan_Cpp(CustomScanState *cscanstate, EState *estate, int /*eflags*/) {
```

src/pgduckdb_table_am.cpp
```diff
@@ -82,13 +82,12 @@ duckdb_scan_end(TableScanDesc sscan) {
 }
 
 static void
-duckdb_scan_rescan(TableScanDesc sscan, ScanKey key, bool set_params, bool allow_strat, bool allow_sync,
-                   bool allow_pagemode) {
+duckdb_scan_rescan(TableScanDesc, ScanKey, bool, bool, bool, bool) {
```
JelteF, line 85: Pretty much all of the arguments in this file should be commented out, not fully removed. Maybe even ignore the warning for this specific file.

src/scan/postgres_seq_scan.cpp
```diff
@@ -68,25 +68,23 @@ PostgresSeqScanFunction::PostgresSeqScanFunction()
 }
 
 duckdb::unique_ptr<duckdb::GlobalTableFunctionState>
-PostgresSeqScanFunction::PostgresSeqScanInitGlobal(duckdb::ClientContext &context,
-                                                   duckdb::TableFunctionInitInput &input) {
+PostgresSeqScanFunction::PostgresSeqScanInitGlobal(duckdb::ClientContext &, duckdb::TableFunctionInitInput &input) {
 	auto &bind_data = input.bind_data->CastNoConst<PostgresSeqScanFunctionData>();
 	auto global_state = duckdb::make_uniq<PostgresSeqScanGlobalState>(bind_data.m_rel, input);
 	global_state->m_global_state->m_snapshot = bind_data.m_snapshot;
-	return std::move(global_state);
```
JelteF, line 76: Is this one not needed anymore?

JelteF (changes requested):

Makefile
JelteF, line 38: I'm curious about why this `-isystem` flag is now needed

src/scan/postgres_seq_scan.cpp
Y--: It was redundant (prevents some compiler optimizations then), and caught by a compiler warning.

JelteF (approved):

PR #447, opened 2024-11-22
Keep our "get_setop_query" consistent with PG

Leo-XM-Zeng:
After discussion on the pgsql-hackers mailinglist followed by #436, tom lane proposed an additional change:
https://github.com/postgres/postgres/commit/94131cd53c1552efe85c3e8120356c22cd658163#diff-6c12827097e95912476267d73f3f56e8250b13aad37ab4286c3c7fdab7f30ec7R6367

Let's incorporate that change in our ruleutils files.


Leo-XM-Zeng:
#436

PR #456, opened 2024-11-27
Let centos 8 compile through

Leo-XM-Zeng:
we need a std::move around that return (for it to compile on centos 8).
https://github.com/duckdb/pg_duckdb/discussions/455


JelteF:
I guess we need to disable this warning there: `-Werror=redundant-move`

Y--:
Thanks for the PR @Leo-XM-Zeng - I've added an override to disable warning for now.
Eventually I'd like to have that be compiler-dependent.

Y-- (approved):

JelteF (approved):

PR #466, opened 2024-11-30
Handles exceptions when using the "explain verbose" statement

### LINKED ISSUE #465: Stack depth limit exceeded when using explain verbose
### What happens?

Stack depth limit exceeded when using explain verbose
0245ee83d25c14791802b4c096b4c38

### To Reproduce

```plpgsql
SET duckdb.force_execution = true;
CREATE TABLE ta(a INT);
EXPLAIN VERBOSE SELECT * FROM ta;
``` 

### OS:

centos8

main

16.6

### Hardware:

_No response_

### Full Name:

Man Zeng

### Affiliation:

Halo

Yes

- [X] Yes, I have

- [X] Yes, I have

Leo-XM-Zeng:
Fixes #465


JelteF (commented):
This is not how this should be fixed. Basically this now disables `EXPLAIN VERBOSE` flag completely to hide the bug. That might be acceptable for a first iteration, except for the fact that this disables EXPLAIN VERBOSE even for queries that don't use DuckDB. I opened #468 now to fix this issue in a better way. I also change the tests to be part of our existing explain tests.

Leo-XM-Zeng:
> This is not how this should be fixed. Basically this now disables `EXPLAIN VERBOSE` flag completely to hide the bug. That might be acceptable for a first iteration, except for the fact that this disables EXPLAIN VERBOSE even for queries that don't use DuckDB. I opened #468 now to fix this issue in a better way. I also change the tests to be part of our existing explain tests.

oh, I looked at your approach and it was more appropriate, I just took into account that duckdb does not support "explain verbose" and that all the operations on "explain" are performed in duckdb, not PG.

JelteF:
To be clear: thanks a lot for the bug report and the initial attempt at a fix.

Leo-XM-Zeng:
> To be clear: thanks a lot for the bug report and the initial attempt at a fix.

Thank you for your encouragement and guidance. haha~

PR #468, opened 2024-12-02
Correctly populate targetlist of CustomScan

### LINKED ISSUE #465: Stack depth limit exceeded when using explain verbose

JelteF:
`EXPLAIN VERBOSE` was erroring out when the plan contained one of our `CustomScan` nodes. The reason for this was that we were not populating the `targetlist` and `custom_scan_tlist` fields in a way that PG understood so it was recursing infinitely into the same function. This changes how we populate those fields to match what PG expects.

Sadly the additional info in the plan is currently not super useful, and arguably wrong. It will always show the following line in the explain plan.
```
   Output: duckdb_scan.explain_key, duckdb_scan.explain_value
```

To fix that we shouldn't include the EXPLAIN part when sending it to DuckDB for initial planning. Sadly that then breaks the code that we use for executing the query. So for now we're happy with not throwing an error, and that we're populating `targetlist` correctly (which might have some more benefits that we didn't know of).

Fixes #465
Fixes #466
Fixes #463

### DIFF
--- a/src/pgduckdb_planner.cpp
+++ b/src/pgduckdb_planner.cpp
@@ -90,7 +90,7 @@ CreatePlan(Query *query, bool throw_error) {
 		/* We fill in the varno later, once we know the index of the custom RTE
 		 * that we create. We'll know this at the end of DuckdbPlanNode. This
 		 * can probably be simplified when we don't call the standard_planner
-		 * anymore innside DuckdbPlanNodei, because then we only need a single
+		 * anymore inside DuckdbPlanNode, because then we only need a single
 		 * RTE. */
 		Var *var = makeVar(0, i + 1, postgresColumnOid, typtup->typtypmod, typtup->typcollation, 0);
 
--- a/src/pgduckdb_planner.cpp
+++ b/src/pgduckdb_planner.cpp
@@ -87,13 +87,22 @@ CreatePlan(Query *query, bool throw_error) {
 
 		typtup = (Form_pg_type)GETSTRUCT(tp);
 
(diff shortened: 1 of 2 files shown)


src/pgduckdb_planner.cpp
```diff
@@ -87,14 +87,23 @@ CreatePlan(Query *query, bool throw_error) {
 
 		typtup = (Form_pg_type)GETSTRUCT(tp);
 
-		Var *var = makeVar(INDEX_VAR, i + 1, postgresColumnOid, typtup->typtypmod, typtup->typcollation, 0);
+		/* We fill in the varno later, once we know the index of the custom RTE
+		 * that we create. We'll know this at the end of DuckdbPlanNode. This
+		 * can probably be simplified when we don't call the standard_planner
+		 * anymore innside DuckdbPlanNodei, because then we only need a single
```
Y--, line 93: vim user detected :-)
```suggestion
		 * anymore innside DuckdbPlanNode, because then we only need a single
```

Y-- (approved):
LGTM!

PR #470, opened 2024-12-03
Allow user to configure default MotherDuck database in postgresql.conf

naoyak:
Fixes #459

Allow the user to configure a different default MotherDuck database from `my_db` in `postgresql.conf`. This affords flexibility because there is currently no user-facing API to change the internal "default" DB in MotherDuck.

### DIFF
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -15,6 +15,7 @@
 extern "C" {
 #include "postgres.h"
 #include "catalog/namespace.h"
+#include "lib/stringinfo.h"
 #include "utils/lsyscache.h"  // get_relname_relid
 #include "utils/fmgrprotos.h" // pg_sequence_last_value
 #include "common/file_perm.h"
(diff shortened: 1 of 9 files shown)


src/pgduckdb_duckdb.cpp
```diff
@@ -164,7 +164,11 @@ DuckDBManager::Initialize() {
 	auto &context = *connection->context;
 
 	auto &db_manager = duckdb::DatabaseManager::Get(context);
-	default_dbname = db_manager.GetDefaultDatabase(context);
+	if (duckdb_motherduck_default_database[0] == '\0') {
+		default_dbname = db_manager.GetDefaultDatabase(context);
+	} else {
+		default_dbname = duckdb_motherduck_default_database;
+	}
```
JelteF, line 171: Instead of overriding the default database that motherduck sets, let's instead make motherduck set the correct default database by providing the database in the connection string:

```c
connection_string = psprintf("md:%s?motherduck_token=%s", duckdb_motherduck_default_database, duckdb_motherduck_token);
```

For that you need to escape the database name though using URI escaping. I think the following function should work for that:
```c
char *
uri_escape(const char *input)
{
    StringInfoData buf;
    initStringInfo(&buf);

    while (*str)
    {
        char c = *str++;
        if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '-' || c == '_' || c == '.' || c == '~')
        {
            appendStringInfoChar(&buf, c);
        }
        else
        {
            appendStringInfo(&buf, "%%%02X", (unsigned char)c);
        }
    }

    return buf.data;
}
```

JelteF (changes requested):
Thanks a lot for the PR!

src/pgduckdb_duckdb.cpp
naoyak: Thanks for the tip! I didn't realize you could just stick the db name in the connection string, clean solution.

Is it ok that I just stuck the uri_escape function in the same file?

JelteF (approved):

JelteF:
Thanks a lot for your work on this. I played a bit with your changes and realized that `duckdb` tables were not being cleaned up when changing the default database. So I implemented a fix for that. I also updated did some small other tweaks, and updated the docs in some more places.

Y-- (approved):
LGTM!

JelteF (approved):

Issue #467, opened 2024-12-02
Unable to support huge number


dentiny:
### What happens?

`pg_duckdb` seems to fail on huge value insertion, in detail, prepared statement generation fails.

### To Reproduce

```
postgres (pid: 599) =# CREATE TEMP TABLE t (large_number NUMERIC) USING duckdb;
CREATE TABLE
postgres (pid: 599) =# INSERT INTO t (large_number) VALUES (200141183460469231731687303715884105727);
WARNING:  INSERT INTO pg_temp.main.t (large_number) VALUES ('200141183460469231731687303715884105727'::numeric)
ERROR:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Conversion Error: Could not convert string "200141183460469231731687303715884105727" to DECIMAL(18,3)
LINE 1: ...00141183460469231731687303715884105727'::numeric)
```

### OS:

Linux

### pg_duckdb Version (if built from source use commit hash):

535b092b7c7448e8f349a15243d76c63a6a73462

### Postgres Version (if built from source use commit hash):

17

### Hardware:

aarch64, 64-bit

### Full Name:

Hao

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [X] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [X] Yes, I have

dentiny:
Some initial investigation:
`query_string` produced b https://github.com/duckdb/pg_duckdb/blob/c3e80d8c27cc681380cf87b0e1f1ea95da87dcad/src/pgduckdb_planner.cpp#L35 default uses `numeric(18, 3)` as default type, which fails to parse.

JelteF:
NUMERIC in DuckDB works quite different than NUMERIC in Postgres. With [DuckDB the default numeric size is 18 (width), 3 (scale)](https://duckdb.org/docs/sql/data_types/numeric.html#fixed-point-decimals). You're trying to insert a larger number than that into the table. If you need larger numbers you can provide a different width and scale to use something different than the default.

DuckDB doesn't support numerics nearly as large as postgres though.

dentiny:
Understand, thanks you for the confirmation and explanation!

Issue #309, opened 2024-10-18
[bug]: Lose numeric zeros when rounding in duckdb


JelteF:
### Description

```sql
create table t(a numeric);
insert into t values (1.8000), (1.346);
select round(a, 2) from t;
```

With PG execution this returns:
```
 round
───────
  1.80
  1.35
```

With DuckDB execution this returns:
```
> select round(a, 2) from t;
 round
───────
   1.8
  1.35
```

So we somehow lose the trailing 0 after rounding to two decimals. This causes differences in TPCDS output.

PR #471, opened 2024-12-03
Fails value insertion if type unsupported by duckdb

dentiny:
I would like to followup on issue: https://github.com/duckdb/pg_duckdb/issues/467

I think current implementation is not ideal:
- If user creates a table with duckdb unsupported numeric type, we **silently** fallback to double type
- Later write operation **doesn't always** fail, based on the input numeric value, which adds more confusion to the users

Here I propose two improvements:
- method-1: No silent fallback on type conversion "failure"; so we print a warning message with `elog(WARNING, ...)`, and get users aware of the fallback behavior;
- method-2: Instead of falling back to double type, we fallback to user-defined type, since it's not supported by duckdb (system) anyway.

In this PR, I selected method-2, but I'm open to other feedbacks and suggestions.
The goal here is to disallow silent fallback and surprising write behavior.

Edit:
Followup on feedback https://github.com/duckdb/pg_duckdb/pull/471#issuecomment-2514571362, I fail the table creation if column type is not supported by duckdb.

### DIFF
--- a/src/pgduckdb_ddl.cpp
+++ b/src/pgduckdb_ddl.cpp
@@ -46,6 +46,45 @@ static bool ctas_skip_data = false;
 static bool top_level_ddl = true;
 static ProcessUtility_hook_type prev_process_utility_hook = NULL;
 
+// Get precision and scale from postgres type modifier.
+static int
+GetPgNumericPrecision(int32 typmod) {
+	return ((typmod - VARHDRSZ) >> 16) & 0xffff;
+}
+static int
+GetPgNumericScale(int32 typmod) {
+	return (((typmod - VARHDRSZ) & 0x7ff) ^ 1024) - 1024;
+}
+
+// If column is of type "numeric", check whether it's acceptable for duckdb;
+// If not, exception is thrown via `elog`.
+static void
+ValidateColumnNumericType(Form_pg_attribute &attribute) {
+	auto &type = attribute->atttypid;
+	if (type != NUMERICOID) {
+		return;
+	}
+	auto &typmod = attribute->atttypmod;
+	auto precision = GetPgNumericPrecision(typmod);
+	auto scale = GetPgNumericScale(typmod);
+	// duckdb's "numeric" type's max supported precision is 38.
+	if (typmod == -1 || precision < 0 || scale < 0 || precision > 38) {
+		elog(ERROR,
+		     "Unsupported type when creating column: type precision %d with scale %d is not supported by duckdb, which "
+		     "only allows maximum precision 38",
+		     precision, scale);
+	}
+}
+
+// Validate new relation creation, if invalid throw exception via `elog`.
+static void
+ValidateRelationCreation(TupleDesc desc) {
+	for (int i = 0; i < desc->natts; i++) {
+		Form_pg_attribute attr = &desc->attrs[i];
+		ValidateColumnNumericType(attr);
+	}
+}
+
 static void
 DuckdbHandleDDL(Node *parsetree) {
 	if (!pgduckdb::IsExtensionRegistered()) {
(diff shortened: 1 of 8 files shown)


src/pgduckdb_types.cpp
```diff
@@ -787,7 +789,7 @@ ConvertPostgresToDuckColumnType(Form_pg_attribute &attribute) {
 		break;
 	}
 	default: {
-		return duckdb::LogicalType::USER("UnsupportedPostgresType (Oid=" + std::to_string(type) + ")");
+		return duckdb::LogicalType::USER(duckdb::StringUtil::Format("UnsupportedPostgresType (Oid=%d)", type));
```
dentiny, line 792: `std::to_string` is inefficient:
- Underlyingly it calls `std::vsnprintf`,
- `std::sprintf` family internally invokes `std::localeconv_l` which acquires a mutex, thus hindering parallel execution

dentiny: Newer C++ standards (starting from C++26), STL implements `to_string` with `std::format`.

Some reference for the bad performance of `std::string`:
- https://en.cppreference.com/w/cpp/string/basic_string/to_string
- https://news.ycombinator.com/item?id=30472423

JelteF:
Thank you for the contribution.

The fact that we silently fall back to double has been bothering me too (see #309). I think it would be nice to make this a configurable setting that should probably be off by default, called something like `duckdb.use_double_for_large_numerics`.

However, afaict in the case you're describing this is actually not what is happening. In this case `NUMERIC` is converted to `DECIMAL(18, 3)`, not `double`.

```

**Comment by JelteF** [comment]
To be clear, I think it's probably reasonable to disallow specifying plain `NUMERIC` as the type for duckdb tables and require specifying a width and scale. But I don't think the current code actually has that effect.

**Comment by dentiny** [comment]
Both suggestion makes sense to me, thank you! I will make changes these days.

**Comment by dentiny** [comment]
Hi @JelteF , I tried to implemented your suggestion, which fails to create table if column type is not supported by duckdb.
I verified by existing SQL test statements with `make check`.

File: src/pgduckdb_ddl.cpp [hunk 2]
```diff
@@ -46,6 +46,45 @@ static bool ctas_skip_data = false;
...
+	if (type != NUMERICOID) {
+		return;
+	}
+	auto &typmod = attribute->atttypmod;
+	auto precision = GetPgNumericPrecision(typmod);
+	auto scale = GetPgNumericScale(typmod);
+	// duckdb's "numeric" type's max supported precision is 38.
+	if (typmod == -1 || precision < 0 || scale < 0 || precision > 38) {
+		elog(ERROR,
+		     "Unsupported type when creating column: type precision %d with scale %d is not supported by duckdb, which "
```
dpxcc, line 73: Add a test for this code path?

src/pgduckdb_ddl.cpp
```diff
@@ -253,6 +292,10 @@ DECLARE_PG_FUNCTION(duckdb_create_table_trigger) {
 		elog(ERROR, "Expected single table to be created, but found %" PRIu64, static_cast<uint64_t>(SPI_processed));
 	}
 
+	// Check whether new table creation is supported in duckdb.
+	// TODO(hjiang): Same type validation should be applied to other DDL as well.
+	ValidateRelationCreation(SPI_tuptable->tupdesc);
```
dpxcc, line 297: nit: prob move this check to `pgduckdb_get_tabledef()`, where other lockdowns (e.g., generated columns, identity columns) are handled

dentiny: I think it makes sense, my concern is, other DDL statements (i.e. ALTER TABLE) require the same validation logic.
I would prefer to put the validation logic in the same file.

dentiny: > where other lockdowns (e.g., generated columns, identity columns) are handled

That's a good point to place all validation logic into one function / one place.
My personal preference is to move all logic into DDL file, since other validations (i.e. generated columns, as you mentioned) should also apply to other DDL statements.

Would like to know how pg_duckdb's owner think about it. :)

src/pgduckdb_ddl.cpp
dentiny: Sure.

dentiny:
Hi @JelteF , I'm wondering if you could take another look when you have some time? Thank you!
Let me know if there's anything else I should change.

dentiny:
There's branch conflict and unit test failure, let me resolve them before re-requesting reviews.

dentiny:
@JelteF Sorry for the late response, and thanks for helping me trigger the CI last time.
I have fixed the failed sql test last time, should be good now.

JelteF:
FYI, I've not forgotten about this one. But it's been busy, and this PR is not trivial one to review. I hope to have time to look at it ~end of next week.

PR #477, opened 2024-12-04
Use PostgreSQL nodes for scanning tables

mkaruza:
Idea is to reconstruct query based on duckdb filtering information, for each table and use that information to plan postgres execution. This plan will, potentially exeucute with parallel workers. If no workers are available we will scan this local to thread.


src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,177 @@
...
+PostgresTableReader::GetNextTuple() {
+	MinimalTuple worker_minmal_tuple;
+	if (nreaders > 0) {
+		worker_minmal_tuple = GetNextWorkerTuple();
+		if (HeapTupleIsValid(worker_minmal_tuple)) {
+			ExecStoreMinimalTuple(worker_minmal_tuple, slot, false);
+			return slot;
+		}
+	} else {
+		std::lock_guard<std::mutex> lock(DuckdbProcessLock::GetLock());
```
JelteF, line 122: Instead of taking the lock for every tuple I expect it might actually result in better performance to take the lock to fill a whole `DataChunk`, so in `PostgresScanTableFunction::PostgresScanFunction`.

That way we're not releasing and acquiring the lock so often. Obviously it has the downside that the lock is held for a longer time. So if you're reading from multiple postgres tables, one of those threads would be paused for a longer time. But I expect that's actually better for performance. Because with these very short locks we're switching back and forth between those threads, and that probably results in lots of thrashing of the L1 cache etc.

src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		next_parallel_reader++;
+		if (next_parallel_reader >= nreaders)
+			next_parallel_reader = 0;
+
+		nvisited++;
+		if (nvisited >= nreaders) {
+			(void)WaitLatch(MyLatch, WL_LATCH_SET | WL_EXIT_ON_PM_DEATH, 0, WAIT_EVENT_EXECUTE_GATHER);
+			ResetLatch(MyLatch);
```
JelteF, line 171: Calling these functions requires the global lock, and it doesn't seem like we're holding it here.

JelteF (commented):
Initial round of review. I really like the direction of the code. One general feedback is that this definitely needs a bunch of tests too.

src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include <cmath>
+
+namespace pgduckdb {
+
+PostgresTableReader::PostgresTableReader(Cardinality cardinality, const char *table_scan_query)
+    : parallel_worker_readers(nullptr), nreaders(0), next_parallel_reader(0) {
+
+	std::lock_guard<std::mutex> lock(DuckdbProcessLock::GetLock());
+
+	pg_stack_base_t current_stack = set_stack_base();
```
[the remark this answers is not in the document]
mkaruza: Postgresql asserts because of stack depth exceeds. With this we reset/set base of stack to current memory and returns to previous one once we exit function.

include/pgduckdb/pgduckdb_process_latch.hpp
```diff
@@ -0,0 +1,46 @@
...
+		static std::mutex lock;
+		if (lock.try_lock()) {
+			WaitLatch(MyLatch, WL_LATCH_SET | WL_EXIT_ON_PM_DEATH, 0, PG_WAIT_EXTENSION);
+			ResetLatch(MyLatch);
+			cv.notify_all();
+			lock.unlock();
+		} else {
+			std::unique_lock<std::mutex> lk(lock);
+			cv.wait(lk);
+		}
```
JelteF, line 42: Could you explain a bit what the goal is of this logic. I had expected it to basically be:

1. Take global process lock
2. Call WaitLatch
3. Call ResetLatch
4. Unlock global process lock 

Is this specific logic, for performance reasons? Or is it necessary to avoid concurrency bugs? e.g. is this meant to make sure that if two threads wait for the latch, both get notified instead of one thread "eating" the notification that the other was waiting for?

lepsa0321:
I've been looking into the PR you submitted, and I'm curious about its impact on index-scan functionality. Could you please clarify if this PR addresses the issue where index-scans were not being used when expected?

If it does resolve this problem, could you also provide guidance on how to enforce the use of an index-scan? Specifically:

Are there any configuration settings or parameters that can be adjusted to prefer index-scans over other scan methods?
Is there a way to hint or force the query planner to choose an index-scan for certain queries?
Any additional steps or considerations we should be aware of when trying to ensure that index-scans are utilized?
Your insights would be greatly appreciated as they will help us better understand the improvements and how to leverage them effectively in our environment.

Thank you

include/pgduckdb/pgduckdb_process_latch.hpp
mkaruza: We can wait only on single Latch in process. So first thread arriving will wait on latch while others will wait for condition variable (that will be released once latch waiting is done). Idea was to "simulate" different threads to wait on latch.

mkaruza:
> I've been looking into the PR you submitted, and I'm curious about its impact on index-scan functionality. Could you please clarify if this PR addresses the issue where index-scans were not being used when expected?
> 
> If it does resolve this problem, could you also provide guidance on how to enforce the use of an index-scan? Specifically:
> 
> Are there any configuration settings or parameters that can be adjusted to prefer index-scans over other scan methods? Is there a way to hint or force the query planner to choose an index-scan for certain queries? Any additional steps or considerations we should be aware of when trying to ensure that index-scans are utilized? Your insights would be greatly appreciated as they will help us better understand the improvements and how to leverage them effectively in our environment.
> 
> Thank you

Hi, i have added regression test that shows that Bitmap/Index/IndexOnly scan is used when appropriate - hopefully that will help you answer your question.

lepsa0321:
> > I've been looking into the PR you submitted, and I'm curious about its impact on index-scan functionality. Could you please clarify if this PR addresses the issue where index-scans were not being used when expected?
> > If it does resolve this problem, could you also provide guidance on how to enforce the use of an index-scan? Specifically:
> > Are there any configuration settings or parameters that can be adjusted to prefer index-scans over other scan methods? Is there a way to hint or force the query planner to choose an index-scan for certain queries? Any additional steps or considerations we should be aware of when trying to ensure that index-scans are utilized? Your insights would be greatly appreciated as they will help us better understand the improvements and how to leverage them effectively in our environment.
> > Thank you
> 
> Hi, i have added regression test that shows that Bitmap/Index/IndexOnly scan is used when appropriate - hopefully that will help you answer your question.

Thank you very much~

src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+std::string
+PostgresTableReader::ExplainScanPlan(QueryDesc *query_desc) {
+	ExplainState *es = (ExplainState *)PostgresFunctionGuard(palloc0, sizeof(ExplainState));
+	es->str = makeStringInfo();
+	es->format = EXPLAIN_FORMAT_TEXT;
+	PostgresFunctionGuard(ExplainPrintPlan, es, query_desc);
+	std::string explain_scan(es->str->data);
+	return explain_scan;
+}
```
Y--, line 163: Maybe we want to return a `const char*`? And as mentioned offline, either have caller wrap with `PostgresFunctionGuard` or introduce a new function and wrap it here?
```suggestion
const char*
PostgresTableReader::ExplainScanPlan_Unsafe(QueryDesc *query_desc) {
	ExplainState *es = (ExplainState *)palloc0(sizeof(ExplainState));
	es->str = makeStringInfo();
	es->format = EXPLAIN_FORMAT_TEXT;
	ExplainPrintPlan(es, query_desc);
	return es->str->data;
}

const char*
PostgresTableReader::ExplainScanPlan(QueryDesc *query_desc) {
	return PostgresFunctionGuard(ExplainScanPlan_Unsafe, query_desc);
}
```

src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,261 @@
...
+		worker_minmal_tuple = GetNextWorkerTuple();
+		if (HeapTupleIsValid(worker_minmal_tuple)) {
+			PostgresFunctionGuard(ExecStoreMinimalTuple, worker_minmal_tuple, slot, false);
+			return slot;
+		}
+	} else {
+		std::lock_guard<std::mutex> lock(GlobalProcessLock::GetLock());
+		PostgresScopedStackReset scoped_stack_reset;
+		table_scan_query_desc->estate->es_query_dsa = parallel_executor_info ? parallel_executor_info->area : NULL;
+		thread_scan_slot = ExecProcNode(table_scan_planstate);
```
Y--, line 211: Shouldn't this be `PostgresFunctionGuard`-ed?

src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,261 @@
...
+	return PostgresFunctionGuard(ExecClearTuple, slot);
+}
+
+MinimalTuple
+PostgresTableReader::GetNextWorkerTuple() {
+	int nvisited = 0;
+	for (;;) {
+		TupleQueueReader *reader;
+		MinimalTuple minimal_tuple;
+		bool readerdone;
```
Y--, line 227: nit- we should either hoist out of the loop or inline no?

include/pgduckdb/scan/postgres_scan.hpp
```diff
@@ -5,52 +5,80 @@
...
-		}
-	}
+	PostgresScanLocalState(PostgresScanGlobalState *global_state);
+	~PostgresScanLocalState() override;
 
-	uint32_t m_output_vector_size;
-	bool m_exhausted_scan;
-	std::vector<Datum, DuckDBMallocator<Datum>> values;
-	std::vector<bool, DuckDBMallocator<bool>> nulls;
+public:
```
Y--, line 46: `struct` members are public by default - so this (and the one above) are redundant; same in the other ones below.

```suggestion
```

src/scan/postgres_scan.cpp
```diff
@@ -78,99 +68,161 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
...
-	List *name_list = NIL;
-	if (!schema.empty()) {
-		name_list = lappend(name_list, makeString(pstrdup(schema.c_str())));
+	scan_query << " FROM " << ConstructFullyQualifiedTableName();
+
+	first = true;
+
+	for (auto const &[attr_num, duckdb_scanned_index] : columns_to_scan) {
+		auto filter = column_filters[duckdb_scanned_index];
+		if (filter) {
```
Y--, line 98: nit
```suggestion
		if (!filter) { continue; }
```

src/scan/postgres_scan.cpp
```diff
@@ -78,99 +68,161 @@ PostgresScanGlobalState::InitGlobalState(duckdb::TableFunctionInitInput &input)
...
+		InsertTupleIntoChunk(output, local_state, slot);
 	}
 
-	auto relForm = (Form_pg_class)GETSTRUCT(tuple);
-	if (relForm->relkind != RELKIND_VIEW) {
-		PostgresFunctionGuard(ReleaseSysCache, tuple);
-		return nullptr;
+	/* If we finish before reading complete vector means that scan was exhausted. */
+	if (i != STANDARD_VECTOR_SIZE) {
+		local_state.exhausted_scan = true;
```
Y--, line 216: Isn't it already done L207? And since the only other way to exit the loop above is to have `i==STANDARD_VECTOR_SIZE` this branch is redundant no?

include/pgduckdb/scan/postgres_scan.hpp
```diff
@@ -5,52 +5,80 @@
...
-	duckdb::map<int, Datum> m_relation_missing_attrs;
+private:
+	std::string ConstructFullyQualifiedTableName();
+
+public:
+	Snapshot snapshot;
+	Relation rel;
+	TupleDesc table_tuple_desc;
+	bool count_tuples_only;
+	duckdb::vector<duckdb::pair<duckdb::idx_t, AttrNumber>> output_columns;
```
Y--, line 33: It seems that the `index` isn't used anymore- we could simplify that:
```suggestion
	duckdb::vector<AttrNumber> output_columns;
```

src/scan/postgres_table_reader.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+	parallel_worker_readers = NULL;
+
+	PostgresFunctionGuard(ExecutorFinish, table_scan_query_desc);
+	PostgresFunctionGuard(ExecutorEnd, table_scan_query_desc);
+	PostgresFunctionGuard(FreeQueryDesc, table_scan_query_desc);
+
+	if (entered_parallel_mode) {
+		ExitParallelMode();
+	}
```
Y--, line 144: It seems that it would be worth moving all that in a `Dispose` private method that would be called here wrapped only once by `PostgresFunctionGuard`?

include/pgduckdb/catalog/pgduckdb_table.hpp
```diff
@@ -11,35 +11,27 @@ namespace pgduckdb {
...
+
 	virtual ~PostgresTable();
 
+public:
+	duckdb::unique_ptr<duckdb::BaseStatistics> GetStatistics(duckdb::ClientContext &context,
+	                                                         duckdb::column_t column_id) override;
+	duckdb::TableFunction GetScanFunction(duckdb::ClientContext &context,
+	                                      duckdb::unique_ptr<duckdb::FunctionData> &bind_data) override;
+	duckdb::TableStorageInfo GetStorageInfo(duckdb::ClientContext &context) override;
+
```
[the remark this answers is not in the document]
mkaruza: It doesn't look that these virtual functions are defined as const in `TableCatalogEntry`. (comment also for below remark)

include/pgduckdb/pgduckdb_process_latch.hpp
JelteF: As discussed over a call, we need some counter here to make sure that one thread doesn't consume the notifcations for another thread. Something like this:
```c
		if (lock.lock()) {
			static global_counter = 0;
			if (previous_counter != global_counter) { /* previous_counter should be thread_local */
				WaitLatch(MyLatch, WL_LATCH_SET | WL_EXIT_ON_PM_DEATH, 0, PG_WAIT_EXTENSION);
				ResetLatch(MyLatch);
				global_counter++;
			}
			previous_counter = global_counter; 
			lock.unlock();
		}
```

include/pgduckdb/pgduckdb_guc.h
```diff
@@ -8,7 +8,7 @@ extern bool duckdb_enable_external_access;
 extern bool duckdb_allow_unsigned_extensions;
 extern bool duckdb_autoinstall_known_extensions;
 extern bool duckdb_autoload_known_extensions;
-extern int duckdb_max_threads_per_postgres_scan;
+extern int max_workers_per_postgres_scan;
```
JelteF, line 11: Let's prefix this with `duckdb_` to not polute the global namespace with a fairly generic name.
```suggestion
extern int duckdb_max_workers_per_postgres_scan;
```

src/pgduckdb_hooks.cpp
```diff
@@ -181,8 +197,7 @@ IsAllowedStatement(Query *query, bool throw_error = false) {
 	/*
 	 * When accessing the partitioned table, we temporarily let PG handle it instead of DuckDB.
 	 */
-	if (ContainsPartitionedTable(query->rtable)) {
-		elog(elevel, "DuckDB does not support querying PG partitioned table");
+	if (!ContainsAllowedTableType(query->rtable, elevel)) {
```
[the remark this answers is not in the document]
mkaruza: Removed check.

src/scan/postgres_table_reader.cpp
```diff
@@ -156,12 +156,26 @@ PostgresTableReader::PostgresTableReaderCleanup() {
...
-	static const int base_log = 8;
-	int cardinality_log = std::log2(cardinality);
-	int base = cardinality_log / base_log;
-	return std::max(1, std::min(base, std::max(duckdb_max_workers_per_postgres_scan, max_parallel_workers)));
+	static const int cardinality_threshold = 1 << 16;
+	/* No parallel scan wanted */
+	if (!duckdb_max_workers_per_postgres_scan) {
+		return 0;
+	}
+	/* */
```
JelteF, line 174: empty comment?

JelteF (approved):

JelteF (approved):

JelteF (approved):

PR #484, opened 2024-12-09
Add `autoload` and `autoinstall` GUC

Y--:
Add two GUC settings: `duckdb_autoinstall_known_extensions` and `duckdb_autoload_known_extensions`, which are direct proxy for DuckDB [settings](https://duckdb.org/docs/configuration/overview.html) `autoinstall_known_extensions` and `autoload_known_extensions`.

Mark both option enabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,8 +157,7 @@ Querying data stored in Parquet, CSV, JSON, Iceberg and Delta format can be done
 	LIMIT 100;
 	```
 
-Note, for Azure, you will need to first install the Azure extension:
-You may then store a secret using the `connection_string` parameter as such:
+Note, for Azure, you may store a secret using the `connection_string` parameter as such:
 ```sql
 INSERT INTO duckdb.secrets
 (type, connection_string)
--- a/README.md
+++ b/README.md
@@ -158,10 +158,6 @@ Querying data stored in Parquet, CSV, JSON, Iceberg and Delta format can be done
 	```
 
 Note, for Azure, you will need to first install the Azure extension:
-```sql
-SELECT duckdb.install_extension('azure');
-```
-
 You may then store a secret using the `connection_string` parameter as such:
 ```sql
 INSERT INTO duckdb.secrets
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -158,10 +158,6 @@ Querying data stored in Parquet, CSV, JSON, Iceberg and Delta format can be done
 	```
 
 Note, for Azure, you will need to first install the Azure extension:
```
JelteF, line 160: This line should be removed I guess?

JelteF: Actually it seems the intro should just be rewritten, because the "You may then..." sentence on its own doesn't make sense either.

Y--: Ugh hadn't saved the file before pushing. I've pushed my last iteration, let me know

src/pgduckdb.cpp
```diff
@@ -131,6 +133,14 @@ DuckdbInitGUC(void) {
 	                     "Allow DuckDB to load extensions with invalid or missing signatures",
 	                     &duckdb_allow_unsigned_extensions, PGC_SUSET);
 
+	DefineCustomVariable("duckdb.autoinstall_known_extensions",
+	                     "Whether known extensions are allowed to be automatically installed when a DuckDB query depends on them",
+	                     &duckdb_autoinstall_known_extensions, PGC_POSTMASTER, GUC_SUPERUSER_ONLY);
+
+	DefineCustomVariable("duckdb.autoload_known_extensions",
+	                     "Whether known extensions are allowed to be automatically loaded when a DuckDB query depends on them",
+	                     &duckdb_autoload_known_extensions, PGC_POSTMASTER, GUC_SUPERUSER_ONLY);
```
JelteF, line 142: I think PGC_SUSET is fine for these. It seems alright to change these after startup (as long as you're superuser).
```suggestion
	                     &duckdb_autoinstall_known_extensions, PGC_SUSET);

	DefineCustomVariable("duckdb.autoload_known_extensions",
	                     "Whether known extensions are allowed to be automatically loaded when a DuckDB query depends on them",
	                     &duckdb_autoload_known_extensions, PGC_SUSET);
```

JelteF (approved):

PR #491, opened 2024-12-12
Correct handling parameter

### LINKED ISSUE #234: support "PREPARE" statement and "EXECUTE " statement
### Current Behavior
```plpgsql
postgres=# create extension pg_duckdb;
WARNING:  To actually execute queries using DuckDB you need to run "SET duckdb.execution TO true;"
CREATE EXTENSION
postgres=# SET duckdb.execution TO true;
SET
postgres=# SELECT COUNT(*) FROM t;
 count 
-------
  1000
(1 row)

postgres=# PREPARE test_stmt AS SELECT COUNT(*) FROM t;
PREPARE
postgres=# EXECUTE test_stmt;
 count 
-------
     0
(1 row)

```

### LINKED ISSUE #411: Any stored procedure is not working
### What happens?

Stored procedure will run each query with more parameters, thus will always hit assert in duckdb.

### To Reproduce

```
postgres=# create table t(a int);
CREATE TABLE
postgres=# SET duckdb.force_execution TO true;
SET
postgres=# CREATE OR REPLACE PROCEDURE t(b INT)
LANGUAGE plpgsql
AS $$
BEGIN
    select * from t where a=b;
END;
$$;
CREATE PROCEDURE
postgres=# call t(1);
ERROR:  (PGDuckDB/Duckdb_ExecCustomScan) Invalid Input Error: Parameter argument/count mismatch, identifiers of the excess parameters: 2
CONTEXT:  SQL statement "select * from t where a=b"
PL/pgSQL function t(integer) line 3 at SQL statement

```

### OS:

MacOs

### pg_duckdb Version:

0.1.0

### Postgres Version:

16

### Hardware:

_No response_

### Full Name:

Zhou

### Affiliation:

Mooncake Labs

I have tested with a stable release

- [X] Yes, I have

- [X] Yes, I have

Leo-XM-Zeng:
DuckDB complains when it receives parameters that were not actually used in the query. From Postgres we get all supplied parameters though. So this changes the code to filter the unused ones out. To do this we lot DuckDB parse the query and tell us which parameters it expects, and then we remove the unexpected ones.

Fixes #411 
Fixes #234

### DIFF
--- a/src/pgduckdb_node.cpp
+++ b/src/pgduckdb_node.cpp
@@ -11,13 +11,11 @@ extern "C" {
 #include "miscadmin.h"
 #include "tcop/pquery.h"
 #include "nodes/params.h"
-#include "nodes/nodeFuncs.h"
 #include "utils/ruleutils.h"
 }
 
 #include "pgduckdb/pgduckdb_node.hpp"
 #include "pgduckdb/pgduckdb_duckdb.hpp"
-#include "pgduckdb/vendor/pg_list.hpp"
 #include "pgduckdb/utility/cpp_wrapper.hpp"
 
 /* global variables */
@@ -61,7 +59,6 @@ static TupleTableSlot *Duckdb_ExecCustomScan(CustomScanState *node);
 static void Duckdb_EndCustomScan(CustomScanState *node);
 static void Duckdb_ReScanCustomScan(CustomScanState *node);
 static void Duckdb_ExplainCustomScan(CustomScanState *node, List *ancestors, ExplainState *es);
-static bool Get_ParamList(Node *node, void **context);
 
 static Node *
 Duckdb_CreateCustomScanState(CustomScan *cscan) {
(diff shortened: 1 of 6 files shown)


test/regression/sql/prepare.sql
```diff
@@ -0,0 +1,54 @@
...
+
+-- q1 q2 q3
+SELECT name, statement, parameter_types FROM pg_prepared_statements
+    ORDER BY name;
+
+CREATE TABLE tb (a int DEFAULT 1, b int, c varchar DEFAULT 'pg_duckdb');
+INSERT INTO tb(b) VALUES(2);
+PREPARE q4(int, varchar) AS
+	SELECT b FROM tb WHERE a = $1 AND c = $2;
+EXECUTE q4(1, 'pg_duckdb');
```
JelteF, line 42: I'd like to see a query too where the same parameter is used multiple times

test/regression/sql/prepare.sql
```diff
@@ -0,0 +1,54 @@
+-- the view should return the empty
+SELECT name, statement, parameter_types FROM pg_prepared_statements;
+
+CREATE DATABASE testdb;
+CREATE TABLE copy_database(datname text, datistemplate boolean, datallowconn boolean);
+INSERT INTO copy_database SELECT datname, datistemplate, datallowconn FROM pg_database;
```
JelteF, line 6: What's the reason you're creating a new database for this query?

test/regression/sql/prepare.sql
Leo-XM-Zeng: done

test/regression/sql/prepare.sql
Leo-XM-Zeng: I changed it to postgres now

zhousun:
@Leo-XM-Zeng thx for the pr, can you check whether https://github.com/Mooncake-Labs/pg_mooncake/commit/a6262b1591b870ccd9d7bf3b30c18aa3345b9fbc would be an easier way to fix the same issue?

test/regression/sql/prepare.sql
JelteF: I think you misunderstood what I meant. I didn't mean that the statement should be executed multiple times. I meant that we should have a query where the same parameter is used multiple times.

```
PREPARE q5(int, varchar) AS
	SELECT b FROM tb WHERE a = $1 AND b = $1 AND c = $2;
```

And I think it would be good to also check versions where one of the parameters is not used (both the first parameter and the last parameter unused is interesting I think):
```
PREPARE q6(int, varchar) AS
	SELECT b FROM tb WHERE c = $2;
PREPARE q7(int, varchar) AS
	SELECT b FROM tb WHERE a = $1;
```

src/pgduckdb_node.cpp
```diff
@@ -99,28 +102,45 @@ ExecuteQuery(DuckdbScanState *state) {
...
+#if PG_VERSION_NUM >= 150000
+		foreach_node(Param, param, paramlist) {
+#else
+	 	foreach_ptr(Param, param, paramlist) {
+#endif
+			if (param->paramkind == PARAM_EXTERN &&
+				paramLI != NULL && param->paramid > 0 &&
+				param->paramid <= paramLI->numParams) {
+				ParamExternData *prm;
+				ParamExternData prmdata;
```
JelteF, line 124: Can you revert the naming changes for these two? It makes it quite hard to see what actual code changes you made if you also change the naming of existing variables at the same time.

JelteF:
> I can't judge whether your changes are good or bad, perhaps we need to consult someone who has a deeper understanding of these.

Looking at the changes linked by @zhousun, those are indeed quite simple: They let duckdb figure out what parameters are in the query, and then we only pass those parameters to duckdb from the full set of parameters that postgres has. If that indeed solves all these issues, then I like that approach better than us traversing the query tree ourselves to find all the parameters.

@zhousun are you okay with @Leo-XM-Zeng changing this PR to incorporate those changes?

Leo-XM-Zeng:
> Looking at the changes linked by @zhousun, those are indeed quite simple: They let duckdb figure out what parameters are in the query, and then we only pass those parameters to duckdb from the full set of parameters that postgres has. If that indeed solves all these issues, then I like that approach better than us traversing the query tree ourselves to find all the parameters.

@zhousun Your code is much more perfect than mine, I just tested and validated it and found that it handles these scenarios much better. That would be a better solution.

Leo-XM-Zeng:
> @Leo-XM-Zeng thx for the pr, can you check whether [Mooncake-Labs/pg_mooncake@a6262b1](https://github.com/Mooncake-Labs/pg_mooncake/commit/a6262b1591b870ccd9d7bf3b30c18aa3345b9fbc) would be an easier way to fix the same issue?

If you want to submit your own PR, I will turn this off at that time, the current PR code is from you, I just added some test cases.

test/regression/sql/prepare.sql
Leo-XM-Zeng: Related test cases have been added, I really misunderstood your meaning, thank you for your guidance

zhousun:
@Leo-XM-Zeng just take it :)

Leo-XM-Zeng:
> just take it :)

Thank you ~ 😀

JelteF (approved):

JelteF:
Thanks both, I used the `Co-authored-by:` trailer in the git commit to also attribute @zhousun

zhaonaiy:
Hello, I was wondering if this bug has been fixed? checked in which version? Thanks.

Leo-XM-Zeng:
> Hello, I was wondering if this bug has been fixed? checked in which version? Thanks.

You can try compiling the main branch code.

image

PR #496, opened 2024-12-16
Bump clang-format and ruff versions to latest

JelteF:
In #98 we started using the same version of clang-format that DuckDB uses. That version is pretty ancient though. So it misses a bunch of features that we might want to use, and also has a bunch of bugs which result in sub-optimal formatting. This updates `clang-format` to the latest version, and while we're at it we also update `ruff`.

To be able to reformat all existing files, this also adds a `format-all` target to our `Makefile`.

The problem was called out by @dpxcc in #493, but I've been also running into this myself for a while.

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -133,13 +133,15 @@ DuckdbInitGUC(void) {
 	                     "Allow DuckDB to load extensions with invalid or missing signatures",
 	                     &duckdb_allow_unsigned_extensions, PGC_SUSET);
 
-	DefineCustomVariable("duckdb.autoinstall_known_extensions",
-	                     "Whether known extensions are allowed to be automatically installed when a DuckDB query depends on them",
-	                     &duckdb_autoinstall_known_extensions, PGC_SUSET);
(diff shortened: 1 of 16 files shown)


src/pgduckdb.cpp
```diff
@@ -133,13 +133,15 @@ DuckdbInitGUC(void) {
...
 	                     &duckdb_allow_unsigned_extensions, PGC_SUSET);
 
-	DefineCustomVariable("duckdb.autoinstall_known_extensions",
-	                     "Whether known extensions are allowed to be automatically installed when a DuckDB query depends on them",
-	                     &duckdb_autoinstall_known_extensions, PGC_SUSET);
-
-	DefineCustomVariable("duckdb.autoload_known_extensions",
-	                     "Whether known extensions are allowed to be automatically loaded when a DuckDB query depends on them",
-	                     &duckdb_autoload_known_extensions, PGC_SUSET);
+	DefineCustomVariable(
```
Y--, line 136: Interesting that the other in this file weren't changed?

Y-- (approved):
Thanks!

PR #499, opened 2024-12-17
Support approx_count_distinct

dpxcc:
Add support for `approx_count_distinct()`


sql/pg_duckdb--0.2.0--0.3.0.sql
```diff
@@ -0,0 +1,9 @@
+CREATE FUNCTION @extschema@.approx_count_distinct(a anyelement)
```
JelteF, line 1: This should no be added as a function, but as a function but as an aggregate. Currently the following query fails on your test schema because of that:
```
> SELECT approx_count_distinct(a) FROM t group by b;
ERROR:  42803: column "t.a" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: SELECT approx_count_distinct(a) FROM t group by b;
```

JelteF, line 1: I'm wondering if this should be created in `@extschema@` (i.e. the `public` schema) or the `duckdb` schema. I think `@extschema@` is probably fine.

dpxcc: Thanks for the catch!!

dpxcc:
Change `approx_count_distinct` from a user-defined function to a user-defined aggregate

sql/pg_duckdb--0.2.0--0.3.0.sql
dpxcc: I don't have a strong opinion. I was just following the existing functions

dpxcc:
@JelteF Could you please take a look again? Thanks!

JelteF (approved):

Issue #116, opened 2024-08-13
Case insensitivity mismatch between duckdb and postgres can cause issues


JelteF:
DuckDB considers identifiers with different casing the same even if they are quoted.

The most minimal example of this is the following schema:

```
CREATE TABLE a(b int);
CREATE TABLE "A"(c int);
INSERT INTO a VALUES(1);
```
When using postgres execution this correctly outputs:
```
> SELECT * FROM a, "A" as aa;
 b │ c
───┼───
(0 rows)
```
But when enabling duckdb execution both `a` and `"A"` are interpreted as `a`. Resulting in the following incorrect result:
```
> SELECT * FROM a, "A" as aa;
 b │ b
───┼───
 1 │ 1
(1 row)
```

This happens even when I set `preserver_identifier_case` to `true` on the DuckDB connection.

wuputah:
some related discussion here: https://github.com/duckdb/pg_duckdb/issues/56#issuecomment-2270785294, and related to #43

Alzavio:
This may be related, but in either of the two below, the field name cases aren't maintained in the new postgres table

`CREATE TABLE postgres_staging.public.{table_name} AS SELECT * FROM postgres_prod.public.{table_name}`

```
duckdb.sql(f'CREATE TABLE {table_name} AS SELECT * FROM postgres_prod.public.{table_name}')
duckdb.sql(f'CREATE TABLE postgres_staging.public.{table_name} AS SELECT * FROM {table_name}')
```

pantonis:
This shouldn't be in the long term milestone. but an important bug to fix.

Y--:
@pantonis we'll try to address this asap.

JelteF:
Based on some research I did this should be quite easy to address, postgres_scanner is able to do this. I'm pretty sure it's a few lines of code only. So we can use the same approach as is used there. Let's wait until #477 is merged though, to avoid any merge conflicts.

pantonis:
We had the same issue when we were using Hydra and it was fixed quite fast I can say.

Leo-XM-Zeng:
This seems to be solved in the current version. 

> We had the same issue when we were using Hydra and it was fixed quite fast I can say.

JelteF:
If it's indeed resolved we should add some tests to make sure we don't regress again.

CaselIT:
Not sure if it's related, but it seems you can't change the casing of the names with as if the name is upper when reading from parquet.

create a parquet, like
[...]

PR #504, opened 2024-12-20
Add basic test cases about #116

Leo-XM-Zeng:
Add basic test cases 
#116

### DIFF
--- a/test/regression/expected/basic.out
+++ b/test/regression/expected/basic.out
@@ -81,3 +81,23 @@ SELECT * FROM t JOIN s ON a = b;
 
 DROP TABLE t;
 DROP TABLE s;
+CREATE TABLE a  (b int);
+CREATE TABLE "A"(c int);
+INSERT INTO a VALUES(1);
+SET duckdb.force_execution = false;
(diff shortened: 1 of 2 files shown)


test/regression/expected/basic.out
```diff
@@ -81,3 +81,23 @@ SELECT * FROM t JOIN s ON a = b;
...
+INSERT INTO a VALUES(1);
+SET duckdb.force_execution = false;
+SELECT * FROM a, "A" as aa;
+ b | c 
+---+---
+(0 rows)
+
+SET duckdb.force_execution = true;
+SELECT * FROM a, "A" as aa;
+WARNING:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Binder Error: Table "aa" does not have a column named "c"
```
JelteF, line 95: So, the error is still there right?

Leo-XM-Zeng: Yes, there will be this warning, because DuckDB failed to run, so it was run by PG, which makes the result ultimately consistent with PG. If we don't want to see this warning here, maybe we need to wait for #477 as you said.

PR #511, opened 2024-12-22
Add support for BYTEA/BLOB

### LINKED ISSUE #464: Conversion: DuckDB Type BLOB -> postgres bytea failed
### What happens?

When querying parquet that has blob columns (some engines write strings as blob) postrges can't convert them into `bytea`

### To Reproduce

```
select * from read_parquet('s3://bucket/file.parquet') as (product_id varchar(255));
```
```
select * from read_parquet('s3://bucket/file.parquet') as (product_id bytea);
```
the same issue occured on both convertions
```
WARNING:  (PGDuckDB/GetPostgresDuckDBType) Could not convert DuckDB type: BLOB to Postgres type
ERROR:  (PGDuckDB/CreatePlan) Cache lookup failed for type 0
```

### OS:

Linux aarch64

0.1.0

15.10

### Hardware:

_No response_

### Full Name:

Bobirjon Mardonov

### Affiliation:

Korzinka GO

I have tested with a stable release

- [X] Yes, I have

- [X] Yes, I have

ritwizsinha:
Fixes #464


ritwizsinha:
@JelteF currently the  output from `select * from blob_tbl` is an escaped version of inserted rows, is this fine or do we need to change that?

test/regression/expected/type_support.out
```diff
@@ -308,6 +308,27 @@ SELECT * FROM json_tbl;
...
+    ('\x07'),
+    (NULL)
+) t(a);
+SELECT * from blob_tbl;
+                                                                 a                                                                  
+------------------------------------------------------------------------------------------------------------------------------------
+ \x
+ \x5c7831315c7830315c7830325c7830335c7830345c7830355c7830365c7830375c7830385c7830395c7830415c7830425c7830435c7830445c7830455c783046
+ \x
+ \x5c783030
```
JelteF, line 327: > currently the output from select * from blob_tbl is an escaped version of inserted rows, is this fine or do we need to change that?

Okay, this needs to change. All the non empty bytea results are wrong. When using postgres execution the it instead gives the following rows:
```
 \x
 \x110102030405060708090a0b0c0d0e0f
 \x
 \x00
 \x07

 ```

It seems like you're encoding the string as hex somewhere as an additional time, because `5c783030` is `\x00` in ASCII, i.e. including the `\` and `x` characters. So I think when converting between from DuckDB to PG type you're encoding the string representation of the type instead of the raw bytes.

Comment (ritwizsinha) [reply test/regression/expected/type_support.out:327]: Found it, converting it to string was trying to convert those bytes to string

Comment (ritwizsinha) [reply test/regression/expected/type_support.out:327]: Changed GetValue to GetValueUnsafe for getting raw bytes in string_t

File: test/regression/expected/type_support.out [hunk 2]
```diff
@@ -308,6 +308,27 @@ SELECT * FROM json_tbl;
...
+                 a                  
+------------------------------------
+ \x
+ \x110102030405060708090a0b0c0d0e0f
+ \x
+ \x00
+ \x07
+ 
+(6 rows)
+
```
JelteF, line 331: Can you add a comparison test too? Something like:

SELECT * FROM blob_tbl WHERE a = '\x00';

JelteF (approved):

src/pgduckdb_types.cpp
```diff
@@ -199,6 +200,17 @@ ConvertVarCharDatum(const duckdb::Value &value) {
 	return PointerGetDatum(result);
 }
 
+static Datum
+ConvertBinaryDatum(const duckdb::Value &value) {
+	auto str = value.GetValueUnsafe<duckdb::string_t>();
+	auto blob_len = str.GetSize();
+	auto blob = str.GetDataUnsafe();
+	bytea* result = (bytea *)palloc0(blob_len + VARHDRSZ);
```
dpxcc, line 208: nit: `make format` will re-format this line as
```
bytea *result = (bytea *)palloc0(blob_len + VARHDRSZ);
```

JelteF: Fixed by: https://github.com/duckdb/pg_duckdb/pull/518

PR #513, opened 2024-12-31
JSONB support

ritwizsinha:
Adding support for JSONB and JSONB[]


JelteF (changes requested):
Thanks! These tests cover some nice ground around JSON support. Sadly half of them result in warnings at the moment, but I think it'd be good to add them already anyway. At least to make sure that we don't crash/error, but fall back gracefully to Postgres execution.

test/regression/expected/json_operator_support.out
```diff
@@ -0,0 +1,334 @@
...
+);
+-- Insert test data
+INSERT INTO test_json (data) VALUES
+('{"a": 1, "b": {"c": 2, "d": [3, 4]}, "e": "hello"}'),
+('{"f": 10, "g": {"h": 20, "i": 30}, "j": [40, 50, 60]}'),
+('{"k": true, "l": null, "m": {"n": "world", "o": [7, 8, 9]}}');
+-- Test Case 1: Access JSON Object Field (->)
+SELECT id, data->'a' AS a_value
+FROM test_json;
+WARNING:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Binder Error: No function matches the given name and argument types 'json_extract("UnsupportedPostgresType (Oid=3802)", VARCHAR)'. You might need to add explicit type casts.
```
JelteF, line 53: It would be really good to start supporting JSONB.

test/regression/expected/json_operator_support.out
```diff
@@ -0,0 +1,334 @@
...
+set duckdb.force_execution to true;
+-- -> int operator
+select '[{"a":"foo"},{"b":"bar"},{"c":"baz"}]'::json->2;
+  ?column?   
+-------------
+ {"c":"baz"}
+(1 row)
+
+-- -> text operator
+select '{"a": {"b":"foo"}}'::json->'a'
```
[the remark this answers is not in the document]
ritwizsinha: Removed them

src/pgduckdb_types.cpp
```diff
@@ -896,6 +897,7 @@ ConvertPostgresToBaseDuckColumnType(Form_pg_attribute &attribute) {
 		return duckdb::LogicalTypeId::UUID;
 	case JSONOID:
 	case JSONARRAYOID:
+	case JSONBOID:
```
JelteF, line 900: We should also support jsonb insida a postgres array:

```suggestion
	case JSONBOID:
	case JSONBARRAYOID:
```

This would need some tests in array_type_support.sql

test/regression/sql/json_operator_support.sql
```diff
@@ -0,0 +1,20 @@
+-- Create Table
+-- Create table for JSON testing
+CREATE TABLE test_json (
```
JelteF, line 3: Let's call this `test_jsonb`. I think it would be good to also have the same test for the actual postgres `JSON` type.

test/regression/schedule
```diff
@@ -26,6 +26,7 @@ test: transaction_errors
 test: secrets
 test: prepare
 test: function
+test: json_operator_support
```
JelteF, line 29: Let's change the name of the file now, since it's not specific to operators anymore. 
```suggestion
test: json_support
```

ritwizsinha: Removed the file and added tests to type support and array type support

src/pgduckdb_types.cpp
ritwizsinha: Done

ritwizsinha:
@JelteF your requested changes were made

JelteF (approved):
Thanks for this work. This looks good now. Next steps in my order of priority would be:

1. Start supporting some of duckdb its [JSON functions](https://duckdb.org/docs/data/json/json_functions)
2. Introduce some of the JSONB operators that postgres supports to DuckDB.
3. Introduce a PG compatibility mode in duckdb to make existing operators behave exactly the same.

PR #521, opened 2025-01-08
Use AddStringOrBlob to create to duckdb string from blob value

mkaruza:
To construct output string use`AddStringOrBlob` which doesn't require that input is valid UTF8.


Y--:
Looks reasonable - can we add a small test?

mkaruza:
> Looks reasonable - can we add a small test?

Issue that this PR is fixing will be triggered if regression suite is run on debug build.

Y-- (approved):

PR #524, opened 2025-01-10
Fix approx_count_distinct for queries without a FROM

JelteF:
I noticed two issues with the new `approx_count_distinct` implementation:

1. If no FROM clause was used it was not possible to use it
2. It would not be detected correctly as duckdb-only without
   `duckdb.force_execution = true` (or some  other mechanism). This
   fixes both of those issues.

Related to #499


src/pgduckdb_hooks.cpp
```diff
@@ -160,10 +167,22 @@ IsAllowedStatement(Query *query, bool throw_error = false) {
...
 
 	/*
-	 * If there's no rtable, we're only selecting constants. There's no point
-	 * in using DuckDB for that.
+	 * XXX: This is a bit of a weird one, because it is only allowed when
+	 * IsAllowedStatement is called with throw_error.
+	 *
+	 * If there's no rtable, we're only selecting constants. From a performance
+	 * perspective there's not really a point in using DuckDB. If we remove
+	 * this heck many common queries that are used to inspect postgres will
```
Y--, line 175:
```suggestion
	 * this hack many common queries that are used to inspect postgres will
```

src/pgduckdb_hooks.cpp
```diff
@@ -160,10 +167,22 @@ IsAllowedStatement(Query *query, bool throw_error = false) {
...
+	 *
+	 *    SELECT current_setting('work_mem');
+	 *
+	 * So even if a user sets duckdb.force_execution = true, we still won't
+	 * forward such queries to DuckDB. With one exception: If the query
+	 * requires duckdb e.g. due to a duckdb-only function being used, we'll
+	 * still executing this in DuckDB.
 	 */
-	if (!query->rtable) {
+	if (!query->rtable && !throw_error) {
```
Y--, line 185: This looks odd - if `!query->rtable` we should stop the function here (otherwise `ContainsCatalogTable` and subsequent calls would fail pretty badly no?)
Shouldn't it be something like

```
if (!query->rtable) {
   if (throw_error) {
      elog(...)

File: src/pgduckdb_hooks.cpp [hunk 3]
```diff
@@ -135,6 +142,27 @@ NeedsDuckdbExecution(Query *query) {
...
+ *
+ *    SELECT current_setting('work_mem');
+ *
+ * So even if a user sets duckdb.force_execution = true, we still won't
+ * forward such queries to DuckDB. With one exception: If the query
+ * requires duckdb e.g. due to a duckdb-only function being used, we'll
+ * still executing this in DuckDB.
+ */
+static bool
+ContainsFromClause(Query *query, bool throw_error = false) {
```
Y--, line 162: Do we really want to keep `throw_error`?

src/pgduckdb_hooks.cpp
```diff
@@ -197,7 +216,7 @@ DuckdbPlannerHook_Cpp(Query *parse, const char *query_string, int cursor_options
 			IsAllowedStatement(parse, true);
 
 			return DuckdbPlanNode(parse, query_string, cursor_options, bound_params, true);
-		} else if (duckdb_force_execution && IsAllowedStatement(parse)) {
+		} else if (duckdb_force_execution && IsAllowedStatement(parse) && ContainsFromClause(parse)) {
```
Y--, line 219: You want to move `ContainsFromClause(parse)` before `IsAllowedStatement(parse)` no? o/w we could hit the same issue calling `ContainsCatalogTable` w/ a `nullptr` no?

JelteF: It's fine to call ContainsCatalogTable and ContainsPartitionedTable with a nullptr. They both call `foreach_node` on that pointer, and the nullptr is a valid (and I think the only valid) representation of an empty list.

src/pgduckdb_hooks.cpp
JelteF: nope, copy paste mistake. Fixed now.

Y-- (approved):

PR #525, opened 2025-01-10
Various cleanups

Y--:


src/scan/postgres_table_reader.cpp
```diff
@@ -313,11 +308,12 @@ PostgresTableReader::GetNextWorkerTuple() {
 			return minimal_tuple;
 		}
 
-		next_parallel_reader++;
```
JelteF, line 316: *Bikeshedding incomming*

I'm more used to using the postfix `++` operator. And I looked at the projects we interact with a lot, and in all of those that seems to be the case too. 
```
pg_duckdb
❯ rg '\w\+\+' src | wc -l
352

pg_duckdb
❯ rg '\+\+\w' src | wc -l
7

postgres
❯ rg '\w\+\+' src | wc -l
9946

postgres
❯ rg '\+\+\w' src | wc -l
895

duckdb
❯ rg '\w\+\+' src | wc -l
3786

duckdb
❯ rg '\+\+\w' src | wc -l
495
```

So let's try to be consistent and use the postfix `++` operator as our default choice. While I don't think we necessarily have to revert or block PRs on this, I'd rather not having us refactor away from the status quo (without some reasonable reason at least).

Y--: Fair enough

JelteF (changes requested):
Looks like reasonable changes, only the `++` thing I'm not a fan of. If you feel strongly about using prefix `++` let's discuss that in a separate issue.

JelteF (approved):

PR #527, opened 2025-01-11
Add support for uhuge int

### LINKED ISSUE #280: Support UHUGEINT
This is the only integer type that we don't support reading yet, now that UBIGINT support was introduced in #275. This number should be converted into a NUMERIC in Postgres.

ritwizsinha:
Fixes #280

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -51,6 +51,7 @@ public:
 };
 
 using duckdb::hugeint_t;
+using duckdb::uhugeint_t;
 
 struct DecimalConversionInteger {
 	static int64_t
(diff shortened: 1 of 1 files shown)


JelteF (commented):
It looks good, could you add a test for this though?

ritwizsinha:
@JelteF how do I add a test for that, don't  I have to create a motherduck table with uhuge int type for this? I don't see a sample test like that

JelteF (approved):
Hmm, I don't really think there's a way we can currently create a test for this. I tried storing a `uhugeint` in a parquet file, but that becomes a float it seems because parquet does not seem to support 128bit integers.

So I tested this manually instead by indeed creating a motherduck table with a uhugeint column. That worked, so I'm merging this.

PR #543, opened 2025-01-16
[easy] [no-op] Suppress compiler warning

dentiny:
... otherwise have compiler warning
```
[4/492] Building CXX object src/planner/binder...query_node.dir/ub_duckdb_bind_query_node.cpp.osrc/pgduckdb_detoast.cpp: In function 'varlena* pgduckdb::Lz4DecompresDatum(const varlena*)':
src/pgduckdb_detoast.cpp:53:41: warning: unused parameter 'value' [-Wunused-parameter]
   53 | Lz4DecompresDatum(const struct varlena *value) {
      |                   ~~~~~~~~~~~~~~~~~~~~~~^~~~~
```


Y-- (approved):
Thanks!

PR #545, opened 2025-01-16
Start one background worker per database

Y--:


src/pgduckdb_metadata_cache.cpp
```diff
@@ -328,7 +324,7 @@ IsMotherDuckTable(Relation relation) {
 
 bool
 IsMotherDuckEnabled() {
-	return IsMotherDuckEnabledAnywhere() && IsMotherDuckPostgresDatabase();
+	return IsMotherDuckEnabledAnywhere();
```
JelteF, line 327: Maybe just make `IsMotherDuckEnabled` and `IsMotherDuckEnabledAnywhere` a single now.

src/pgduckdb_background_worker.cpp
```diff
@@ -101,17 +102,20 @@ bool CanTakeBgwLockForDatabase(Oid database_oid);
...
+
 	// Set up a signal handler for SIGTERM
 	pqsignal(SIGTERM, die);
 	BackgroundWorkerUnblockSignals();
 
-	BackgroundWorkerInitializeConnection(duckdb_motherduck_postgres_database, NULL, 0);
+	BackgroundWorkerInitializeConnectionByOid(database_oid, InvalidOid, 0);
 
 	SpinLockAcquire(&pgduckdb::BgwShmemStruct->lock);
 	pgduckdb::BgwShmemStruct->bgw_latch = MyLatch;
```
JelteF, line 121: This is problematic with multiple background workers. Only the background worker that started last will now be woken up when TriggerActivity is called.

JelteF (changes requested):

src/pgduckdb_background_worker.cpp
```diff
@@ -418,10 +416,24 @@ TriggerActivity(void) {
...
 	SpinLockAcquire(&BgwShmemStruct->lock);
 	BgwShmemStruct->activity_count++;
-	/* Force wake up the background worker */
-	if (BgwShmemStruct->bgw_latch) {
-		SetLatch(BgwShmemStruct->bgw_latch);
+	/* Iterate over all background workers and notify them */
+	const uint32_t numProcs = ProcGlobal->allProcCount;
+	for (uint32_t i = 0; i < numProcs; i++) {
+		PGPROC *proc = &ProcGlobal->allProcs[i];
+		if (!proc->isBackgroundWorker || proc->databaseId == InvalidOid) {
```
JelteF, line 423: As discussed in person, this loop is quite. Let's see if we can use a shared memory hash table instead. You can initialize it with `ShmemInitHash`. As maximum size we can probably use `max_worker_processes`, maybe with a little bit of extra buffer.

src/pgduckdb_background_worker.cpp
```diff
@@ -280,9 +312,11 @@ ShmemStartup(void) {
...
 		MemSet(BgwShmemStruct, 0, size);
 		SpinLockInit(&BgwShmemStruct->lock);
-		BgwShmemStruct->bgw_latch = nullptr;
-		BgwShmemStruct->activity_count = 0;
-		BgwShmemStruct->bgw_session_hint_is_reused = false;
+
+		HASHCTL info;
+		info.keysize = sizeof(Oid);
+		info.entrysize = sizeof(BgwStatePerDB);
+		BgwShmemStruct->statePerDB = ShmemInitHash("ProcBgwStatePerDB", 16, 2048, &info, HASH_ELEM | HASH_BLOBS);
```
Y--, line 319: Curious to hear your thoughts on these defaults? (16=initial item #, 2048=maximum item #)

src/pgduckdb_background_worker.cpp
```diff
@@ -417,11 +451,9 @@ TriggerActivity(void) {
 	}
 
 	SpinLockAcquire(&BgwShmemStruct->lock);
-	BgwShmemStruct->activity_count++;
-	/* Force wake up the background worker */
-	if (BgwShmemStruct->bgw_latch) {
```
Y--, line 422: This `if` meant that somehow we were notifying changes before the bgw was available.
This should not happen anymore.

src/pgduckdb_background_worker.cpp
JelteF: I think as a maximum `max_worker_processes` should be enough for a maximum size, 2048 is a bit arbitrary (and usually excessive). Lowering it to `max_worker_processes` has one issue though currently: we don't remove entries from the hashmap when a background worker dies. I think we should do that with PG_TRY + PG_FINALLY. Also the `// Unreachable` comment is just plain wrong, we have a break in the loop now.

Y--: `max_worker_processes` is a good idea. I also had planned to work on the unregister, but it shouldn't be in our way for the number of items here since it is indexed per database `Oid` - so if one dies there's hopefully one that will come in its place (or the slot will be wasted, but there won't be a continuous increase)

src/pgduckdb_background_worker.cpp
```diff
@@ -280,9 +315,12 @@ ShmemStartup(void) {
...
 		MemSet(BgwShmemStruct, 0, size);
 		SpinLockInit(&BgwShmemStruct->lock);
-		BgwShmemStruct->bgw_latch = nullptr;
-		BgwShmemStruct->activity_count = 0;
-		BgwShmemStruct->bgw_session_hint_is_reused = false;
+
+		HASHCTL info;
+		info.keysize = sizeof(Oid);
+		info.entrysize = sizeof(BgwStatePerDB);
+		auto max_entries = max_worker_processes > 16 ? max_worker_processes : 16;
```
JelteF, line 322: This can use `std::max` (C++) or `Max` (Postgres C) 
```suggestion
		auto max_entries = max_worker_processes > 16 ? max_worker_processes : 16;
```

JelteF (approved):

Issue #516, opened 2025-01-02
Support JSONB


JelteF:
### Description

We currently only support the Postgres `json` type, not the `jsonb` type. The `jsonb` type is much more common in practice in Postgres though, so we should support it. Related to #513.

ritwizsinha:
Taking this up

ritwizsinha:
Next in line
> Next steps in my order of priority would be:
> 
> 1. Start supporting some of duckdb its [JSON functions](https://duckdb.org/docs/data/json/json_functions)
> 2. Introduce some of the JSONB operators that postgres supports to DuckDB.
> 3. Introduce a PG compatibility mode in duckdb to make existing operators behave exactly the same.

JelteF:
Item 1 was also requested now in #533

jp-agenta:
Hey, hi !
First of all, kudos on `pg_duckdb`.
Second, `JSONB` support is a must for us and any large-scale-observability-on-postgres-with-duck-db kind of work.
We'll be waiting for this to be resolved before we move our GB/TBs to `pg_duckdb`.

JelteF:
can you clarify what specific things you're currently missing? You can read jsonb columns just fine, you'll simply need to use the duckdb functions instead of the Postgres jsonb functions to extract data from it.

PR #546, opened 2025-01-17
Added support for json functions

ritwizsinha:
Addresses #516


ritwizsinha:
@JelteF I have added all the json functions present in [duckdb](https://duckdb.org/docs/data/json/json_functions#json-scalar-functions) to the sql file, is this fine or do we need to break this up. All of this is quite repeated I doubt a big pr would be a problem

sql/pg_duckdb--0.2.0--0.3.0.sql
```diff
@@ -17,3 +17,227 @@ CREATE AGGREGATE @extschema@.approx_count_distinct(anyelement)
 
 CREATE DOMAIN pg_catalog.blob AS bytea;
 COMMENT ON DOMAIN pg_catalog.blob IS 'The DuckDB BLOB alias for BYTEA';
+
+-- json_exists
+CREATE FUNCTION @extschema@.json_exists(json VARCHAR , path VARCHAR)
```
JelteF, line 22: It seems confusing that these functions expect a `VARCHAR` type for the `json` argument. I'd expect that argument to be of the `json` or `jsonb` (or have functions for both of them).

src/pgduckdb_metadata_cache.cpp
```diff
@@ -109,8 +109,9 @@ BuildDuckdbOnlyFunctions() {
 	 * each of the found functions is actually part of our extension before
 	 * caching its OID as a DuckDB-only function.
 	 */
-	const char *function_names[] = {"read_parquet",      "read_csv",   "iceberg_scan", "iceberg_metadata",
-	                                "iceberg_snapshots", "delta_scan", "read_json",    "approx_count_distinct"};
+	const char *function_names[] = {"read_parquet",      "read_csv",     "iceberg_scan",       "iceberg_metadata",
+	                                "iceberg_snapshots", "delta_scan",   "read_json",          "approx_count_distinct",
+	                                "json_exists",       "json_extract", "json_extract_string"};
```
JelteF, line 114: It seems this list is missing a bunch of the functions that you added.

test/regression/sql/json_functions_duckdb.sql
```diff
@@ -0,0 +1,144 @@
+-- <JSON_EXISTS>
+-- Test 1: Path exists in a simple JSON object
+SELECT json_exists('{"a": {"b": 1}}', '$.a.b'); -- Expected: true
```
JelteF, line 3: I think most of these tests simply test that DuckDB works correctly, not so much the pg_duckdb intgration.

I think for each function it should give us enough confidence into pg_duckdb to add one test with reasonable arguments, and then tests with the various combinations of `NULL` arguments. So while I really appreciate the effort to add extensive tests, I don't think they add much value and would in practice only increase test runtime.

JelteF (changes requested):

test/regression/sql/json_functions_duckdb.sql
ritwizsinha: Reduced the tests

src/pgduckdb_metadata_cache.cpp
ritwizsinha: Added

ritwizsinha:
@JelteF structures are not supported as types in pg_duckdb so not possible to implement functions like:
```
json_transform(json, structure)	Transform json according to the specified structure.
from_json(json, structure)	Alias for json_transform.
json_transform_strict(json, structure)	Same as json_transform, but throws an error when type casting fails.
from_json_strict(json, structure)	Alias for json_transform_strict.
```

Added all the other ones

ritwizsinha:
Curious why this fails on postgres 17

sql/pg_duckdb--0.2.0--0.3.0.sql
```diff
@@ -17,3 +17,226 @@ CREATE AGGREGATE @extschema@.approx_count_distinct(anyelement)
...
 COMMENT ON DOMAIN pg_catalog.blob IS 'The DuckDB BLOB alias for BYTEA';
+
+-- json_exists
+CREATE FUNCTION @extschema@.json_exists("json" json , path VARCHAR)
+RETURNS boolean LANGUAGE 'plpgsql'
+SET search_path = pg_catalog, pg_temp
+AS
+$func$
+BEGIN
+    RAISE EXCEPTION 'Function `jsonb_exists(json, VARCHAR)` only works with Duckdb execution.';
```
JelteF, line 28: I just merged #531. Which introduces a more re-usable way of doing this:

```
LANGUAGE C;
```

ritwizsinha: Fixed

JelteF (approved):
I played around with this a bunch and added a few more tests and fixed a bunch of issues I found. I now think this is good enough to merge. @Y-- @mkaruza can either of you do another small review.

Y-- (approved):

PR #548, opened 2025-01-20
Duckdb 1.2 update

mkaruza:


mkaruza:
Preparation for DuckDB 1.2 release, should be merged once version is officially released with extensions

src/scan/postgres_scan.cpp
```diff
@@ -157,12 +182,12 @@ PostgresScanTableFunction::PostgresScanTableFunction()
...
-PostgresScanTableFunction::ToString(const duckdb::FunctionData *data) {
-	auto &bind_data = data->Cast<PostgresScanFunctionData>();
-	std::ostringstream oss;
-	oss << "(POSTGRES_SCAN) " << GetRelationName(bind_data.rel);
-	return oss.str();
+duckdb::InsertionOrderPreservingMap<duckdb::string>
+PostgresScanTableFunction::ToString(duckdb::TableFunctionToStringInput &input) {
+	auto &bind_data = input.bind_data->Cast<PostgresScanFunctionData>();
+	duckdb::InsertionOrderPreservingMap<duckdb::string> result;
+	result["Table"] = GetRelationName(bind_data.rel);
```
Y--, line 189: Maybe interesting to note that it is a PG table?
```suggestion
	result["Table"] = GetRelationName(bind_data.rel) + " (PG)";
```

src/scan/postgres_scan.cpp
```diff
@@ -13,6 +15,29 @@ namespace pgduckdb {
...
+		scan_query << filter->ToString(column_name).c_str();
+		break;
+	case duckdb::TableFilterType::OPTIONAL_FILTER: {
+		auto optional_filter = reinterpret_cast<duckdb::OptionalFilter *>(filter);
+		ConstructQueryFilter(optional_filter->child_filter.get(), column_name);
+		break;
+	}
+	case duckdb::TableFilterType::STRUCT_EXTRACT:
+	case duckdb::TableFilterType::DYNAMIC_FILTER:
+		scan_query << "1 = 1";
```
JelteF, line 36: This seems like a bit of a hack. Do we really need this? If so, can you add a column explaining why?

Also, is this actually correct? Because now we don't actually filter anything for these filter types inside postgres, but duckdb will not filter on that side either. I feel like we probably need some tests that show that we return correct result when using these kind of things.

mkaruza: Updated construction of pushdown filters. Looking at duckdb code `STRUCT_EXTRACT` is only created if `struct_extract` is used in query, while `DYNAMIC_FILTER` is constructed when there is topN node (LIMIT). Both of these should not matter to filtering.

src/scan/postgres_scan.cpp
```diff
@@ -7,12 +9,65 @@
...
+		}
+		return conjuction_child_filters.size();
+	}
+	case duckdb::TableFilterType::OPTIONAL_FILTER: {
+		auto optional_filter = reinterpret_cast<duckdb::OptionalFilter *>(filter);
+		return ExtractQueryFilters(optional_filter->child_filter.get(), column_name, query_filters);
+	}
+	/* DYNAMIC_FILTER is push down filter from topN execution */
+	case duckdb::TableFilterType::DYNAMIC_FILTER:
+		return 0;
```
JelteF, line 60: Why do we return 0 here instead of throwing an error? Do we have a query that triggers this?

mkaruza: DYNAMIC_QUERY happens when there is `LIMIT` in query so we'll see this filter pushed down to scan, but we can't do anything about it except to skip it.

src/scan/postgres_scan.cpp
mkaruza: Explain already show it is `POSTGRES_SCAN` node. We can add it if needed.

src/scan/postgres_scan.cpp
JelteF: Alright, I looked a bit closer at the code and understand much better what all these things are: I think a somewhat cleaner/future-proof way of handling this is by remembering if we're in an optional filter or not (probably by passing it as a third argument to the recursive function). Then for all these filters that we don't support, we do nothing if it's inside an optional filter, or throw an error if it's not inside an optional filter.

src/scan/postgres_scan.cpp
```diff
@@ -7,12 +9,71 @@
...
+		}
+		return conjuction_child_filters.size();
+	}
+	case duckdb::TableFilterType::OPTIONAL_FILTER: {
+		auto optional_filter = reinterpret_cast<duckdb::OptionalFilter *>(filter);
+		return ExtractQueryFilters(optional_filter->child_filter.get(), column_name, query_filters, true);
+	}
+	/* DYNAMIC_FILTER is push down filter from topN execution */
+	case duckdb::TableFilterType::DYNAMIC_FILTER:
+		return 0;
```
JelteF, line 60: Now it should be possible to remove this line right?
```suggestion
```

src/scan/postgres_scan.cpp
```diff
@@ -7,12 +9,70 @@
...
 
+static duckdb::string
+FilterJoin(duckdb::vector<duckdb::string> &filters, duckdb::string &&delimiter) {
+	return std::accumulate(filters.begin() + 1, filters.end(), filters[0],
+	                       [&delimiter](duckdb::string l, duckdb::string r) { return l + delimiter + r; });
+}
+
+int
+PostgresScanGlobalState::ExtractQueryFilters(duckdb::TableFilter *filter, const char *column_name,
+                                             duckdb::string &query_filters, bool is_optional_filter_parent) {
```
JelteF, line 28: nit: "parent" seems like the wrong naming imo. Maybe one of these names instead:
```suggestion
                                             duckdb::string &query_filters, bool is_optional_filter_child) {
```

```suggestion
                                             duckdb::string &query_filters, bool is_inside_optional_filter) {
```

Y-- (approved):

PR #550, opened 2025-01-20
Create dedicated cache FS

Y--:
Today, non super-users are going to face an error when querying S3 objects like:

```
ERROR:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Permission Error: File system LocalFileSystem has been disabled by configuration
```

The reason is that the current cache mechanism uses the DB's `FileSystem` which non-super users cannot access.

To fix it, we create a dedicated FS that the extension is allow to access (but not the user directly).


third_party/cached_httpfs/include/http_file_cache.hpp
```diff
@@ -78,6 +80,13 @@ class CachedFileHandle {
 	shared_ptr<CachedFile> file;
 };
 
+class LocalCacheFileSystem: public LocalFileSystem {
+	// TODO: we could lock down the LocalFileSystem to only allow path that are in the cache directory
```
JelteF, line 84: Would be nice to implement this indeed as a bit of defense in depth. Seems like it should only be necessary to override `OpenFile`, with one that first resolves symlinks and `../..` and then checks the prefix.

JelteF (commented):
Tried it out. It indeed solves the problem. Also the general approach seems quite reasonable.

third_party/cached_httpfs/include/http_file_cache.hpp
Y--: Yeah definitely.

mkaruza (approved):

third_party/cached_httpfs/http_file_cache.cpp
```diff
@@ -90,14 +81,31 @@ void CachedFileHandle::Read(void *buffer, idx_t length, idx_t offset) {
...
 }
 
+LocalCacheFileSystem& HTTPFileCache::GetFS(const string &cache_dir) {
+	if (cached_fs) {
+		cached_fs->AssertSameCacheDir(cache_dir);
+		return *cached_fs;
+	}
+
+	{
+		lock_guard<mutex> lock(cached_fs_mutex);
```
Y--, line 91: This is probably a bit overkill since `GetFS` is only called under a lock already, but feels more future-proof

third_party/cached_httpfs/include/http_file_cache.hpp
```diff
@@ -1,18 +1,84 @@
...
+			throw PermissionException("Must provide a file name, not a path. Got: '", file_name, "'");
+
+		}
+
+		std::ostringstream oss;
+		oss << cache_file_directory << "/" << file_name;
+		return LocalFileSystem::OpenFile(oss.str(), flags, opener);
+	}
+
+	void AssertSameCacheDir(const string &other_dir) {
```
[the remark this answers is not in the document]
Y--: But then I have to make the `cache_file_directory` public just for this method? I preferred to not leak it :-)

third_party/cached_httpfs/include/http_file_cache.hpp
JelteF, line 84: let's remove this comment

JelteF (approved):

Y--:
For posterity - we've discussed [`24891f4` (#550)](https://github.com/duckdb/pg_duckdb/pull/550/commits/24891f42aab50f8a703c669a95085198711958e5) with @JelteF and decided it wasn't worth it.

PR #553, opened 2025-01-23
Remove custom pushdown filtering files

mkaruza:
We don't do any filtering inside pg_duckdb. That is now handled by Postgres execution. So remove files that are not needed.


Y-- (approved):
Woot!

PR #559, opened 2025-01-26
Add implicit % to bridge TABLESAMPLE syntax

### LINKED ISSUE #558: Feature Request: Support TABLESAMPLE
### Discussed in https://github.com/duckdb/pg_duckdb/discussions/429

<div type='discussions-op-text'>

<sup>Originally posted by **dpxcc** November 12, 2024</sup>
The `TABLESAMPLE` clause is useful for sampling from large tables, but pg_duckdb doesn't support it right now due to syntax differences in `TABLESAMPLE` clause between Postgres and DuckDB: https://duckdb.org/docs/sql/samples.html
```sql
SET duckdb.force_execution = ON;
CREATE TABLE t (a int);
SELECT * FROM t TABLESAMPLE SYSTEM (10);
```
throws
```
WARNING:  (PGDuckDB/CreatePlan) Prepared query returned an error: 'Parser Error: Sample method System cannot be used with a discrete sample count, either switch to reservoir sampling or use a sample_size
 a 
---
(0 rows)
```
but it could be an error if we added some DuckDB-only function to the query</div>

nbiscaro:
Bridge syntax differences in `TABLESAMPLE` by adding implicit `%`. 

Fixes #558 
Fixes #429


src/vendor/pg_ruleutils_17.c
```diff
@@ -12493,26 +12493,33 @@ static void
...
+	appendStringInfo(buf, " TABLESAMPLE %s (", tsm_name);
 
-	nargs = 0;
-	foreach(l, tablesample->args)
+	if (pg_strcasecmp(tsm_name, "system") == 0 && list_length(tablesample->args) == 1)
 	{
-		if (nargs++ > 0)
-			appendStringInfoString(buf, ", ");
-		get_rule_expr((Node *) lfirst(l), context, false);
+		get_rule_expr((Node *) linitial(tablesample->args), context, false);
```
JelteF, line 12511: These changes also need to be made in the other `pg_ruleutils_1x.c` files. And probably we want to share some of this code to `pgduckdb_ruleutils.cpp` so the code can be shared between these files.

JelteF (changes requested):
Thank you for the PR! This needs some tests though.

src/vendor/pg_ruleutils_14.c
```diff
@@ -11424,26 +11424,33 @@ static void
...
+	}
+	else 
+	{
+		nargs = 0;
+		foreach(l, tablesample->args)
+		{
+			if (nargs++ > 0)
+				appendStringInfoString(buf, ", ");
+			get_rule_expr((Node *) lfirst(l), context, false);
+		}
```
JelteF, line 11453: This code is currently added to each ruleutils file. We try to keep modifications to those files to a minimum to keep clear what is our code and what code we vendored in. I think the easiest way of doing that here would be to remove the whole `get_tablesample_def` function to `pgduckdb_ruleutils.cpp` and name it there `pgduckdb_get_tablesample_def`. And then add a an extra rename macro to `include/pgduckdb/utility/rename_ruleutils.h`, like we already do for `generate_qualified_relation_name`. 

```c
#define get_tablesample_def          pgduckdb_get_tablesample_def
```

test/regression/expected/tablesample.out
```diff
@@ -0,0 +1,19 @@
+SET duckdb.force_execution = true;
+CREATE TABLE t (a INT);
+INSERT INTO t SELECT i FROM generate_series(1, 10) AS i;
+SELECT * FROM t TABLESAMPLE SYSTEM (100);
```
JelteF, line 4: I think we should expand the testing a bit.

1. We should have a test with `bernoulli` as the method
2. We should add a repeatable.

So something like this:
```
SELECT * from t TABLESAMPLE bernoulli(100);
SELECT * from t TABLESAMPLE bernoulli(50) repeatable(42);
```

JelteF:
@nbiscaro do you still intend to address these comments?

nbiscaro:
> @nbiscaro do you still intend to address these comments?

Yes! Will get to this within the next 24 hours.

test/regression/expected/tablesample.out
nbiscaro: Currently we do an explicit check for SYSTEM. https://github.com/duckdb/pg_duckdb/pull/559/files#diff-df1642e6c5dec889185ca370bddf9f15783a7fa26793de986f6d4aec605b899fR11851

Do we also want to add this for `bernoulli` and `reservoir`?

src/vendor/pg_ruleutils_14.c
JelteF: Hmm, yeah you're right, I didn't realize that. In that case I think it would be good to try and make the the diff as small as it can be. I think the smallest diff we can probably do is completely keep the original implementation of this function, but add 1 additional function call:

```c
		foreach(l, tablesample->args)
		{
			if (nargs++ > 0)
				appendStringInfoString(buf, ", ");
			get_rule_expr((Node *) lfirst(l), context, false);
			// this call is to be added:
			pgduckdb_add_tablesample_percent(buf, tablesample);
		}
```

JelteF:
@nbiscaro any progress?

src/vendor/pg_ruleutils_14.c
nbiscaro: We need to resolve tsm_name from within each ruleutils file using `generate_function_name` which has a vendor specific definition. I don't think we can pass tablesample directly. I think the best we can do is:

```cpp
  const char *tsm_name = generate_function_name(tablesample->tsmhandler, 1,
									  NIL, argtypes,
									  false, NULL, EXPR_KIND_NONE);
  pgduckdb_add_tablesample_percent(tsm_name, buf, list_length(tablesample->args));
```

test/regression/expected/tablesample.out
nbiscaro: Sounds good, added some testing for bernoulli.

JelteF (approved):

JelteF (approved):

PR #563, opened 2025-01-28
Fix case insensitivity issues

### LINKED ISSUE #116: Case insensitivity mismatch between duckdb and postgres can cause issues

JelteF:
Without this change it's impossible to reference two Postgres tables in the same query that only differ in casing. This fixes that by not using a `case_insentive_map` to track the Postgres table and schema names. While this doesn't completely fix all issues caused by case insentivity mismatch between Postgres and DuckDB it does fix the major one described in #116.

Fixes #116

### DIFF
--- a/include/pgduckdb/catalog/pgduckdb_transaction.hpp
+++ b/include/pgduckdb/catalog/pgduckdb_transaction.hpp
@@ -2,6 +2,7 @@
 
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/main/client_context_state.hpp"
+#include "duckdb/common/unordered_map.hpp"
 #include "duckdb/transaction/transaction.hpp"
 #include "pgduckdb/pg/declarations.hpp"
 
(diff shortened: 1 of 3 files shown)


Y-- (commented):
Can we add a test?

JelteF:
Ugh yes... I forgot the `git add`...

Y--:
> Ugh yes... I forgot the `git add`...

Git is hard 😆

Y-- (approved):
LGTM! Thanks

Issue #564, opened 2025-01-28
Cannot change case using output name (`as`) of an upper case column


CaselIT:
Not sure if it's related, but it seems you can't change the casing of the names with as if the name is upper when reading from parquet.

create a parquet, like
```bash
duckdb -c "COPY (select generate_series as \"UPPER\", generate_series as \"lower\" FROM generate_series(100)) TO '/tmp/tmp.parquet' (FORMAT 'parquet')"
```

then from the postgresql
```sql
select * from read_parquet('/tmp/tmp.parquet') as t(upper int8, lower int8) limit 2;

select upper as "upper", lower as "LOWER" from read_parquet('/tmp/tmp.parquet') as t(upper int8, lower int8) limit 2;

select upper as "upper_", lower as "LOWER_" from read_parquet('/tmp/tmp.parquet') as t(upper int8, lower int8) limit 2;
```
output of the above queries is
```
 UPPER | lower
-------+-------
     0 |     0
     1 |     1
(2 rows)

 UPPER | LOWER
-------+-------
     0 |     0
     1 |     1
(2 rows)

 upper_ | LOWER_
--------+--------
      0 |      0
      1 |      1
(2 rows)
```

can't rename `UPPER` to lowercase. If you change the name it works fine.

interestingly it only happens UPPERCASE to lowercase, not the opposite as shown in the lower column

let me know if I should open a new issue 

 _Originally posted by @CaselIT in [#116](https://github.com/duckdb/pg_duckdb/issues/116#issuecomment-2619643360)_

JelteF:
Can you update the example to include ones with the new syntax? That makes copy pasting easier when debugging.

CaselIT:
not sure I understand what you are asking, but I've split the sql from the output

JelteF:
See #565 with what I meant with the "new syntax", i.e no `AS t(upper int8, lower int8)` is necessary anymore.

CaselIT:
oh nice change. I didn't know about it, since I did not find it yet mentioned in the readme

JelteF:
Yeah docs+readme updates are planned soon

JelteF:
I'm curious how your old syntax worked with the new docker images though. Because those should fail now.

CaselIT:
I'm just using whatever is in `docker.io/pgduckdb/pgduckdb:17-main`

```
scott=# \dx
                  List of installed extensions
   Name    | Version |   Schema   |         Description
[...]

PR #570, opened 2025-01-31
Try fixing docker

JelteF:
Our docker build has been broken for ~a week. Noticed because the user in #564
was still getting images the old read_parquet syntax.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -18,7 +18,8 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        postgres: ["14", "15", "16", "17"]
+        # postgres: ["14", "15", "16", "17"]
+        postgres: ["14"]
 
     steps:
       - name: Login to Docker Hub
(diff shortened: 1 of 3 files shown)


wuputah:
This is curious. I'm not seeing an error in the logs, and I wasn't able to reproduce locally (on `main`) with Postgres 17 and arm64. I'm going to try an amd64 build next. This is the command I used. 

```
POSTGRES_VERSION=17 docker buildx bake --set '*.platform=linux/arm64' --load postgres
```

I wonder if this the build hitting a resource constraint of the worker.

In mostly unrelated news, public arm64 workers are here, though how to get the build to properly split across platforms and push to Dockerhub correctly is not trivial based on our previous attempts.
https://github.com/orgs/community/discussions/148648

wuputah:
one last idea, could try disabling ninja e.g. with `DUCKDB_GEN=make make -j$(nproc)`. I had to do that for pgxman, though I don't remember the error that resulted in me doing so.

JelteF:
Closing this in favor of #578

PR #572, opened 2025-02-03
Update security restrictions to allow non-superuser extension installation

JelteF:
This updates a bunch of our security related code. For previous releases we needed to be very careful with allowing arbitrary SQL code to be executed in DuckDB because DuckDB queries could read all Postgres tables. This is not the case anymore since #466 was merged, because now any access to Postgres tables goes through the Postgres planner and executor instead of custom code. Lots of code wasn't updated with that new behaviour in mind though.

1. Allow running `duckdb.raw_query`, `duckdb.cache`, `duckdb.cache_info`, `duckdb.cache_delete` and `duckdb.recycle_db` as any user (with the duckdb role).
2. Allow running `duckdb.install_extension` as regular users, if required permissions are explicitly granted. This is not allowed by default for non-superusers because it's still considered a very high privilege.
3. Disallow running queries on tables with RLS enabled in a different place, so that it is checked for every Postgres table that DuckDB opens (also when using `duckdb.query`/`duckdb.raw_query`).
4. Add `duckdb.allow_community_extensions` setting.


Y-- (approved):
LGTM!

sql/pg_duckdb--0.2.0--0.3.0.sql
```diff
@@ -824,3 +824,6 @@ RETURNS SETOF duckdb.row
 SET search_path = pg_catalog, pg_temp
 AS 'MODULE_PATHNAME', 'duckdb_only_function'
 LANGUAGE C;
+
+GRANT ALL ON FUNCTION duckdb.raw_query(TEXT) TO PUBLIC;
+GRANT ALL ON FUNCTION duckdb.cache(TEXT, TEXT) TO PUBLIC;
```
Bodobolero, line 829: in consequence, would it make sense to also add

```sql
GRANT ALL ON FUNCTION duckdb.cache_info() TO PUBLIC;
GRANT ALL ON FUNCTION duckdb.cache_delete(cache_key TEXT) TO PUBLIC;
```

For this one I am undecided:
```sql
GRANT ALL ON PROCEDURE duckdb.recycle_ddb() TO PUBLIC;
```

and as you correctly decided we should NOT (!)

```sql
GRANT ALL ON FUNCTION duckdb.install_extension(TEXT) TO PUBLIC;
```
because that implies a security risk
?

src/pgduckdb_options.cpp
```diff
@@ -193,8 +194,17 @@ ReadDuckdbExtensions() {
...
-	auto install_extension_command = duckdb::StringUtil::Format("INSTALL %s;", extension_name.c_str());
+
+	auto install_extension_command = "INSTALL " + duckdb::KeywordHelper::WriteQuoted(extension_name);
+
+	/*
+	 * Temporily allow all filesystems for this query, because INSTALL needs
+	 * local filesystem access.
+	 */
+	auto save_nestlevel = NewGUCNestLevel();
+	SetConfigOption("duckdb.disabled_filesystems", "", PGC_SUSET, PGC_S_SESSION);
```
Bodobolero, line 205: I think I read some code comment or documentation that this GUC (disabled_filesystems) can only be disabled once but not re-enabled once it has been disabled.
I didn't test if it works or not. 
Did you test that you could enable LocalFileSystem temporarily as planned AFTEr it had been previously disabled for the session?

Your non_superuser.out below shows that you did, which is great. Just wanted to verify that you really tested with "disabled_filesystems = LocalFileSystem" ?

Bodobolero (approved):
LGTM but added some suggestions and questions

src/pgduckdb_options.cpp
Bodobolero: what happens if a user uses another session at the same time while extension install is running (and GUC is set to disabled_filesystem = ""). Could the other session access LocalFilesystem concurrently - thus having an attack vector?

sql/pg_duckdb--0.2.0--0.3.0.sql
JelteF: I updated the code to also make it possible to call all other three safely.

src/pgduckdb_options.cpp
JelteF: Updated the code comment to address these questions.

Bodobolero: thanks, makes sense now

PR #575, opened 2025-02-05
Release 0.3.0

JelteF:
Updates the docs, changelog and README for the 0.3.0 release.

PR #577, opened 2025-02-05
Add test that runs DuckDB its test_all_types function

JelteF:
This is an easy way to test conversion from Postgres to DuckDB for types
that don't have an equivalent in Postgres.

It also gives us an easy todo list of types that we don't support yet.

### DIFF
--- a/test/regression/expected/test_all_types.out
+++ b/test/regression/expected/test_all_types.out
@@ -1,0 +1,127 @@
+\x on
+SET bytea_output = 'escape';
+SELECT * FROM duckdb.query($$
+FROM test_all_types()
+SELECT * exclude(
+    tinyint, -- PG14 outputs this differently currently
+    varint,
+    TIME,
+    time_tz,
+    bit,
+    small_enum,
+    medium_enum,
+    large_enum,
+    timestamptz_array,
+    struct,
+    struct_of_arrays,
+    array_of_structs,
+    map,
+    "union",
+    fixed_int_array,
+    fixed_varchar_array,
+    fixed_nested_int_array,
+    fixed_nested_varchar_array,
+    fixed_struct_array,
+    struct_of_fixed_array,
+    fixed_array_of_int_list,
+    list_of_fixed_int_array,
+    nested_int_array, -- The nested array has different lengths, which is not possible in PG
+)
+$$)
+-[ RECORD 1 ]---+-------------------------------------------------------------------------------------------------------------------------------------
+bool            | f
+smallint        | -32768
+int             | -2147483648
+bigint          | -9223372036854775808
+hugeint         | 0
+uhugeint        | 0
+utinyint        | 0
+usmallint       | 0
+uint            | 0
+ubigint         | 0
+date            | 07-14-5881580
+timestamp       | Sun Jan 10 08:01:49.551616 294247
+timestamp_s     | Sun Jan 10 08:01:49.551616 294247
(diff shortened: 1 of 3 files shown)


test/regression/expected/test_all_types.out
```diff
@@ -0,0 +1,129 @@
...
+    nested_int_array, -- The nested array has different lengths, which is not possible in PG
+)
+$$)
+-[ RECORD 1 ]---+-------------------------------------------------------------------------------------------------------------------------------------
+bool            | f
+tinyint         | \200
+smallint        | -32768
+int             | -2147483648
+bigint          | -9223372036854775808
+hugeint         | 0
```
JelteF, line 36: This is a bug! We should be returning:
```suggestion
hugeint         | -170141183460469231731687303715884105728
```

test/regression/expected/test_all_types.out
```diff
@@ -0,0 +1,129 @@
...
+    fixed_struct_array,
+    struct_of_fixed_array,
+    fixed_array_of_int_list,
+    list_of_fixed_int_array,
+    nested_int_array, -- The nested array has different lengths, which is not possible in PG
+)
+$$)
+-[ RECORD 1 ]---+-------------------------------------------------------------------------------------------------------------------------------------
+bool            | f
+tinyint         | \200
```
JelteF, line 32: This is displayed pretty weirdly. Maybe we should not use the `char` type for this, but instead upscale to a regular `int`. For now I added this as a limitation to the #575 PR.

test/regression/expected/test_all_types.out
```diff
@@ -0,0 +1,129 @@
...
+utinyint        | 0
+usmallint       | 0
+uint            | 0
+ubigint         | 0
+date            | 07-14-5881580
+timestamp       | Sun Jan 10 08:01:49.551616 294247
+timestamp_s     | Sun Jan 10 08:01:49.551616 294247
+timestamp_ms    | Sun Jan 10 08:01:49.551616 294247
+timestamp_ns    | Wed Sep 22 00:00:00 1677
+timestamp_tz    | Sun Jan 10 00:01:49.551616 294247 PST
```
JelteF, line 47: All of these overflow currently. I think we should either truncate or throw an error in this case.

test/regression/expected/test_all_types.out
```diff
@@ -0,0 +1,129 @@
...
+dec_9_4         | 99999.9999
+dec_18_6        | 999999999999.999999
+dec38_10        | 9999999999999999999999999999.9999999999
+uuid            | ffffffff-ffff-ffff-ffff-ffffffffffff
+interval        | @ 83 years 3 mons 999 days 16 mins 39.999999 secs
+varchar         | goo
+blob            | \000\000\000a
+int_array       | {42,999,NULL,NULL,-42}
+double_array    | {42,NaN,Infinity,-Infinity,NULL,-42}
+date_array      | {01-01-1970,07-11-5881580,07-13-5881580,NULL,05-12-2022}
```
JelteF, line 93: the two dates in the middle are displayed by duckdb as `infinity` and `-infinity`

JelteF: Turns out `infinity` and `-infinity` are valid values for dates/timestamps in both Postgres and DuckDB. But based on this our conversion of these is lacking.

Y-- (approved):

Y-- (approved):

PR #582, opened 2025-02-07
Don't sync MotherDuck metadata forever

JelteF:
Keeping a connection open to MotherDuck forever is undesirable from a resourcing perspective. This changes that to behave similarly to the DuckDB CLI: After 5 minutes (by default) syncing will stop. The main difficulty is to restart the syncing. In the DuckDB CLI the syncing will start again after some activity is detected, but the background worker never triggers activity. So instead we need to make sure that activity in other connections triggers a restart of the syncing in the background worker. This is done using some very simple IPC in shared memory.

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -20,6 +20,7 @@ int duckdb_motherduck_enabled = MotherDuckEnabled::MOTHERDUCK_AUTO;
 char *duckdb_motherduck_token = strdup("");
 char *duckdb_motherduck_postgres_database = strdup("postgres");
 char *duckdb_motherduck_default_database = strdup("");
+char *duckdb_motherduck_background_catalog_refresh_inactivity_timeout = strdup("5 minutes");
 char *duckdb_postgres_role = strdup("");
 
 int duckdb_maximum_threads = -1;
@@ -44,7 +45,7 @@ _PG_init(void) {
 	DuckdbInitGUC();
 	DuckdbInitHooks();
 	DuckdbInitNode();
-	DuckdbInitBackgroundWorker();
+	pgduckdb::InitBackgroundWorker();
 	pgduckdb::RegisterDuckdbXactCallback();
 }
 } // extern "C"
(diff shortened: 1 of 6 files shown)


src/pgduckdb_background_worker.cpp
```diff
@@ -546,30 +656,25 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
...
-
-void
-SyncMotherDuckCatalogsWithPg(bool drop_with_cascade) {
-	InvokeCPPFunc(SyncMotherDuckCatalogsWithPg_Cpp, drop_with_cascade);
+static void
+SyncMotherDuckCatalogsWithPg(bool drop_with_cascade, duckdb::ClientContext &context) {
+	/*
+	 * TODO: Passing a reference through InvokeCPPFunc doesn't work here
+	 * for some reason. So to work around that we use a pointer instead.
+	 * We should fix the underlying problem instead.
```
JelteF, line 664: @Y-- maybe something for your template magic to fix.

Y--: I think I've seen something like that in the past. Never got a chance to investigate though... Let's open an issue and I'll try to look at it soon?

src/pgduckdb.cpp
```diff
@@ -20,6 +20,7 @@ int duckdb_motherduck_enabled = MotherDuckEnabled::MOTHERDUCK_AUTO;
 char *duckdb_motherduck_token = strdup("");
 char *duckdb_motherduck_postgres_database = strdup("postgres");
 char *duckdb_motherduck_default_database = strdup("");
+char *duckdb_motherduck_background_catalog_refresh_inactivity_timeout = strdup("5 minutes");
```
Y--, line 23: Can we not set the default? (and keep default setting on MD side?)

src/pgduckdb_hooks.cpp
```diff
@@ -353,6 +356,18 @@ DuckdbExplainOneQueryHook(Query *query, int cursorOptions, IntoClause *into, Exp
...
+static bool
+IsOutdatedMotherduckCatalogErrcode(int error_code) {
+	switch (error_code) {
+	case ERRCODE_UNDEFINED_COLUMN:
+	case ERRCODE_UNDEFINED_SCHEMA:
+	case ERRCODE_UNDEFINED_TABLE:
+		return true;
+	default:
+		return false;
+	}
```
Y--, line 368: nit- let's use a non-branching version?
```suggestion
  return error_code == ERRCODE_UNDEFINED_COLUMN ||
         error_code == ERRCODE_UNDEFINED_SCHEMA ||
         error_code == ERRCODE_UNDEFINED_TABLE
```

Y-- (approved):
LGTM, main comment is around timeout default

src/pgduckdb_hooks.cpp
JelteF: As discussed over chat, I'm keeping the old one because clang-format formats that better imo (possibly something we want to address separately). Since it's not in a hot path this seems totally fine.

src/pgduckdb.cpp
JelteF: Done

Issue #523, opened 2025-01-09
DuckDB errors converted to PG error always report the same source location


JelteF:
### Description

```
\set VERBOSITY verbose
select * from duckdb.raw_query('aaaaa');
ERROR:  XX000: (PGDuckDB/pgduckdb_raw_query) Parser Error: syntax error at or near "aaaaa"
LINE 1: aaaaa
        ^
LOCATION:  __CPPFunctionGuard__, cpp_wrapper.hpp:28
Time: 9.314 ms
```

It would be great if this could report the source location that the duckdb error originated from (e.g. by extracting it from the stack trace)

PR #586, opened 2025-02-12
Print function name rather than caller

Y--:
When an error is raised, we used to print the `__func__` string, which contains the current function name where the macro is placed, hence the actual "caller" of the function being wrapped.

This PR replaces the caller name with the name of the function being wrapped.

In an initial version I was displaying both, but found it to be redundant.

An alternative would be to check if caller name contains the function name (or vice-versa) which happens often when we have functions such as `TopLevelFunc` and `TopLevelFunc_cpp`, and:
* if true, only display one
* if false, display both

But that sounds a bit much.

Happy to hear some feedback :-)

### DIFF
--- a/test/regression/expected/non_superuser.out
+++ b/test/regression/expected/non_superuser.out
@@ -26,7 +26,7 @@ ERROR:  permission denied for table extensions
 -- permissions checked even if it happens straight from DuckDB.
 SET duckdb.force_execution = false;
 SELECT * FROM duckdb.raw_query($$ SELECT * FROM pgduckdb.public.t $$);
-ERROR:  (PGDuckDB/pgduckdb_raw_query) Executor Error: (PGDuckDB/PostgresTableReader) permission denied for table t
+ERROR:  (PGDuckDB/pgduckdb_raw_query_cpp) Executor Error: (PGDuckDB/ExecutorStart) permission denied for table t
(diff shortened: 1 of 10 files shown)


test/regression/expected/non_superuser.out
```diff
@@ -108,13 +108,13 @@ SELECT * FROM duckdb.install_extension('iceberg');
 
 -- We should handle SQL injections carefully though to only allow INSTALL
 SELECT * FROM duckdb.install_extension($$ '; select * from hacky '' $$);
-ERROR:  (PGDuckDB/install_extension) HTTP Error: Failed to download extension " '; select * from hacky '' " at URL "http://extensions.duckdb.org/v1.2.0/linux_amd64/ '; select * from hacky '' .duckdb_extension.gz" (HTTP 403)
+ERROR:  (PGDuckDB/install_extension_cpp) HTTP Error: Failed to download extension " '; select * from hacky '' " at URL "http://extensions.duckdb.org/v1.2.0/osx_arm64/ '; select * from hacky '' .duckdb_extension.gz" (HTTP 403)
```
Y--, line 111: Will revert the platform changes :/

JelteF (approved):
Looks good. For now at least. We might want to change this once we fix: https://github.com/duckdb/pg_duckdb/issues/523

PR #588, opened 2025-02-13
fix sql syntax error

tjwebb:
previous version was incorrect. would give this error:

```
ERROR:  type "i" does not exist
LINE 1: ...ELECT COUNT(i) FROM read_parquet('file.parquet') AS (int i);
```


JelteF:
Thanks for the contribution, but with #575 this syntax and code examples were completely changed. So closing this in favor of #575.

PR #603, opened 2025-02-14
Avoid shadowing locks with locks taken in macros

JelteF:
I got some warnings when using pd_log in certain places. This fixes those. It also changes the name in the `__PostgresFunctionGuard__` template to `lock`, because variables in templates don't shadow variables at invocation.

### DIFF
--- a/include/pgduckdb/pgduckdb_utils.hpp
+++ b/include/pgduckdb/pgduckdb_utils.hpp
@@ -81,7 +81,7 @@ struct PostgresScopedStackReset {
 template <typename Func, Func func, typename... FuncArgs>
 typename std::invoke_result<Func, FuncArgs...>::type
 __PostgresFunctionGuard__(const char *func_name, FuncArgs... args) {
-	std::lock_guard<std::recursive_mutex> lk(pgduckdb::GlobalProcessLock::GetLock());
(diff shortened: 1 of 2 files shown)


Y-- (approved):

include/pgduckdb/pgduckdb_utils.hpp
```diff
@@ -81,7 +81,7 @@ struct PostgresScopedStackReset {
 template <typename Func, Func func, typename... FuncArgs>
 typename std::invoke_result<Func, FuncArgs...>::type
 __PostgresFunctionGuard__(const char *func_name, FuncArgs... args) {
-	std::lock_guard<std::recursive_mutex> lk(pgduckdb::GlobalProcessLock::GetLock());
+	std::lock_guard<std::recursive_mutex> __pg_func_guard_lock(pgduckdb::GlobalProcessLock::GetLock());
```
Y--, line 84: Let's rename this to `lock`: this is not a macro but a template

Y-- (approved):

dentiny:
Just came across this PR, a commonly used technique in macro is to assigned it a unique suffix, this is what I did in ray:
- macro definition: https://github.com/ray-project/ray/blob/master/src/ray/common/macros.h
- usage: https://github.com/ray-project/ray/blob/69ded2b0f4a1edffac276bb3c31ce0cfaa5cd70e/src/ray/common/status_or.h#L36-L42

Issue #567, opened 2025-01-30
Server crashes when executing "SELECT approx_count_distinct(a), approx_count_distinct(b) FROM t;"


saygoodbyye:
### What happens?

Server crashes when executing the script below
PostgreSQL build:
```
CFLAGS='-Og' ./configure --enable-tap-tests --with-openssl --enable-debug --enable-cassert --with-icu --with-lz4 --with-libxml
```

### To Reproduce

Sometimes you need to reproduce this sql-scripts several time to reproduce segfault.
```
SELECT duckdb.install_extension('aws');
SELECT duckdb.install_extension('postgres');
CREATE TABLE table_missing_attrs(id int);
INSERT INTO table_missing_attrs VALUES (1);
DROP TABLE table_missing_attrs;
CREATE TABLE t (a int, b text);
INSERT INTO t VALUES (1, 'a'), (2, 'b'), (3, 'c'), (4, 'd'), (5, 'e');
INSERT INTO t VALUES (2, 'f'), (3, 'g'), (4, 'h');
SELECT approx_count_distinct(a), approx_count_distinct(b) FROM t; --server crashes here
```

backtrace:
```
#0  0x00007f67705dd135 in duckdb::CatalogEntry::~CatalogEntry() () from /tmp/pgsql/lib/libduckdb.so
#1  0x00007f6774f9ea89 in pgduckdb::PostgresTable::~PostgresTable (this=0x55b610cd0df0, __in_chrg=<optimized out>) at src/catalog/pgduckdb_table.cpp:25
#2  0x00007f6774fa176e in std::default_delete<pgduckdb::PostgresTable>::operator() (__ptr=<optimized out>, this=<optimized out>) at /usr/include/c++/12/bits/unique_ptr.h:89
#3  std::unique_ptr<pgduckdb::PostgresTable, std::default_delete<pgduckdb::PostgresTable> >::~unique_ptr (this=0x55b610cd1e18, __in_chrg=<optimized out>)
    at /usr/include/c++/12/bits/unique_ptr.h:396
#4  duckdb::unique_ptr<pgduckdb::PostgresTable, std::default_delete<pgduckdb::PostgresTable>, true>::~unique_ptr (this=0x55b610cd1e18, __in_chrg=<optimized out>)
    at third_party/duckdb/src/include/duckdb/common/unique_ptr.hpp:13
[...]

PR #608, opened 2025-02-14
Fix compiler warning

Y--:
As reported here https://github.com/duckdb/pg_duckdb/issues/567#issuecomment-2626610442


JelteF (approved):
Ugh, random compilers complaining about dumb things.

Issue #598, opened 2025-02-14
Support for converting DuckDB VARINT type to Postgres NUMERIC


JelteF:
### Description

We should be able to convert the DuckDB VARINT type to the Postgres NUMERIC type.  

Once we do we can un-exclude varint here: https://github.com/duckdb/pg_duckdb/blob/02ebeaec1b42619fab806fc881e67de0821b380d/test/regression/expected/test_all_types.out#L7

dentiny:
Hi can I take this ticket? I plan to work on it next week, let me know if it's ok.
Feel free to ping me if you think I work slow or timeline tight, thank you!

JelteF:
Sounds great and no rush

PR #626, opened 2025-02-26
VARINT and numerics type conversion

dentiny:
Fixes https://github.com/duckdb/pg_duckdb/issues/598

This uses the VARINT its string representation as an intermediate translation between a DuckDB VARINT and a Postgres NUMERIC. While this is probably not the most efficient, that's acceptable due to making the code much simpler and VARINTs not being particularly popular types.


src/pgduckdb_types.cpp
```diff
@@ -337,6 +337,16 @@ ConvertNumeric(const duckdb::Value &ddb_value, idx_t scale, NumericVar &result)
 static Datum
 ConvertNumericDatum(const duckdb::Value &value) {
 	auto value_type_id = value.type().id();
+
+	// Special handle duckdb VARINT type.
+	if (value.type().id() == duckdb::LogicalTypeId::VARINT) {
+		std::string value_str = value.ToString();
+		Datum pg_numeric = DirectFunctionCall3(numeric_in, CStringGetDatum(value_str.c_str()),
+		                                       ObjectIdGetDatum(InvalidOid), Int32GetDatum(-1));
```
JelteF, line 345: For this conversion I appreciate the simplicity that the string conversion gives us. Especially since VARINTs are fairly rare types and already a pretty expensive anyway. So converting making the conversion to Postgres a bit more expensive than strictly necessary seems acceptable. This does need a `PostgresFunctionGuard` though, to keep Postgres C code a bit more physically separated from the C++ code let's move this to logic to the `src/pg/types.cpp` file. (we're steadily moving more code to the `src/pg` directory, but we're not done with that migration yet)

JelteF (changes requested):
Thanks for working on this. I think this is mostly good, but as in the other comment. Let's move the code to `src/pg/types.cpp`.

src/pgduckdb_types.cpp
dentiny: > This does need a PostgresFunctionGuard though, to keep Postgres C code a bit more physically separated from the C++ code

Make sense! Function guard added.

Just to confirm, I don't think we need type conversion for numerics <-> VARINT, right?
Because the entrypoint is in pg and we have other mapping (double) in duckdb.

> let's move this to logic to the src/pg/types.cpp file.

I could understand the motivation (i.e. place pg related code under `pg` directory for better maintainability), but I'm not sure which part to place, should I move all functions there?

As you might be aware of, I'm also working on other type conversions; do you think it ok after I finish them?
I'm willing to help migration later on :)

src/pgduckdb_types.cpp
```diff
@@ -334,9 +334,29 @@ ConvertNumeric(const duckdb::Value &ddb_value, idx_t scale, NumericVar &result)
 	}
 }
 
+// A wrapper function to convert a string to a numeric value, used for `PostgresFunctionGuard` macro.
+static Datum
+ConvertStringToNumerics(Datum str, Datum typelem, Datum typmod) {
+	Datum pg_numeric = DirectFunctionCall3(numeric_in, str, typelem, typmod);
```
dentiny, line 340: otherwise compilation failure
```
src/pgduckdb_types.cpp: In function 'Datum pgduckdb::ConvertNumericDatum(const duckdb::Value&)':
include/pgduckdb/pgduckdb_utils.hpp:106:68: error: no matching function for call to '__PostgresFunctionGuard__<Datum (*)(FunctionCallInfo), numeric_in>(const char [11], Datum, Datum, Datum)'
  106 |         pgduckdb::__PostgresFunctionGuard__<decltype(&FUNC), &FUNC>(#FUNC, ##__VA_ARGS__)
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
src/pgduckdb_types.cpp:346:42: note: in expansion of macro 'PostgresFunctionGuard'
  346 |                 const Datum pg_numeric = PostgresFunctionGuard(numeric_in, CStringGetDatum(value_str.c_str()),
      |                                          ^~~~~~~~~~~~~~~~~~~~~
include/pgduckdb/pgduckdb_utils.hpp:83:1: note: candidate: 'template<class Func, Func func, class ... FuncArgs> typename std::invoke_result<Func, FuncArgs ...>::type pgduckdb::__PostgresFunctionGuard__(const char*, FuncArgs ...)'
   83 | __PostgresFunctionGuard__(const char *func_name, FuncArgs... args) {
      | ^~~~~~~~~~~~~~~~~~~~~~~~~
```

src/pgduckdb_types.cpp
```diff
@@ -334,9 +334,29 @@ ConvertNumeric(const duckdb::Value &ddb_value, idx_t scale, NumericVar &result)
...
 	auto value_type_id = value.type().id();
+
+	// Special handle duckdb VARINT type.
+	if (value.type().id() == duckdb::LogicalTypeId::VARINT) {
+		// The performant way to handle the translation is to parse VARINT out, here we leverage string conversion and
+		// parsing mainly for code simplicity.
+		const std::string value_str = value.ToString();
+		const Datum pg_numeric =
+		    PostgresFunctionGuard(ConvertStringToNumerics, /*str=*/CStringGetDatum(value_str.c_str()),
+		                          /*typelen=*/ObjectIdGetDatum(InvalidOid), /*typmod=*/Int32GetDatum(-1));
```
JelteF, line 355: I think it makes most sense to create a function
```c
Datum
StringToNumeric(char* str) {
	// ...
}
```

And then have that function call the actual function with a PostgresFunctionGuard and name that actual function `StringToNumeric_C`. Similarly to how `GetBaseDuckColumnType` is implemented.
https://github.com/duckdb/pg_duckdb/blob/a8425cb1d5599312d24f4e6f9cd9aad1c1a2bee9/src/pg/types.cpp#L35-L54

src/pgduckdb_types.cpp
```diff
@@ -13,22 +13,21 @@ extern "C" {
...
 #include "executor/tuptable.h"
+#include "utils/array.h"
 #include "utils/builtins.h"
+#include "utils/date.h"
+#include "utils/lsyscache.h"
+#include "utils/jsonb.h"
 #include "utils/numeric.h"
 #include "utils/uuid.h"
-#include "utils/array.h"
-#include "fmgr.h"
```
dentiny, line 26: Notice `fmgr.h` is included twice, sort the headers so we don't get confused.

src/pgduckdb_types.cpp
```diff
@@ -13,22 +13,21 @@ extern "C" {
 
 #include "pgduckdb/vendor/pg_numeric_c.hpp"
 
-#include "postgres.h"
 #include "fmgr.h"
+#include "postgres.h"
```
JelteF, line 17: `postgres.h` should be at the top rest looks fine.

dentiny: gotcha, updated

JelteF (approved):

Issue #597, opened 2025-02-14
Support for BIT type


JelteF:
### Description

We don't support conversion of the BIT/BITSTRING types yet. https://duckdb.org/docs/sql/data_types/bitstring

Once we do we can un-exclude it here: https://github.com/duckdb/pg_duckdb/blob/02ebeaec1b42619fab806fc881e67de0821b380d/test/regression/expected/test_all_types.out#L10

PR #628, opened 2025-02-26
BIT/VARBIT type conversion between pg and duckdb

dentiny:
Resolves https://github.com/duckdb/pg_duckdb/issues/597

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -157,6 +157,14 @@ struct DecimalConversionDouble {
 	}
 };
 
+static Datum
+ConvertVarbitDatum(const duckdb::Value &value) {
+	const std::string value_str = value.ToString();
+	Datum pg_varbit = DirectFunctionCall3(varbit_in, CStringGetDatum(value_str.c_str()), ObjectIdGetDatum(VARBITOID),
+	                                      Int32GetDatum(-1));
+	return pg_varbit;
+}
+
 static inline Datum
 ConvertBoolDatum(const duckdb::Value &value) {
 	return value.GetValue<bool>();
@@ -765,6 +773,10 @@ ConvertDuckToPostgresValue(TupleTableSlot *slot, duckdb::Value &value, idx_t col
 	Oid oid = slot->tts_tupleDescriptor->attrs[col].atttypid;
 
 	switch (oid) {
+	case VARBITOID: {
+		slot->tts_values[col] = ConvertVarbitDatum(value);
+		break;
+	}
 	case BOOLOID:
 		slot->tts_values[col] = ConvertBoolDatum(value);
 		break;
@@ -1071,6 +1083,8 @@ GetPostgresArrayDuckDBType(const duckdb::LogicalType &type) {
 Oid
 GetPostgresDuckDBType(const duckdb::LogicalType &type) {
 	switch (type.id()) {
+	case duckdb::LogicalTypeId::BIT:
+		return VARBITOID;
 	case duckdb::LogicalTypeId::BOOLEAN:
 		return BOOLOID;
 	case duckdb::LogicalTypeId::TINYINT:
--- a/test/regression/expected/test_all_types.out
+++ b/test/regression/expected/test_all_types.out
@@ -7,7 +7,6 @@ SELECT * exclude(
     varint,
     TIME,
     time_tz,
-    bit,
     small_enum,
     medium_enum,
     large_enum,
@@ -55,6 +54,7 @@ uuid            | 00000000-0000-0000-0000-000000000000
 interval        | @ 0
 varchar         | 🦆🦆🦆🦆🦆🦆
 blob            | thisisalongblob\000withnullbytes
+bit             | 0010001001011100010101011010111
 int_array       | {}
 double_array    | {}
 date_array      | {}
@@ -87,6 +87,7 @@ uuid            | ffffffff-ffff-ffff-ffff-ffffffffffff
 interval        | @ 83 years 3 mons 999 days 16 mins 39.999999 secs
 varchar         | goo
 blob            | \000\000\000a
+bit             | 10101
 int_array       | {42,999,NULL,NULL,-42}
 double_array    | {42,NaN,Infinity,-Infinity,NULL,-42}
 date_array      | {01-01-1970,07-11-5881580,07-13-5881580,NULL,05-12-2022}
@@ -119,6 +120,7 @@ uuid            |
 interval        | 
 varchar         | 
 blob            | 
+bit             | 
 int_array       | 
 double_array    | 
 date_array      | 
--- a/test/regression/sql/test_all_types.sql
+++ b/test/regression/sql/test_all_types.sql
@@ -7,7 +7,6 @@ SELECT * exclude(
     varint,
     TIME,
     time_tz,
-    bit,
     small_enum,
     medium_enum,
     large_enum,


src/pgduckdb_types.cpp
```diff
@@ -157,6 +157,14 @@ struct DecimalConversionDouble {
 	}
 };
 
+static Datum
+ConvertVarbitDatum(const duckdb::Value &value) {
+	const std::string value_str = value.ToString();
+	Datum pg_varbit = DirectFunctionCall3(varbit_in, CStringGetDatum(value_str.c_str()), ObjectIdGetDatum(VARBITOID),
+	                                      Int32GetDatum(-1));
```
JelteF, line 164: Similar feedback as for time/timetz PR. I think it should be fairly easy to convert the internal representations directly to eachother, without having to go through string formatting/parsing.

JelteF (changes requested):
Same feedback as for the time/timetz PR. This should also implement conversion from PG types to DuckDB types (both for [`bit` and `bit varying`](https://www.postgresql.org/docs/current/datatype-bit.html))

src/pgduckdb_types.cpp
dentiny: In duckdb `BIT` type is stored as `VARCHAR` physically.
https://github.com/duckdb/duckdb/blob/057f9753cbd3eb6d07543c2f1f1085de961fffc4/src/common/types.cpp#L120-L122

Even if we want to the parsing ourselves (from char* and size_t), we need to have a way to get pointer and its length.

- We're able to get the pointer via `GetPointer`: https://github.com/duckdb/duckdb/blob/057f9753cbd3eb6d07543c2f1f1085de961fffc4/src/include/duckdb/common/types/value.hpp#L227
- But the length information is stored inside of extra value info: https://github.com/duckdb/duckdb/blob/057f9753cbd3eb6d07543c2f1f1085de961fffc4/src/include/duckdb/common/types/value.hpp#L341, the actual derived type is `StringValueInfo`, which lives inside of an anonymous namespace: https://github.com/duckdb/duckdb/blob/057f9753cbd3eb6d07543c2f1f1085de961fffc4/src/common/types/value.cpp#L74-L95

So IMO, convert the `Value` into a string is a simple and reasonable implementation.

JelteF: Okay looking closer at the way that the bit types are stored internally I now agree that going through the string representation is probably the best. I thought you could do something like this:

Do the same as `ConvertBinaryDatum`to get the raw value, except that you need to use some functions from duckdb its `src/common/types/bit.cpp`. I think you can do something very similar to what BitToBlob does: https://github.com/duckdb/duckdb/blob/057f9753cbd3eb6d07543c2f1f1085de961fffc4/src/common/types/bit.cpp#L175-L190
```c
	duckdb::string_t str = value.GetValueUnsafe<duckdb::string_t>();
    // get padding
	// copy raw bytes
```

Afaict that would work fine, were it not for the fact that the padding bit in the PG representation are on the end of the bitstring, and not on the beginning like in DuckDB. So to copy you couldn't use a plain `memcpy`, because you'd need to copy the actual data with bit offset (that's not divisible by 8). That definitely seems too much complexity for this rather exotic data type. So yeah, let's go through the string representation instead.

Let's move this logic to `pg/types.cpp` though just like you did for VARINT.

dentiny: Thank for explaining your thoughts! That's very helpful!

My initial thought is memcpy as well, I was just thinking if it's possible to avoid handling padding and byte order myself.

JelteF: I think this code looks good. But it still needs conversion the other way around, i.e. from PG to DuckDB  (just like you did for the `time` type).

dentiny: Actually I thought about it, for postgres the entrypoint type if BITOID / VARBITOID, when duckdb takes on the read side the type is BIT, so there's no 1-1 mapping between pg/duckdb types.

JelteF: The mapping, doesn't have to be 1-1. As long as each type converts into something sensible in the other typesystem it's fine. For example, if you have a PG table that stores a BITOID, and we then we convert that to a DuckDB BIT, and then back to PG VARBITOID, yes you've lost a bit of type information. But that's still a lot better than not being able to read a BITOID columns at all. It's a tradeoff, and so far we're trading nice interoperability for strictly correct 1-1 mappings.

test/regression/expected/test_all_types.out
```diff
@@ -6,7 +6,6 @@ SELECT * exclude(
     tinyint, -- PG14 outputs this differently currently
     TIME,
     time_tz,
-    bit,
```
JelteF, line 9: The `docs/types.md` file should also be updated that these types are now supported.

dentiny:
Sorry for the latency, I will re-work on it tonight. Thanks for the patience!

JelteF:
@dentiny any progress?

dentiny:
> @dentiny any progress?

Hi @JelteF sorry I was distracted by something else, I have restarted working on it. I'm truly sorry for the delay!

test/regression/expected/test_all_types.out
dentiny: sure, add a new line for bit related types.

dentiny:
> @dentiny any progress?

Hi @JelteF , I apologize again for the delay, it should be good to review now.

JelteF (approved):
I made some changes to this and I think it's good now. I added a few more tests and simplified the conversion logic by using duckdb its `Bit::ToBit` function to create the bitstring instead of manually building it.

src/pg/types.cpp
```diff
@@ -67,4 +67,26 @@ StringToNumeric(const char *str) {
 	return PostgresFunctionGuard(StringToNumeric_C, str);
 }
 
+static Datum
+StringToVarbit_C(const char *str) {
+	Datum pg_varbit = DirectFunctionCall3(varbit_in, CStringGetDatum(str), /*typelen=*/ObjectIdGetDatum(VARBITOID),
+	                                      /*typmod=*/Int32GetDatum(-1));
+	return pg_varbit;
```
Y--, line 74: nit-
```suggestion
	return DirectFunctionCall3(varbit_in, CStringGetDatum(str), /*typelen=*/ObjectIdGetDatum(VARBITOID),
	                                      /*typmod=*/Int32GetDatum(-1));
```

src/pgduckdb_types.cpp
```diff
@@ -158,6 +160,18 @@ struct DecimalConversionDouble {
...
+// Util function to convert duckdb `BIT` value to postgres bitstring type.
+// There're two possible corresponding types `BITOID` and `VARBITOID`, here we convert to `VARBITOID` for generality.
+static Datum
+ConvertVarbitDatum(const duckdb::Value &value) {
+	const std::string value_str = value.ToString();
+
+	// Here we rely on postgres conversion function, instead of manual parsing, because BIT string type involves padding
+	// and duckdb/postgres handle it differently, it's non-trivial to memcpy the bits.
+	Datum pg_varbit = pgduckdb::pg::StringToVarbit(value_str.c_str());
+	return pg_varbit;
```
Y--, line 172:
```suggestion
	// Here we rely on postgres conversion function, instead of manual parsing, because BIT string type involves padding
	// and duckdb/postgres handle it differently, it's non-trivial to memcpy the bits.
	return pgduckdb::pg::StringToVarbit(value.ToString().c_str());
```

Y-- (approved):

dentiny:
> I made some changes to this and I think it's good now. I added a few more tests and simplified the conversion logic by using duckdb its `Bit::ToBit` function to create the bitstring instead of manually building it.

It looks perfect, thank you!

PR #633, opened 2025-02-27
Avoid copy for VARCHAR type conversion

dentiny:
If I don't read it wrong, the actual content inside of `VARCHAR` will be mem-copied into `result`, so we could use fat-pointer here (as long as there's no tricky padding issues like VARBIT).

How I tested:
- I build with `make install -j32`
- I tested with `make installcheck`

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -192,9 +192,9 @@ ConvertInt8Datum(const duckdb::Value &value) {
 
 static Datum
 ConvertVarCharDatum(const duckdb::Value &value) {
-	auto str = value.GetValue<duckdb::string>();
-	auto varchar = str.c_str();
-	auto varchar_len = str.size();
+	auto str = value.GetValueUnsafe<duckdb::string_t>();
+	auto varchar = str.GetDataUnsafe();
+	auto varchar_len = str.GetSize();
 
 	text *result = (text *)palloc0(varchar_len + VARHDRSZ);
 	SET_VARSIZE(result, varchar_len + VARHDRSZ);


JelteF:
I'm not sure what the benefit of this is? Is this supposed to be faster? If so do you have some numbers?

dentiny:
> I'm not sure what the benefit of this is? Is this supposed to be faster? If so do you have some numbers?

Yeah that's my motivation. Sorry I should provide more thoughts in the PR description.

From previous PRs, I feel we do care about type conversion's performance, since it sits in the center for pg/duckdb interaction. Use TIME/TIMETZ for example, one of the reasons is performance (another is definitely not too hard to write).

Initially I took the short path because TIME/TIMETZ is supposed to be short, so we have an upper bound for how much memory / parse CPU we use; checking the code, I found VARCHAR conversion is taking memory allocation every time, which is worse than TIME becase
- (major) VARCHAR could be large, which is easier to cause performance issue
- (minor) I use string types much often than other types

As for benchmark, I could try to get some numbers :)

JelteF:
Looking at the duckdb implementation I can see how `GetValueUnsafe` is faster than `GetValue` but not how string_t is faster than `string` in this case:
```c
template <>
string Value::GetValue() const {
	return ToString();
}

// snip

string Value::ToString() const {
	if (IsNull()) {
		return "NULL";
	}
	return StringValue::Get(DefaultCastAs(LogicalType::VARCHAR));
}

//snip

template <>
string Value::GetValueUnsafe() const {
	return StringValue::Get(*this);
}

template <>
DUCKDB_API string_t Value::GetValueUnsafe() const {
	return string_t(StringValue::Get(*this));
}
```

The `GetValueUnsafe` for `string_t` does strictly more than the one for `string`. Also `string_t` its constructor from a `string` doesn't convert it to a fat pointer:
```
	string_t(const string &value) // NOLINT: Allow implicit conversion from `const char*`
	    : string_t(value.c_str(), UnsafeNumericCast<uint32_t>(value.size())) {
	}
```

JelteF:
Looking a bit closer, it seems that this is primarily because we use the `GetValue()` API here (which interestingly already has a FIXME comment for another reason):

https://github.com/duckdb/pg_duckdb/blob/66cb6e64feb40bc1b0a7511485703442e39e60ad/src/pgduckdb_node.cpp#L211-L214

And duckdb then transforms that `string_t` value in the vector to an actual `string` (thus doing an allocation):
https://github.com/duckdb/duckdb/blob/d76863224db11a712edb95cd861d98f7900b0ada/src/common/types/vector.cpp#L694-L697

dentiny:
Thank you for the explanation! I think I mis-understand the issue --- my initial thought is when we do conversion for `VARCHAR` the `Value` object is always valid so we should be able to save a memory allocation + memory copy via `string_view`-like API/data structure.

Actually after your reference code above, I feel there're quite more things we could do here.

PR #636, opened 2025-02-28
postgres/duckdb TIMETZ conversion

### LINKED ISSUE #596: Add support for TIME and TIME_TZ types
### Description

These are the only time related types that we don't support yet. Once supported we can un-exclude these columns: https://github.com/duckdb/pg_duckdb/blob/02ebeaec1b42619fab806fc881e67de0821b380d/test/regression/expected/test_all_types.out#L8-L9

dentiny:
Resolve the second part for https://github.com/duckdb/pg_duckdb/issues/596 (TIMETZ conversion).

Fixes #596

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -236,6 +236,18 @@ ConvertTimeDatum(const duckdb::Value &value) {
 	return Int64GetDatum(pg_time);
 }
 
+static Datum
+ConvertTimeTzDatum(const duckdb::Value &value) {
+	duckdb::dtime_tz_t dt_tz = value.GetValue<duckdb::dtime_tz_t>();
+	const int64_t micros = dt_tz.time().micros;
+	const int32_t tz_offset = dt_tz.offset();
+
+	TimeTzADT *result = static_cast<TimeTzADT *>(palloc(sizeof(TimeTzADT)));
+	result->time = micros;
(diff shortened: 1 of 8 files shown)


src/pgduckdb_types.cpp
```diff
@@ -236,6 +236,18 @@ ConvertTimeDatum(const duckdb::Value &value) {
...
 
+static Datum
+ConvertTimeTzDatum(const duckdb::Value &value) {
+	duckdb::dtime_tz_t dt_tz = value.GetValue<duckdb::dtime_tz_t>();
+	const int64_t micros = dt_tz.time().micros;
+	const int32_t tz_offset = dt_tz.offset();
+
+	TimeTzADT *result = static_cast<TimeTzADT *>(palloc(sizeof(TimeTzADT)));
+	result->time = micros;
+	result->zone = -tz_offset;
```
JelteF, line 247: Why use `-` in this line? That's unexpected enough that I think it deserves a comment.

JelteF (approved):
Looks good, except for some code comments that should be added for clarity.

src/pgduckdb_types.cpp
dentiny: Totally makes sense, add a comment on how pg/duckdb stores timezone.

JelteF (approved):

PR #642, opened 2025-03-03
Attempt to fix httpfs semantics

dentiny:
I tried to upstream the fix to duckdb httpfs extension: https://github.com/duckdb/duckdb-httpfs/pull/17 (also contains details for what I try to fix), but unfortunately no reply yet. I know pg_duckdb forks httpfs by source code, the same fix should apply.

Test command and result:
```sql
postgres (pid: 15453) =# SELECT * FROM read_csv('https://raw.githubusercontent.com/dentiny/duck-read-cache-fs/refs/heads/main/test/data/stock-exchanges.csv');
-[ RECORD 1 ]--+-----------------------------------------------------------
csvbase_row_id | 1
Continent      | Africa
Country        | Lesotho
Name           | HYBSE
MIC            | 
Last changed   | 2019-03-25
<omit other lines>
postgres (pid: 15453) =# SELECT COUNT(*) FROM read_csv('https://raw.githubusercontent.com/dentiny/duck-read-cache-fs/refs/heads/main/test/data/stock-exchanges.csv');
-[ RECORD 1 ]
count | 251
```

### DIFF
--- a/third_party/cached_httpfs/httpfs.cpp
+++ b/third_party/cached_httpfs/httpfs.cpp
@@ -508,7 +508,6 @@ void HTTPFileSystem::Read(FileHandle &handle, void *buffer, int64_t nr_bytes, id
 			throw InternalException("Cached file not initialized properly");
 		}
 		hfh.cached_file_handle->Read(buffer, nr_bytes, location);
-		hfh.file_offset = location + nr_bytes;
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


dentiny:
> In general we try to keep the same behavior as upstream DuckDB. Even if there's a bug. Otherwise if the bug gets fixed differently upstream, we somehow need to revert our fix again.

Make sense.

PR #643, opened 2025-03-03
Add additional functionality for the duckdb.unresolved_type

JelteF:
In 0.3.0 we added the `duckdb.unresolved_type` type and implemented a bunch of
operators, aggregates and functions for it. Of course the list will always be
incomplete and casts to actual types might always be necessary in some cases.
But for usability we should still try to implement the most commonly used
functionality. So this adds a bunch of additional things that I ran into over
the past weeks:

1. The `min` aggregate, this was not present in 0.3.0 by accident.
   Probably because DuckDB has two `min` aggregates.
2. Casts to `time`, `timetz`, `interval` and `bit`
3. Support for `date_trunc`, `length` and `regexp_replace`
4. Add general support for the `strftime` and `strptime` functions. This
   is useful even for already resolved/casted `timestamp`/`date` values
   because Postgres its date formatting functions like `to_char` are not
   available in DuckDB.

### DIFF
--- a/sql/pg_duckdb--0.3.0--0.4.0.sql
+++ b/sql/pg_duckdb--0.3.0--0.4.0.sql
@@ -1,2 +1,197 @@
 -- Add "url_style" column to "secrets" table
 ALTER TABLE duckdb.secrets ADD COLUMN url_style TEXT;
+
+-- The min aggregate was somehow missing from the list of aggregates in 0.3.0
+CREATE AGGREGATE @extschema@.min(duckdb.unresolved_type) (
+    SFUNC = duckdb_unresolved_type_state_trans,
+    STYPE = duckdb.unresolved_type,
+    FINALFUNC = duckdb_unresolved_type_final
+);
+
+CREATE FUNCTION @extschema@.strftime(date, text) RETURNS text
+SET search_path = pg_catalog, pg_temp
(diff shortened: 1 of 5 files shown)


src/pg/guc.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "utils/guc.h"
+}
+
+namespace pgduckdb::pg {
+
+const char *
+GetConfigOption(const char *name, bool missing_ok, bool restrict_privileged) {
+	return PostgresFunctionGuard(::GetConfigOption, name, missing_ok, restrict_privileged);
+}
+} // namespace pgduckdb::pg
```
Y--, line 15: I'm a bit on the fence since they're only used in one file, but should we move the `DefineCustomVariable` helpers from `pgduckdb.cpp` here?

Y-- (approved):

src/pg/guc.cpp
JelteF: I think we should make all our functions in `src/pg/*.cpp` files be functions that are safe to call from C++. Since `DefineCustomVariable` are not, and don't really make sense to be, since they are called from `_PG_init`, I don't think that makes sense to move them. I'd like to move all our GUC definitions to a different file in general though.

PR #644, opened 2025-03-03
Remove code related to http caching

JelteF:
We had implemented our own http caching in #88 when nothing like that existed yet. Now there are two different threads that are implementing something like this in a better way than we have done. So this PR removes our custom caching code from our codebase. Instead in a future release we'll make sure that pg_duckdb works well with these at least one of these other caching solutions.

For reference the two projects that are currently implementing caching are:
1. An official in-memory cache by DuckDB, which will eventually improved to also support on-disk caching: https://github.com/duckdb/duckdb/pull/16463
2. A community provided extension called `cache_httpfs`: https://github.com/dentiny/duck-read-cache-fs


third_party/pg_duckdb_extensions.cmake
```diff
@@ -1,7 +1,7 @@
 duckdb_extension_load(json)
 duckdb_extension_load(icu)
 # This is relative to third_party/duckdb dir
-duckdb_extension_load(cached_httpfs
-    SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../cached_httpfs"
-    INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../cached_httpfs/include"
-)
\ No newline at end of file
+duckdb_extension_load(httpfs
```
Y--, line 4: Why not using the public one? (like `json`/`icu`?)

Y--: Doesn't `duckdb_extension_load(httpfs)` just do exactly that?

For example with json:
```
y=# select * from duckdb.raw_query($$select extension_name, loaded, installed, install_mode, install_path from duckdb_extensions() where extension_name = 'json';$$);
NOTICE:  result: extension_name	loaded	installed	install_mode	install_path
VARCHAR	BOOLEAN	BOOLEAN	VARCHAR	VARCHAR
[ Rows: 1]
json	true	true	STATICALLY_LINKED	(BUILT-IN)


 raw_query
-----------

(1 row)
```

Y--: Ah sorry I thought JSON was also out of tree...

Y--: Ok, this is it: https://github.com/duckdb/pg_duckdb/pull/644/commits/7579326fb8fb580e0e72fd6bb82d616cebc0d829

wuputah:
I would prefer to see more progress on supporting community extensions (it's [not clear][1] whether community extensions are currently supported) and possibly supporting "persistent" settings (i.e. `SET` commands that are run when initializing DuckDB) before kicking this out, unless we have a strong reason that this has to go immediately. 

That said, if we are going to remove / replace it -- which I do think we should -- I would certainly like to remove it before 1.0.

I would be curious, among those watching the repo, whether removing this feature without an immediate alternative would break their usage / use cases.

[1]: https://github.com/duckdb/pg_duckdb/blob/main/docs/extensions.md

JelteF:
@wuputah now that #647 is merged. Do you think we can merge this?

Y-- (approved):

wuputah:
My main concern now is that extensions [introduce settings](https://duckdb.org/community_extensions/extensions/cache_httpfs.html#added-settings) that the user may want to change. Since a pg_duckdb user does not directly interact with the DuckDB instance lifecycle, it's not clear if and when the user needs to (re-)set these settings. Basically, it's the same problem as secrets.

That said, the defaults for cache_httpfs seem sane to me, though I would like to see [this feature implemented](https://github.com/dentiny/duck-read-cache-fs/issues/103) which would then have some sort of setting to limit the disk cache size.

JelteF:
Okay, I'm merging this then. And we can add something to configure settings (or even general "duckdb setup commands") in a follow up PR.

PR #648, opened 2025-03-04
Make cpp_wrapper.hpp self contained

JelteF:
I was seeing errors in my editor when opening this file, because some symbols were undefined. This fixes that by including the necessary DuckDB header.


Y-- (approved):

Y--:
Technically we probably could have gotten away with a forward declaration, but that's maybe better to have it fully self contained.

PR #650, opened 2025-03-05
Fixes bugs an quality of life issues in MotherDuck support

JelteF:
This fixes a handful of issues with MotherDuck tables that I ran into while
implementing automated testing for MotherDuck integration:

1. When creating a table in a `ddb$` schema, default that table to be an MD table.
2. Do not allow creating non-MD tables in `ddb$` schemas when explicitly specifying another Table AM (like `heap`).
3. When creating MD tables, automatically make them owned by `duckdb.postgres_role`
4. Fix syncing of MotherDuck tables. This was broken by #632
5. Fix escaping of a MotherDuck schema name when syncing schema names


Y-- (approved):
👍🏼

PR #652, opened 2025-03-07
Add ALTER TABLE support for DuckDB tables

### LINKED ISSUE #536: add ALTER TABLE support for duckdb (motherduck) tables
### Description

Currently, you can't use ALTER TABLE to modify a table that uses the `duckdb` access method:

```
postgres=# ALTER TABLE a RENAME TO b;
ERROR:  DuckDB does not support ALTER TABLE yet
```

The basics to consider:

* [ ] #537
* [ ] #538
* [ ] #539
* [ ] #540
* [ ] #541

### LINKED ISSUE #537: ALTER TABLE RENAME support
```sql
CREATE TEMP TABLE abc(a int) using duckdb;
ALTER TABLE abc RENAME TO def;
```
currently throws the error:
```
ERROR:  XX000: DuckDB does not support ALTER TABLE yet
LOCATION:  duckdb_alter_table_trigger_cpp, pgduckdb_ddl.cpp:768
```

### LINKED ISSUE #538: add column

### LINKED ISSUE #539: rename column

### LINKED ISSUE #540: drop column

### LINKED ISSUE #541: ALTER TABLE ALTER COLUMN TYPE support

### LINKED ISSUE #542: inserting data from generate_series into motherduck table returns error
### Description

Given:

```sql
CREATE TABLE md (i integer) USING duckdb;
```

An insert using `generate_series` call will fail as follows:

```sql
INSERT INTO md SELECT generate_series(1, 1000);
```

```
ERROR:  (PGDuckDB/Duckdb_ExecCustomScan) Conversion Error: Unimplemented type for cast (BIGINT[] -> INTEGER)
LINE 1: INSERT INTO my_db.main.md (i) SELECT generate_series(1, 1000) AS generate...
```

Most likely, this would have to be fixed in duckdb itself, since I think this is a mismatch between Postgres and DuckDB behavior, but posting here first.

JelteF:
Adds all ALTER TABLE commands that DuckDB can parse except for COLLATION changes and foreign keys. Not all of these ALTER commands are actually supported by DuckDB (e.g. adding CHECK constraints). But this starts to least forward those queries correctly to DuckDB, so once DuckDB supports it pg_duckdb will start to automatically support it as well.

Fixes #536
Fixes #537
Fixes #538
Fixes #539
Fixes #540
Fixes #541
Fixes #542


src/pgduckdb_ddl.cpp
```diff
@@ -181,6 +187,7 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
 		char *access_method = stmt->accessMethod ? stmt->accessMethod : default_table_access_method;
 		bool is_duckdb_table = strcmp(access_method, "duckdb") == 0;
 		if (is_duckdb_table) {
+			pgduckdb::top_level_duckdb_ddl_type = pgduckdb::DDLType::CREATE_TABLE;
```
Y--, line 190: Shouldn't we assert that this is `DDLType::NONE` before updating it? (same everywhere we update it). Or we're confident this can never happen?

src/pgduckdb_ddl.cpp
```diff
@@ -416,6 +439,9 @@ DECLARE_PG_FUNCTION(duckdb_create_table_trigger) {
 		PG_RETURN_NULL();
 	}
 
+	/* Reset since we don't need it anymore */
+	pgduckdb::top_level_duckdb_ddl_type = pgduckdb::DDLType::NONE;
```
Y--, line 443: ~How is it reset if an exception occurs before?~ Actually expanding more the code it seems that it is reset because we're starting the function? Is the comment above correct?

src/pgduckdb_ddl.cpp
```diff
@@ -859,6 +885,9 @@ DECLARE_PG_FUNCTION(duckdb_alter_table_trigger) {
...
```
Y--, line 889: Same here?

src/pgduckdb_ruleutils.cpp
```diff
@@ -754,6 +762,358 @@ pgduckdb_get_tabledef(Oid relation_oid) {
...
+		}
+
+		default:
+			ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+			                errmsg("DuckDB does not support this ALTER TABLE command")));
+		}
+	}
+	relation_close(relation, AccessShareLock);
+
+	return (buffer.data);
```
Y--, line 1115: nit
```suggestion
	return buffer.data;
```

Y-- (commented):
LGTM, especially given all the tests it seems we have a reasonable coverage. Main question is around the global state lifecycle?

src/pgduckdb_ddl.cpp
JelteF: I'm pretty confident it cannot happen, but I guess it's better safe than sorry. So now it throws an error in that case.

src/pgduckdb_ddl.cpp
JelteF: Changed the logic now and updated the comment. On failure (and also on success to be safe) this value is also reset in `DuckdbXactCallback_Cpp` in `src/pgduckdb_xact.cpp`.

src/pgduckdb_ddl.cpp
Y--: Should we do the same below (L216?)

Issue #595, opened 2025-02-14
Handle dates that underflow differently


JelteF:
### Description

DuckDB dates can be further into the past than Postgres dates. We should either truncate the date to the lowest possible date that's representable in Postgres, or we should throw an error. Underflowing and returning a date in the future is the worst possible behavior.

This is a query that creates these underflowing dates:
```sql
SELECT * FROM duckdb.query($$
    select
        '5877642-06-25 (BC)'::date as date,
        '290309-12-22 (BC) 00:00:00'::timestamp as timestamp,
        '290309-12-22 (BC) 00:00:00'::timestamp_s as timestamp_s,
        '290309-12-22 (BC) 00:00:00'::timestamp_ms as timestamp_ms,
        '290309-12-22 (BC) 00:17:30+00:17'::timestamptz as timestamptz,
$$);
     date      │          timestamp           │         timestamp_s          │         timestamp_ms         │           timestamptz
───────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼─────────────────────────────────
 5881580-07-14 │ 294247-01-10 08:01:49.551616 │ 294247-01-10 08:01:49.551616 │ 294247-01-10 08:01:49.551616 │ 294247-01-10 09:02:19.551616+01
(1 row)
```

Example of underflowing dates are in this test: https://github.com/duckdb/pg_duckdb/blob/02ebeaec1b42619fab806fc881e67de0821b380d/test/regression/expected/test_all_types.out#L42-L47

Found by #577

kamuli4:
@JelteF I'd like to help with this, how should I start?

kamuli4:
Hi, I've created a pull request with my attempt to fix this issue. I closed #611 because it was accidentally opened as a draft

Issue #629, opened 2025-02-26
Datetime/date infinity is converted incorrectly


JelteF:
### Description

```sql
create table t(ts timestamp, d date);
insert into t values ('infinity', 'infinity'), ('-infinity', '-infinity');
select isfinite(ts), isfinite(d) from t; -- returns false everywhere (as it should)
set duckdb.force_execution = true;
select isfinite(ts), isfinite(d) from t; -- returns true everywhere (wrong)
```

This is also clearly wrong:
```sql
> select * from duckdb.query($$ select 'infinity'::timestamp, 'infinity'::date $$);
 CAST('infinity' AS TIMESTAMP) │ CAST('infinity' AS DATE)
───────────────────────────────┼──────────────────────────
 294247-01-10 04:00:54.775807  │ 5881580-07-11
```

Might be related to #595

shah-nirmit:
@JelteF this seems to be solved now

JelteF:
Fixed by #653

PR #653, opened 2025-03-07
Add boundary check for converting duckdb date to PG dates

shah-nirmit:
Postgres Only Supports date specific date range But DuckDb can return values outside this range , we should detect this and Either error out or pass boundary values.

Fixes https://github.com/duckdb/pg_duckdb/issues/595

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -36,6 +36,7 @@ extern "C" {
 namespace pgduckdb {
 
 NumericVar FromNumeric(Numeric num);
+bool ValidDate(duckdb::date_t dt);
 
 struct NumericAsDouble : public duckdb::ExtraTypeInfo {
 	// Dummy struct to indicate at conversion that the source is a Numeric
@@ -216,6 +217,12 @@ ConvertBinaryDatum(const duckdb::Value &value) {
 inline Datum
 ConvertDateDatum(const duckdb::Value &value) {
 	duckdb::date_t date = value.GetValue<duckdb::date_t>();
+	if(!ValidDate(date))
+		throw duckdb::OutOfRangeException(
+			"The value should be between min and max value (%s <-> %s)",
+			duckdb::Date::ToString(pgduckdb::PGDUCKDB_PG_MIN_DATE_VALUE),
+			duckdb::Date::ToString(pgduckdb::PGDUCKDB_PG_MAX_DATE_VALUE));
+	
 	return date.days - pgduckdb::PGDUCKDB_DUCK_DATE_OFFSET;
 }
 
@@ -1646,4 +1653,11 @@ FromNumeric(Numeric num) {
 	return dest;
 }
 
+bool
+ValidDate(duckdb::date_t dt)
+{
+	if (dt < pgduckdb::PGDUCKDB_PG_MIN_DATE_VALUE || dt > pgduckdb::PGDUCKDB_PG_MAX_DATE_VALUE)
+		return false;
+	return true;
+}
 } // namespace pgduckdb
--- a/include/pgduckdb/pgduckdb_types.hpp
+++ b/include/pgduckdb/pgduckdb_types.hpp
@@ -15,6 +15,17 @@ constexpr int32_t PGDUCKDB_DUCK_DATE_OFFSET = 10957;
 constexpr int64_t PGDUCKDB_DUCK_TIMESTAMP_OFFSET =
     static_cast<int64_t>(PGDUCKDB_DUCK_DATE_OFFSET) * static_cast<int64_t>(86400000000) /* USECS_PER_DAY */;
 
+// Check from regress/sql/date.sql
+#define PG_MINYEAR (-4713)
+#define PG_MINMONTH (11)
+#define PG_MINDAY (24)
+#define PG_MAXYEAR (5874897)
+#define PG_MAXMONTH (12)
+#define PG_MAXDAY (31)
+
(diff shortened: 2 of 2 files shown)


shah-nirmit:
@JelteF 
These changes throw error if the values are out of bounds , which breaks regression test named `test_all_types`
```
...
not ok 9     - test_all_types                             21 ms
....
1..34
# 1 of 34 tests failed.
```

JelteF:
Thanks for the contribution! Some thoughts:

1. We need the same thing for `timestamp` and `timestamptz`, not just for `date`.
2. We should probably also protect conversion the other way around for `timestamp` and `timestamptz`.
3. Instead of hardcoding the dates, maybe we instead compare the 
4. Throwing an error seems the correct thing to do. That's what both DuckDB and postgres do on out-of-bound dates. For the `test_all_types` test, Just add the necessary types to the `exclude` list, and then indeed do what you propose: test the bounds correctly.

shah-nirmit:
> Also, do you know if this also fixes this issue? #629
> 
> If so we should include that in the tests.

With my changes it throws out of bounds Error
```
postgres=# select * from duckdb.query($$ select 'infinity'::timestamp, 'infinity'::date $$);
ERROR:  (PGDuckDB/Duckdb_ExecCustomScan_Cpp) Out of Range Error: The value should be between min and max value (4714-11-24 (BC) <-> 5874897-12-31)
```

JelteF:
> With my changes it throws out of bounds Error

Hmm, I guess that's better than returning something wrong. But it would be good if we actually converted infinity/-infinity correctly for dates and timestamps, because both databases support it.

src/pgduckdb_types.cpp
```diff
@@ -216,6 +235,20 @@ ConvertBinaryDatum(const duckdb::Value &value) {
...
+		                                  duckdb::Date::ToString(pgduckdb::PGDUCKDB_PG_MAX_DATE_VALUE));
+
+	// Special Handling for +/-infinity date values
+	if (!duckdb::Date::IsFinite(date)) {
+		// -infinity value is different for PG date
+		if (date == duckdb::date_t::ninfinity())
+			return date.days - 1;
+		else
+			return date.days;
+	}
```
JelteF, line 250: Instead of doing this `-1` calculation let's use the constants provided by PG/DuckDB. Same for other places in the new code.
```suggestion
	if (date == duckdb::date_t::ninfinity())
		return DATEVAL_NOEND;
	else if (date == duckdb::date_t::infinity())
		return DATEVAL_NOBEGIN;
```

test/regression/expected/timestamp_timestamptz.out
```diff
@@ -0,0 +1,144 @@
...
+-- Cleanup
+set duckdb.force_execution = false;
+DROP TABLE t;
+SELECT * FROM duckdb.query($$ SELECT '4714-11-24 (BC) 00:00:00'::timestamp as timestamp $$);
+          timestamp          
+-----------------------------
+ Mon Nov 24 00:00:00 4714 BC
+(1 row)
+
+SELECT * FROM duckdb.query($$ SELECT '4714-11-23 (BC) 23:59:59'::timestamp as timestamp $$);  -- out of range
```
JelteF, line 47: Let's check at the maximum precision of each of the timestamp types then.
```suggestion
SELECT * FROM duckdb.query($$ SELECT '4714-11-23 (BC) 23:59:59.999999'::timestamp as timestamp $$);  -- out of range
```

test/regression/sql/test_all_types.sql
```diff
@@ -23,5 +23,11 @@ SELECT * exclude(
     fixed_array_of_int_list,
     list_of_fixed_int_array,
     nested_int_array, -- The nested array has different lengths, which is not possible in PG
+    date,
```
JelteF, line 26: Let's add a comment explaining why we exclude these
```suggestion
    -- the min/max values of dates/timestamps overflow in Postgres so selecting these would throw an error
    date,
```

JelteF (changes requested):

JelteF:
Thanks for the  work. This looks almost good, left some mostly cosmetic feedback. CI is also failing  due to a compiler error/warning.

shah-nirmit:
Had to use PG_INT64* Macros for timestamp as specific macros used before were introduces in specific majors only. This will also avoid ugly pg version checks...

src/pgduckdb_types.cpp
```diff
@@ -216,6 +234,20 @@ ConvertBinaryDatum(const duckdb::Value &value) {
 inline Datum
 ConvertDateDatum(const duckdb::Value &value) {
 	duckdb::date_t date = value.GetValue<duckdb::date_t>();
+	if (!ValidDate(date))
+		throw duckdb::OutOfRangeException("The value should be between min and max value (%s <-> %s)",
+		                                  duckdb::Date::ToString(pgduckdb::PGDUCKDB_PG_MIN_DATE_VALUE),
+		                                  duckdb::Date::ToString(pgduckdb::PGDUCKDB_PG_MAX_DATE_VALUE));
+
+	// Special Handling for +/-infinity date values
+	if (!duckdb::Date::IsFinite(date)) {
```
JelteF, line 243: No need for this additional check now. That basically does the two checks below with an `||` inbetween.

JelteF (approved):
Let's use the datum conversion functions for the correct types (even though they are functionally equivalent). Other than that this is now good to merge.

JelteF (approved):

PR #657, opened 2025-03-10
Support the MotherDuck session_hint connection option

JelteF:
This option can be used when using MotherDuck its read-scaling feature to make sure that related PG connections connect to the same MotherDuck instance.


Y-- (approved):
LGTM!

Y-- (approved):
LGTM!

PR #659, opened 2025-03-11
Support time_bucket function

Reminiscent:
tl;dr: support `time_bucket` function.
DuckDB doc: https://duckdb.org/docs/stable/sql/functions/timestamp#time_bucketbucket_width-timestamp-offset
But I found in the duckdb source code there are more functions' definition for the `time_bucket` function: https://github.com/duckdb/duckdb/blob/main/extension/core_functions/scalar/date/time_bucket.cpp#L353-L366

The test cases are also from the duckdb: `test/sql/function/date/test_time_bucket_date.test` and `test/sql/function/timestamp/test_time_bucket_timestamp.test`


sql/pg_duckdb--0.3.0--0.4.0.sql
```diff
@@ -224,6 +224,42 @@ SET search_path = pg_catalog, pg_temp
 AS 'MODULE_PATHNAME', 'duckdb_only_function'
 LANGUAGE C;
 
+CREATE FUNCTION @extschema@.time_bucket(bucket_width interval, ts date)
```
JelteF, line 227: You should make sure these functions also work with `duckdb.unresolved_type`, otherwise `time_bucket` cannot be easily used with the results from the `read_csv`/`read_parquet`/etc functions. See the https://github.com/duckdb/pg_duckdb/blob/main/test/regression/sql/unresolved_type.sql test file for examples on how to test that, you can add some basic tests there.

In theory any argument could be of `duckdb.unresolved_type`. But I think it's fine to only define ones where the `ts` argument is `unresolved_type`. I expect that in all usage the other two arguments would be literals, i.e. they won't be coming from a csv/parquet file. That way we don't make the amount of function definitions explode.

```
CREATE FUNCTION @extschema@.time_bucket(bucket_width interval, ts duckdb.unresolved_type)
```

sql/pg_duckdb--0.3.0--0.4.0.sql
```diff
@@ -224,6 +224,42 @@ SET search_path = pg_catalog, pg_temp
...
+RETURNS date
+SET search_path = pg_catalog, pg_temp
+AS 'MODULE_PATHNAME', 'duckdb_only_function'
+LANGUAGE C;
+
+CREATE FUNCTION @extschema@.time_bucket(bucket_width interval, ts timestamp)
```
JelteF, line 233: I looked at the result from:
```sql
select return_type, parameters, parameter_types from duckdb_functions() where function_name = 'time_bucket';
```
And the `timestamptz` versions of all these function are missing.

test/regression/sql/time_bucket.sql
```diff
@@ -0,0 +1,255 @@
...
+INSERT INTO dates VALUES
+	('10 days', '0794-11-15', '1 week', '0790-11-01'),
+	('10 days', '1700-01-01', '0 days', '1700-01-05'),
+	('1 week', '1832-05-03', '0 days', '1970-05-05'),
+	('10 days', '1897-07-05', '2 days', '1970-06-07'),
+	('2 months', '1946-09-14', '0 months', '1970-07-05'),
+	('2 months', '2000-01-01', '1 month 1 week', '1970-05-01'),
+	('2 year', '2004-05-20', '6 months', '1970-12-31');
+
+select d, time_bucket('3 days'::interval, d) from dates;
```
JelteF, line 15: While extensive tests are usually good. I think the amount of tests here is a bit excessive. We don't really need to ensure that the `time_bucket` function returns correct results for all weird inputs. DuckDB already tests for that. All we really need to test is that you can use this function from Postgres. So I think the only tests that we need is:

1. A single test for each of the variants of the function (so one per combination of arguments).
2. A test where the input to the function comes from `duckdb.query`, so that we test the `unresolved_type` support.

JelteF (changes requested):

JelteF:
Tests are failing for some seemingly unrelated reason but that needs to be addressed too.

JelteF:
Thanks for looking into it a bit, I don't think we can work around that then. I pushed a commit that fixes this by not using DEBUG logging in that test.

sql/pg_duckdb--0.3.0--0.4.0.sql
JelteF: To be clear, this still needs to be done. The rest of the PR looks good now though.

JelteF: That's totally fine, and also what I was expecting to be the case. I mainly wanted to make sure that we both agreed what the PR was still blocked on.

test/regression/sql/time_bucket.sql
```diff
@@ -0,0 +1,52 @@
...
+SELECT time_bucket('12 hours'::interval, '2023-01-15 14:30:00+00'::timestamp with time zone);
+
+SELECT time_bucket('1 day'::interval, '2023-01-15 14:30:00+00'::timestamp with time zone, '6 hours'::interval);
+
+SELECT time_bucket('1 week'::interval, '2023-01-15 14:30:00+00'::timestamp with time zone, '2022-12-25 12:00:00+00'::timestamp with time zone);
+
+SELECT time_bucket('1 day'::interval, '2023-01-15 14:30:00+00'::timestamp with time zone, 'America/New_York');
+
+SELECT time_bucket('1 day'::interval, r['date_col'])
+FROM duckdb.query('SELECT DATE ''2023-01-15'' AS date_col') r;
```
JelteF, line 25: Instead of escaping the quotes by duplicating them let's use the double dollar syntax to create these SQL strings, so escaping is not necessary.
```suggestion
FROM duckdb.query($$ SELECT DATE '2023-01-15' AS date_col $$) r;
```

JelteF (changes requested):
Looks great now, except for one final nitpick on the syntax used in the tests. After fixing that, we can merge this.

PR #663, opened 2025-03-13
Fix leak of PreparedStatement on query failure

### LINKED ISSUE #662: Potential leak of file handles (and memory) in case of query cancel or failure
### What happens?

```
void
Duckdb_BeginCustomScan_Cpp(CustomScanState *cscanstate, EState *estate, int /*eflags*/) {
        ....

	duckdb::unique_ptr<duckdb::PreparedStatement> prepared_query =
	    DuckdbPrepare(duckdb_scan_state->query, explain_prefix);

        ....
   
	duckdb_scan_state->duckdb_connection = pgduckdb::DuckDBManager::GetConnection();
	duckdb_scan_state->prepared_statement = prepared_query.release();
}
```

`DuckdbPrepare` may internally initialize file handles, such as opening a `ParquetReader` for `parquet_scan`. These handles are typically released in `EndScan`. However, in cases of query cancel (e.g., by interrupt) or query failure (e.g., network error), `EndScan` might not be invoked. Postgres memory is reclaimed by resetting memory context, but `prepared_query` and its associated file handles may remain unreleased.

Viable fix: Track `prepared_query` in duckdb `ClientContext` (e.g., , via `PostgresContextState`) and reset the query in `QueryEnd`. This ensures the query is properly cleaned up during txn commit or rollback.

### To Reproduce

/

### OS:

centos

d931980fa0af5dee82183a846de2ea92d60230c9

17.3

### Hardware:

_No response_

### Full Name:

Yuwei XIAO

### Affiliation:

zbyte-inc

Yes

- [x] Yes, I have

- [x] Yes, I have

JelteF:
Fixes #662

### DIFF
--- a/src/pgduckdb_node.cpp
+++ b/src/pgduckdb_node.cpp
@@ -203,60 +203,69 @@ ExecuteQuery(DuckdbScanState *state) {
 static TupleTableSlot *
 Duckdb_ExecCustomScan_Cpp(CustomScanState *node) {
 	DuckdbScanState *duckdb_scan_state = (DuckdbScanState *)node;
-	TupleTableSlot *slot = duckdb_scan_state->css.ss.ss_ScanTupleSlot;
-	MemoryContext old_context;
+	try {
+		TupleTableSlot *slot = duckdb_scan_state->css.ss.ss_ScanTupleSlot;
+		MemoryContext old_context;
 
-	if (ActivePortal && ActivePortal->commandTag == CMDTAG_EXPLAIN) {
-		ExecClearTuple(slot);
-		return slot;
-	}
-
-	bool already_executed = duckdb_scan_state->is_executed;
-	if (!already_executed) {
-		ExecuteQuery(duckdb_scan_state);
-	}
-
-	if (duckdb_scan_state->fetch_next) {
-		duckdb_scan_state->current_data_chunk = duckdb_scan_state->query_results->Fetch();
-		duckdb_scan_state->current_row = 0;
-		duckdb_scan_state->fetch_next = false;
-		if (!duckdb_scan_state->current_data_chunk || duckdb_scan_state->current_data_chunk->size() == 0) {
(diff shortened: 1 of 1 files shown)


JelteF:
Reviewing is easiest if you hide whitespace changes

src/pgduckdb_node.cpp
```diff
@@ -203,60 +203,69 @@ ExecuteQuery(DuckdbScanState *state) {
...
-	return slot;
+		ExecStoreVirtualTuple(slot);
+		return slot;
+	} catch (std::exception &ex) {
+		/*
+		 * In case any error happens we need to still cleanup the scan state,
+		 * otherwise we do not clean up the prepared statement and various
+		 * other DuckDB objects.
+		 */
+		CleanupDuckdbScanState(duckdb_scan_state);
```
Y--, line 266: The usual way to do that is to use RAII. We could add this trivial struct above in the file:

```c++
struct ScopedDuckDBScanState {
  ScopedDuckDBScanState(_state) : state(_state);
  ~ScopedDuckDBScanState() {
     CleanupDuckdbScanState(state);
  }

private:
  DuckdbScanState *state;
}
```

And then at the beginning of this function
```c++
  DuckdbScanState *duckdb_scan_state = (DuckdbScanState *)node;
  ScopedDuckDBScanState(duckdb_scan_state);
```

JelteF: But that would also clean it up on success right? We don't want to do that.

JelteF: Updated the comment to reflect this, because indeed it's not immediately obvious.

Y-- (approved):
Ah yes, I should have read more carefully.
We could make the `ScopedDuckDBScanState` to skip cleanup on success but that would be clunky.

I personally would have:
* renamed `Duckdb_ExecCustomScan_Cpp` to `Duckdb_ExecCustomScan_Cpp_Unsafe`
* implemented `Duckdb_ExecCustomScan_Cpp` with only the try/catch+cleanup

IMHO this makes the whole cleanup logic more apparent (and probably lessen the need for the comment). But happy with that version too

PR #665, opened 2025-03-13
Better COPY support

### LINKED ISSUE #265: Tests for COPY and COPY to local parquet file
We completely broke COPY support without us noticing. That problem is being fixed by #264, but it might happen again if we don't add some tests. For the 0.1.0 release we can manage with manual testing though.

I think the easiest way would be to have a test write locally to parquet and read the resulting file to see that it contains the expected content in parquet format. We currently don't support writing parquet locally but that should be fairly easy to do forcing the COPY to go through DuckDB if it contains a `.parquet` suffix.

### LINKED ISSUE #460: Unable to use COPY TO to write Parquet file to S3 if IAM only enabled on bucket/account  - workaround/fix included - allows for local parquet file writes
### What happens?

If the buckets policy only allows IAM enabled access it is not possible to use COPY TO s3://... to write out a parquet file.

The below diff make it possible to write to a local filesystem (or s3 and IAM working with something like https://github.com/awslabs/mountpoint-s3)

This could be used to support any format duckdb COPY TO supports instead of postgres as this is just passed down to duckdb just by adding "local://" at the start of the file location eg:

`COPY (SELECT * FROM mytable) TO 'local://tmp/test.parquet' (FORMAT 'parquet')`

The diff:

```
# diff pg_duckdb/src/utility/copy.cpp-orig pg_duckdb/src/utility/copy.cpp
4d3
<
30a30,31
> static constexpr char local_filename_prefix[] = "local://";
>
202a204,205
> #include ‹string›
>
>212a216
>             !starts_with(copy_stmt->filename, local_filename_prefix) &&
219a224,229
>       }
>
>         if (starts_with(copy_stmt-›filename, local_filename_prefix)) {
>           std::string filename(copy_stmt->filename);
>           filename.erase(0, 7); // 7 is the length of "local:/" 
>           strcpy (copy_stmt->filename, filename.c_str());

```

### To Reproduce

Setup an s3 bucket with IAM policy only and try to do a COPY TO it

### OS:

Linux

na

17

### Hardware:

_No response_

### Full Name:

Michael Wolman

### Affiliation:

Not applicable

- [X] Yes, I have

- [X] Yes, I have

JelteF:
This adds various improvements to our COPY support:
1. Adds the ability to `COPY ... TO`/COPY ... FROM` tables backed by
   DuckDB.
2. Allow copying to the local filesystem using duckdb its COPY
   functionality. Just like for regular queries DuckDB is only used when
   it's detected that it's needed for execution of the query. Or when
   you configure `duckdb.force_execution`. The way it detects if it it's
   necessary is based on:
   a. `FORMAT PARQUET/JSON` and other options that are only supported by
      DuckDB its COPY implementation.
   b. File extensions like `.parquet`, `.json`, `.parquet.zstd` etc.
   c. Files starting with `s3://`, `gs://` and `r2://`
3. Adds test for COPY support, currently this does not test support for
   blob storage yet.

Fixes #460
Fixes #265

### DIFF
--- a/src/pgduckdb_hooks.cpp
+++ b/src/pgduckdb_hooks.cpp
@@ -53,7 +53,7 @@ ContainsCatalogTable(List *rtes) {
 
 		if (rte->relid) {
 			Relation rel = RelationIdGetRelation(rte->relid);
-			auto is_catalog_table = pgduckdb::IsCatalogTable(rel);
+			bool is_catalog_table = pgduckdb::IsCatalogTable(rel);
 			RelationClose(rel);
 			if (is_catalog_table) {
 				return true;
@@ -156,6 +156,7 @@ IsCatalogTable(Relation rel) {
 	auto namespace_oid = RelationGetNamespace(rel);
 	return namespace_oid == PG_CATALOG_NAMESPACE || namespace_oid == PG_TOAST_NAMESPACE;
 }
+
 bool
 IsAllowedStatement(Query *query, bool throw_error) {
 	int elevel = throw_error ? ERROR : DEBUG4;
--- a/src/pgduckdb_hooks.cpp
+++ b/src/pgduckdb_hooks.cpp
@@ -3,6 +3,7 @@
 #include "pgduckdb/pgduckdb_planner.hpp"
 #include "pgduckdb/pg/transactions.hpp"
 #include "pgduckdb/pgduckdb_xact.hpp"
+#include "pgduckdb/pgduckdb_hooks.hpp"
 #include "pgduckdb/pgduckdb_utils.hpp"
 
 extern "C" {
@@ -51,10 +52,10 @@ ContainsCatalogTable(List *rtes) {
 		}
 
 		if (rte->relid) {
-			auto rel = RelationIdGetRelation(rte->relid);
-			auto namespace_oid = RelationGetNamespace(rel);
+			Relation rel = RelationIdGetRelation(rte->relid);
+			auto is_catalog_table = pgduckdb::IsCatalogTable(rel);
 			RelationClose(rel);
-			if (namespace_oid == PG_CATALOG_NAMESPACE || namespace_oid == PG_TOAST_NAMESPACE) {
+			if (is_catalog_table) {
 				return true;
 			}
 		}
@@ -148,8 +149,15 @@ ContainsFromClause(Query *query) {
 	return query->rtable;
 }
 
-static bool
-IsAllowedStatement(Query *query, bool throw_error = false) {
+namespace pgduckdb {
+
+bool
+IsCatalogTable(Relation rel) {
+	auto namespace_oid = RelationGetNamespace(rel);
+	return namespace_oid == PG_CATALOG_NAMESPACE || namespace_oid == PG_TOAST_NAMESPACE;
+}
+bool
+IsAllowedStatement(Query *query, bool throw_error) {
 	int elevel = throw_error ? ERROR : DEBUG4;
 	/* DuckDB does not support modifying CTEs INSERT/UPDATE/DELETE */
 	if (query->hasModifyingCTE) {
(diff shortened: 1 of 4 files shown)


src/pgduckdb_hooks.cpp
```diff
@@ -51,10 +52,10 @@ ContainsCatalogTable(List *rtes) {
 		}
 
 		if (rte->relid) {
-			auto rel = RelationIdGetRelation(rte->relid);
-			auto namespace_oid = RelationGetNamespace(rel);
+			Relation rel = RelationIdGetRelation(rte->relid);
```
Y--, line 55: Just curious - why the change/explicit type specification?
Should we prefer it over `auto`?

src/pgduckdb_hooks.cpp
```diff
@@ -148,8 +149,15 @@ ContainsFromClause(Query *query) {
...
-static bool
-IsAllowedStatement(Query *query, bool throw_error = false) {
+namespace pgduckdb {
+
+bool
+IsCatalogTable(Relation rel) {
+	auto namespace_oid = RelationGetNamespace(rel);
+	return namespace_oid == PG_CATALOG_NAMESPACE || namespace_oid == PG_TOAST_NAMESPACE;
+}
+bool
```
Y--, line 159: ubernit
```suggestion

bool
```

src/utility/copy.cpp
```diff
@@ -208,32 +213,103 @@ CheckRewritten(List *rewritten) {
...
-
-const char *
-MakeDuckdbCopyQuery(PlannedStmt *pstmt, const char *query_string, struct QueryEnvironment *query_env) {
-	CopyStmt *copy_stmt = (CopyStmt *)pstmt->utilityStmt;
-	if (!copy_stmt->filename) {
-		return nullptr;
+static bool
+ContainsDuckdbCopyOption(CopyStmt *stmt) {
+	foreach_node(DefElem, defel, stmt->options) {
+		if (strcmp(defel->defname, "format") == 0) {
```
Y--, line 251: Our code is littered with `strcmp(str1, str2) == 0`, should we add an `IsEq` helper? (not in this PR ofc)

src/utility/copy.cpp
```diff
@@ -208,32 +213,103 @@ CheckRewritten(List *rewritten) {
...
+		           strcmp(defel->defname, "file_extension") == 0 || strcmp(defel->defname, "per_thread_output") == 0 ||
+		           strcmp(defel->defname, "file_size_bytes") == 0 ||
+		           strcmp(defel->defname, "write_partition_columns") == 0 || strcmp(defel->defname, "array") == 0 ||
+		           strcmp(defel->defname, "compression") == 0 || strcmp(defel->defname, "dateformat") == 0 ||
+		           strcmp(defel->defname, "timestampformat") == 0 || strcmp(defel->defname, "nullstr") == 0 ||
+		           strcmp(defel->defname, "prefix") == 0 || strcmp(defel->defname, "suffix") == 0 ||
+		           strcmp(defel->defname, "compression_level") == 0 || strcmp(defel->defname, "field_ids") == 0 ||
+		           strcmp(defel->defname, "row_group_size_bytes") == 0 ||
+		           strcmp(defel->defname, "row_group_size") == 0 ||
+		           strcmp(defel->defname, "row_groups_per_file") == 0) {
```
Y--, line 270: Would be nice to add a `IsOneOf(str[], str)` to make this easier to read :-)
(and the `IsOneOf` could use `IsEq` :-)

Y-- (approved):
LGTM, just some minor codestyle comments ~ we can change in followup PR

src/pgduckdb_hooks.cpp
JelteF: No good reason, mainly because the type isn't that long to type and it gives a bit more info when looking at the code on github.

src/utility/copy.cpp
JelteF: done

src/utility/copy.cpp
JelteF: done

Y-- (approved):

Y-- (approved):

PR #667, opened 2025-03-14
Add cast from unresolved_type to bytea

JelteF:
This was a CAST that I missed, when adding most other casts in #531. The missing cast was reported on Discord by someone trying to get the spatial extension working in pg_duckdb.

### DIFF
--- a/sql/pg_duckdb--0.3.0--0.4.0.sql
+++ b/sql/pg_duckdb--0.3.0--0.4.0.sql
@@ -244,6 +244,11 @@ CREATE CAST (duckdb.unresolved_type AS bit)
 CREATE CAST (duckdb.unresolved_type AS bit[])
     WITH INOUT;
 
+CREATE CAST (duckdb.unresolved_type AS bytea)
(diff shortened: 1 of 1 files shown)


Y-- (commented):
Can we have a test? :)

JelteF:
> Can we have a test? :)

I thought about that, but I didn't add any tests for the other casts either. So seemed weird to have one only for `bytea`. I cannot think of way that we can somehow break these casts in the future, so I'm not really worried about regressions. So adding tests for all casts seems like a waste of CI time to me.

Y-- (approved):

Y--:
> > Can we have a test? :)
> 
> I thought about that, but I didn't add any tests for the other casts either. So seemed weird to have one only for `bytea`. I cannot think of way that we can somehow break these casts in the future, so I'm not really worried about regressions. So adding tests for all casts seems like a waste of CI time to me.

Ok fair enough. I wonder if there could be a systematic way to loop over every type to pro-actively find these issues?

PR #668, opened 2025-03-14
Revamp MotherDuck settings and secret management

Y--:
# WIP

TODO:
- [ ] Finish code :-)
- [ ] Add PR description
- [ ] Update documentation
- [ ] Add more tests
- [ ] Improve code


src/pgduckdb_userdata_cache.cpp
```diff
@@ -0,0 +1,118 @@
...
+	uint64 version;
+
+	/* Anything below is the actual cache content */
+
+	Oid motherduck_foreign_server_oid;
+	Oid motherduck_postgres_role_oid;
+	Oid motherduck_user_mapping_oid;
+} cache = {};
+
+bool callback_is_configured = false;
```
JelteF, line 39: nit: should have `static` I think. Otherwise it can conflict with the same named variable somewhere else I think? Similarly for the `cache` global above.

```suggestion
static bool callback_is_configured = false;
```

src/pgduckdb_fdw.cpp
```diff
@@ -0,0 +1,372 @@
...
+	assert_has_required_options(options_list, type, context);
+}
+
+Datum
+validate_motherduck_server_fdw(List *options_list, Oid context) {
+	// For now only accept one MotherDuck FDW globally
+	// can be relaxed eventually with https://github.com/duckdb/pg_duckdb/pull/545
+
+	// TODO: take a global lock to make this check
+	assert_has_no_motherduck_foreign_server();
```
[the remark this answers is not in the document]
Y--: Great point. I've changed to `validate` since it's a validator after all :-)

src/pgduckdb_fdw.cpp
```diff
@@ -0,0 +1,372 @@
...
+FindMotherDuckUserMappingOid() {
+	return FindOid(R"(
+		SELECT um.oid
+		FROM pg_foreign_server fs
+		INNER JOIN pg_foreign_data_wrapper fdw ON fdw.oid = fs.srvfdw
+		INNER JOIN pg_user_mapping um ON um.umserver = fs.oid
+		WHERE
+			fdw.fdwname = 'pg_duckdb'
+		AND fs.srvtype = 'motherduck';
+	)");
```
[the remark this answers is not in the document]
Y--: Um looks like I messed something up in my recent rebase... I had that limitation initially but then was using `GetUserId`. Will fix.

src/pgduckdb_userdata_cache.cpp
Y--: I've moved all of these in an anonymous namespace, so no need to be static.
I was actually wondering if I shouldn't also do the same in the "metadata" cache. WDYT?

test/regression/expected/foreign_data_wrapper.out
```diff
@@ -0,0 +1,104 @@
...
+ is_motherduck_enabled 
+-----------------------
+ t
+(1 row)
+
+-- Drop server
+DROP SERVER valid_md_server1 CASCADE;
+NOTICE:  drop cascades to 2 other objects
+DETAIL:  drop cascades to user mapping for user1 on server valid_md_server1
+drop cascades to user mapping for runner on server valid_md_server1
```
Y--, line 70: I don't like that `runner` is the username here. It makes the tests green only on GH. I'll rework it.

src/pgduckdb_background_worker.cpp
```diff
@@ -117,22 +119,50 @@ BackgroundWorkerCheck(duckdb::Connection *connection, int64 *last_activity_count
...
+		return true; // should exit
+	}
+
+	if (!pgduckdb::IsExtensionRegistered()) {
+		return false; // XXX: shouldn't we exit actually?
+	}
+
+	if (!ddb_connection) {
+		try {
+			ddb_connection = pgduckdb::DuckDBManager::CreateConnection();
```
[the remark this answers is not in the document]
Y--: Yeah: it exits the BGW, which is what we want I think.

sql/pg_duckdb--0.3.0--0.4.0.sql
```diff
@@ -452,3 +452,31 @@ CREATE OPERATOR pg_catalog.!~~* (
...
+)
+RETURNS void
+AS 'MODULE_PATHNAME', 'pgduckdb_fdw_validator'
+LANGUAGE C STRICT PARALLEL SAFE;
+
+CREATE FOREIGN DATA WRAPPER pg_duckdb
+  HANDLER pgduckdb_fdw_handler
+  VALIDATOR pgduckdb_fdw_validator;
+
+CREATE FUNCTION duckdb.enable_motherduck(TEXT DEFAULT '::FROM_ENV::')
```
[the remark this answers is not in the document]
JelteF: Looking at the docs you wrote, I'm leaning to: Yes the `default_database` should be in there.

test/regression/expected/foreign_data_wrapper.out
JelteF: It's surprising to me that `runner` is even in here? Why does that name even show up here?

sql/pg_duckdb--0.3.0--0.4.0.sql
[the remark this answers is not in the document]
Y--: I wasn't sure about that actually: won't we end up with this function proxying all the optional parameters?

That being said, it is true that while writing the doc I would have liked to also mention this setting here. I guess I'm fine adding this too but I don't think we should go much further :-)

src/pgduckdb_fdw.cpp
```diff
@@ -0,0 +1,359 @@
...
+    // the role bgw assigns to MD tables (by default server creator/owner)
+    {"tables_owner_role", MD, ForeignServerRelationId},
+    {"background_catalog_refresh_inactivity_timeout", MD, ForeignServerRelationId},
+
+    // a specific token to synchronize MD tables, by default the token of the user mapping which belongs creator/owner
+    {"sync_token", MD, ForeignServerRelationId},
+
+    {"host", MD, ForeignServerRelationId},             // motherduck_host
+    {"port", MD, ForeignServerRelationId},             // motherduck_port
+    {"logging", MD, ForeignServerRelationId},          // motherduck_logging
```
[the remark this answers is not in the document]
Y--: These are not wired anyway, so I'll remove them.

> I think you want to be able to enable logging only for a specific connection.

Yes, that is a good point. The main part will be to enable it after connection has started (which shouldn't be too hard either)

docker/README.md
```diff
@@ -35,7 +35,9 @@ docker run -d -e POSTGRES_PASSWORD=duckdb pgduckdb/pgduckdb:16-main
 And with MotherDuck, it is as simple as:
 ```shell
 $ export MOTHERDUCK_TOKEN=<your personal MD token>
-$ docker run -d -e POSTGRES_PASSWORD=duckdb -e MOTHERDUCK_TOKEN pgduckdb/pgduckdb:16-main -c duckdb.motherduck_enabled=true
+$ docker run -d -e POSTGRES_PASSWORD=duckdb -e MOTHERDUCK_TOKEN pgduckdb/pgduckdb:16-main
+
+# And then run `SELECT duckdb.enable_motherduck();` in psql (cf. below)
```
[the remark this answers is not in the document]
Y--: Doing it in the Dockerfile sounds possible but a bit convoluted. Maybe the extension itself when loaded/starting should do it if it detects that variable? WDYT?

test/regression/expected/foreign_data_wrapper.out
Y--: `runner` is the Linux username in Github test runners.
Since we've created the mapping with `CURRENT_USER`, this is by default the name of Linux user.

docker/README.md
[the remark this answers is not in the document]
JelteF: I definitely don't think we want to automatically do this for production systems, just like we don't use it with `auto` now.

Doing it just for docker, while indeed a bit magic, does make the experience quite a bit nicer. We also do something similar already by already creating the extension:
https://github.com/duckdb/pg_duckdb/blob/d6cfc07ba326d3df6fbf5c4b1ed565ec16aeb9f9/Dockerfile#L66

I think we can do something similar, but you probably need some `psql` metacommands in the startup script like `\if` and `\getenv` to make that work. I think that's worth the nicer try-out experience though.

README.md
```diff
@@ -181,28 +183,33 @@ Note: writes to Azure are not yet supported, please see [the current discussion]
...
+```sql
+CREATE SERVER md_server
+TYPE 'motherduck'
+FOREIGN DATA WRAPPER pg_duckdb
+OPTIONS (default_database '<your database>');
 
-The default MotherDuck database will be easiest to use (see below for details), by default this is `my_db`. If you want to specify which MotherDuck database is your default database, then you can also add the following line to your `postgresql.conf` file:
+-- You may use `::FROM_ENV::` to have the token be read from the environment variable
+CREATE USER MAPPING FOR CURRENT_USER SERVER md_server OPTIONS (token '<your token>')
+```
```
JelteF, line 208: This can be removed from the README now. I think we should only show the convenience method in the readme. All other stuff is "expert" mode and can be covered in the full docs.

test/pycheck/conftest.py
```diff
@@ -62,10 +63,13 @@ def conn(pg):
 @pytest.fixture
 def md_cur(pg, ddb, request):
     """A cursor to a MotherDuck enabled pg_duckdb"""
-    _ = ddb  # silence warning, we only need ddb the ddb
+    _ = ddb  # silence warning, we only need ddb
     test_db = request.node.name.removeprefix("test_")
-    pg.configure("duckdb.motherduck_enabled = true")
-    pg.restart()
+
+    if not pg.md_setup:
```
JelteF, line 69: This `md_setup` boolean isn't needed anymore afaict, because `enable_motherduck()` is a no-op if it's already enabled.

src/pgduckdb_options.cpp
```diff
@@ -296,6 +297,60 @@ DECLARE_PG_FUNCTION(pgduckdb_raw_query) {
...
+
+	// If no token provided, check that token exists in the environment
+	if (token == "::FROM_ENV::" && getenv("MOTHERDUCK_TOKEN") == nullptr && getenv("motherduck_token") == nullptr) {
+		elog(ERROR, "No token was provided and `motherduck_token` environment variable was not set");
+	}
+
+	SPI_connect();
+
+	if (pgduckdb::GetMotherduckForeignServerOid() == InvalidOid) {
+		std::string query = "CREATE SERVER md_server TYPE 'motherduck' FOREIGN DATA WRAPPER pg_duckdb";
```
JelteF, line 321: I think it's nicer to just call this `motherduck`.
```suggestion
		std::string query = "CREATE SERVER motherduck TYPE 'motherduck' FOREIGN DATA WRAPPER pg_duckdb";
```

sql/pg_duckdb--0.3.0--0.4.0.sql
```diff
@@ -452,3 +452,31 @@ CREATE OPERATOR pg_catalog.!~~* (
...
+
+CREATE FUNCTION pgduckdb_fdw_validator(
+    options text[],
+    catalog oid
+)
+RETURNS void
+AS 'MODULE_PATHNAME', 'pgduckdb_fdw_validator'
+LANGUAGE C STRICT PARALLEL SAFE;
+
+CREATE FOREIGN DATA WRAPPER pg_duckdb
```
JelteF, line 475: I think we should call this foreign data wrapper `duckdb`:
```suggestion
CREATE FOREIGN DATA WRAPPER duckdb
```
That matches nicely with our table access method which is also called `duckdb` and our schema etc.

docker/README.md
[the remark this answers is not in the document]
Y--: Done here [`73b61c1` (#668)](https://github.com/duckdb/pg_duckdb/pull/668/commits/73b61c172356e4c1d48a85639bf476586ca3d72a)

README.md
```diff
@@ -181,28 +183,33 @@ Note: writes to Azure are not yet supported, please see [the current discussion]
...
+SELECT duckdb.enable_motherduck('your_access_token');
 ```
 
-NOTE: If you don't want to store the token in your `postgresql.conf`file can also store the token in the `motherduck_token` environment variable and then explicitly enable MotherDuck support in your `postgresql.conf` file:
+NOTE: If you don't want use the token in a query, you may also store the token in the `motherduck_token` environment variable and then explicitly enable MotherDuck support by omitting the token:
 
-```ini
-duckdb.motherduck_enabled = true
+```sql
+SELECT duckdb.enable_motherduck();
```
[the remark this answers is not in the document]
Y--: Done here [`ac8da79` (#668)](https://github.com/duckdb/pg_duckdb/pull/668/commits/ac8da797b3cf866f7324d0b9b4f918ec90086445)

README.md
Y--: Done here [`ac8da79` (#668)](https://github.com/duckdb/pg_duckdb/pull/668/commits/ac8da797b3cf866f7324d0b9b4f918ec90086445)

src/pgduckdb_options.cpp
Y--: IIRC we can't name it `motherduck` because it is already a known symbol, let me try again.

Y--: Well I take that back, it seems to work. Not sure what I did before...

JelteF: maybe you used single quotes around it ;)

JelteF (approved):

YuweiXiao:
Do we support [SCOPE](https://duckdb.org/docs/stable/configuration/secrets_manager.html#creating-multiple-secrets-for-the-same-service-type) option? If not, I could help implement.

JelteF:
It definitely should. It now validates the options with duckdb. If it doesn't in practice then there's a bug.

YuweiXiao:
Manually creating the server with option is working. But we didn't expose it through the udf `create_simple_secret`

```
CREATE SERVER my_server TYPE 's3' FOREIGN DATA WRAPPER duckdb OPTIONS (SCOPE 's3://my-other-bucket');
```

JelteF:
The idea was that create_simple_secret was only for the most simple usecases. And then have more complicated ones use the regular create server syntax. I guess scope might be common enough that it warrants being an argument in create_simple_secret.

YuweiXiao:
By the way, the foreign data wrapper DuckDB is not granted to normal users. So only superusers are able to create secrets, while normal users can only utilize them. Is this by design, or can we loosen the permissions?

Also, normal user can see secrets by `select duckdb.raw_query($$FROM duckdb_secrets()$$)`

PR #669, opened 2025-03-16
Support for STRUCT type

### LINKED ISSUE #599: Support for converting DuckDB STRUCT type to Postgres composite ROW type
### Description

DuckDB has anonymous STRUCT types, with named fields. Postgres doesn't have an equivalent type. The best we can probably do is use the ROW type. This is anonymous compound type, but the fields are not named.

Once supported we can un-exclude the STRUCT types from this test: https://github.com/duckdb/pg_duckdb/blob/02ebeaec1b42619fab806fc881e67de0821b380d/test/regression/expected/test_all_types.out#L15-L16

destrex271:
This introduces a `duckdb.struct` type to postgres which can be used to read the `STRUCT` type in DuckDB.

Fixes #599


src/pgduckdb_types.cpp
```diff
@@ -250,6 +250,12 @@ ConvertTimeTzDatum(const duckdb::Value &value) {
 	return TimeTzADTPGetDatum(result);
 }
 
+inline Datum
+ConvertDuckRowDatum(const duckdb::Value &value) {
```
JelteF, line 254: We should really have a separate `duckdb.struct` type and not reuse the `duckdb.row` type (but you can copy and slightly modify the implementation of `duckdb.row` to create this new `duckdb.struct` type). We need this for one important reason (and possibly a few other less important ones): When `duckdb.row` is part of a select list, we replace it with a `*` before sending it to duckdb. We don't want that for this struct type. Apart from that they can basically behave the same, but we need two separate types to make the distinction between a type that is really a multiple columns under the hood (and which thus needs `*` expansion) versus a type that actually is a single column with a composite type in it.

destrex271: @JelteF do we also need to give the ability to create this type from postgres? like - 

`sql
CREATE TABLE tester(a duckdb.struct);
INSERT INTO TABLE tester(a) VALUES('{a:12, b:12, c:Akshat}');
`

destrex271: Also correct me if I am wrong but we are handling the star exapnsion in `pgduckdb_ddl::DuckdbHandleDDL`, right? 
As far as I understood, `pgduckdb_target_list_contains_unresolved_type_or_row` function helps us in this decision, right?

destrex271:
So I have tried out the following for this:

1. Created a table on motherduck with a column of struct type and populated as follows:
image
image

2. From postgres I am able to perform the following operations:
```sql
-- Select entire content of table
postgres=# SELECT * FROM t1;
         s         
-------------------
 {'v': a, 'i': 42}
(1 row)

-- Use subscripts to access individual elements
postgres=# SELECT s['i'] FROM t1;
 s  
----
 42
(1 row)

-- Perform operations on selected values
postgres=# SELECT (s['i'] + 1) as sum FROM t1;
 sum 
-----
  43
(1 row)
```

As of now we do not have:
 - Create new tables with struct type from postgres in duckdb
 - Create duckdb.struct type in general

src/pgduckdb_ruleutils.cpp
```diff
@@ -101,6 +106,14 @@ pgduckdb_var_is_duckdb_row(Var *var) {
 	return pgduckdb_is_duckdb_row(var->vartype);
 }
 
+bool
+pgduckdb_var_is_duckdb_struct(Var *var) {
```
shah-nirmit, line 110: Question: Where is this utility be used?

destrex271: thanks for pointing that out! we won't need that for `duckdb.struct`. I'll get rid of `pgduckdb_var_is_duckdb_struct` and `pgduckdb_is_duckdb` functions

JelteF:
Thanks for working on this. I made some relatively small improvements. The major one is to return the `duckdb.struct` type from the JSON functions that return a STRUCT. That way we actually have some tests for the subscript operator.

JelteF (approved):
This needs more extensive tests

destrex271:
> This looks good to me now

Thanks!

Issue #601, opened 2025-02-14
Support DuckDB UNION type


JelteF:
### Description

Postgres does not support a type similar to the DuckDB UNION type. So we need to make some concessions here. What we can do is the following:
1. If people actually select a UNION column in the output of their query, we should probably just convert it to its text representation.
2. We should add a `duckdb.union` type, for which we implement the `union_extract` and `union_tag` functions, so people can extract `duckdb.unresolved_type` values from it.

shah-nirmit:
Can you please assign the issue to me , Thanks!

PR #678, opened 2025-03-19
Add Support for DuckDB UNION type

shah-nirmit:
Fixes https://github.com/duckdb/pg_duckdb/issues/601


shah-nirmit:
@JelteF Is this the type of output we expect when selecting a union column?

JelteF:
Yes, that looks great. I think we should at least support the following:

```sql
select union_extract(r['u'], 'str') from duckdb.query($$ select * from tbl1 $$) r;
select union_tag(r['u']) from duckdb.query($$ select * from tbl1 $$) r;
```

This would also be nice, but since duckdb doesn't support indexing a union this way I think that will be impossible to do, without changes to DuckDB.
```sql
select r['u']['str'] from duckdb.query($$ select * from tbl1 $$) r;
```

I think you could in theory implement the following syntax without changes to duckdb. But I think a better way would be to add index support for unions to duckdb itself:
```sql
select r['u']::duckdb.union['str'] from duckdb.query($$ select * from tbl1 $$) r;
```

Finally since postgres doesn't support overloading the `.` operator it's impossible to implement support for the following syntax without changes to DuckDB:
```sql
select r['u'].str from duckdb.query($$ select * from tbl1 $$) r;
```

sql/pg_duckdb--0.3.0--0.4.0.sql
```diff
@@ -452,3 +452,24 @@ CREATE OPERATOR pg_catalog.!~~* (
...
+RETURNS duckdb.unresolved_type
+SET search_path = pg_catalog, pg_temp
+AS 'MODULE_PATHNAME', 'duckdb_only_function'
+LANGUAGE C;
+
+CREATE FUNCTION @extschema@.union_tag(union_col duckdb.unresolved_type)
+RETURNS duckdb.unresolved_type
+SET search_path = pg_catalog, pg_temp
+AS 'MODULE_PATHNAME', 'duckdb_only_function'
+LANGUAGE C;
```
JelteF, line 475: These functions also need a variant for that takes `duckdb.union` as an argument.

sql/pg_duckdb--0.3.0--0.4.0.sql
```diff
@@ -452,3 +452,24 @@ CREATE OPERATOR pg_catalog.!~~* (
...
+CREATE TYPE duckdb.union;
+CREATE FUNCTION duckdb.union_in(cstring) RETURNS duckdb.union AS 'MODULE_PATHNAME', 'duckdb_union_in' LANGUAGE C IMMUTABLE STRICT;
+CREATE FUNCTION duckdb.union_out(duckdb.union) RETURNS cstring AS 'MODULE_PATHNAME', 'duckdb_union_out' LANGUAGE C IMMUTABLE STRICT;
+CREATE TYPE duckdb.union(
+    INTERNALLENGTH = VARIABLE,
+    INPUT = duckdb.union_in,
+    OUTPUT = duckdb.union_out
+);
+
+CREATE FUNCTION @extschema@.union_extract(union_col duckdb.unresolved_type, tag text)
```
JelteF, line 465: Needs some tests for these functions.

shah-nirmit:
> Yes, that looks great. I think we should at least support the following:
> 
> ```sql
> select union_extract(r['u'], 'str') from duckdb.query($$ select * from tbl1 $$) r;
> select union_tag(r['u']) from duckdb.query($$ select * from tbl1 $$) r;
> ```
> 
> This would also be nice, but since duckdb doesn't support indexing a union this way I think that will be impossible to do, without changes to DuckDB.
> 
> ```sql
> select r['u']['str'] from duckdb.query($$ select * from tbl1 $$) r;
> ```
> 
> I think you could in theory implement the following syntax without changes to duckdb. But I think a better way would be to add index support for unions to duckdb itself:
> 
> ```sql
> select r['u']::duckdb.union['str'] from duckdb.query($$ select * from tbl1 $$) r;
> ```
> 
> Finally since postgres doesn't support overloading the `.` operator it's impossible to implement support for the following syntax without changes to DuckDB:
> 
> ```sql
> select r['u'].str from duckdb.query($$ select * from tbl1 $$) r;
> ```

I think the first 2 syntax are the most straigth forward , and have implemented that and updated in PR description...

src/pgduckdb_options.cpp
```diff
@@ -559,4 +559,13 @@ DECLARE_PG_FUNCTION(duckdb_only_function) {
 	elog(ERROR, "Function '%s' only works with DuckDB execution", function_name);
 }
 
+DECLARE_PG_FUNCTION(duckdb_union_in) {
+	elog(ERROR, "Creating the duckdb.union type is not supported");
+}
+
+DECLARE_PG_FUNCTION(duckdb_union_out) {
+	// TODO : Is this correct??
```
JelteF, line 567: Yeah this is correct

JelteF (changes requested):

sql/pg_duckdb--0.3.0--0.4.0.sql
shah-nirmit: Added

sql/pg_duckdb--0.3.0--0.4.0.sql
shah-nirmit: Added

JelteF (approved):

PR #680, opened 2025-03-20
Fix issue with too long schema names breaking MotherDuck sync

JelteF:
If you have a database+schema that would result in a `ddb$` schema with
more than 63 characters it would completely break the MotherDuck
background syncing, due to an error-restart-loop. This changes the code
to skip that schema.

It also codes a bit more defensively in the place where it was also
previously throwing the error continuously. If the schema cannot be found
when we try to add a dependency for it, we say that we failed to create
the schema and skip it instead of getting into this error loop.

### DIFF
--- a/src/pgduckdb_background_worker.cpp
+++ b/src/pgduckdb_background_worker.cpp
@@ -818,7 +818,7 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
 		 */
 		schema_oid = get_namespace_oid(postgres_schema_name, true);
 		if (schema_oid == InvalidOid) {
-			elog(WARNING, "Failed to create schema %s for unknown reason, skipping sync", postgres_schema_name);
+			elog(WARNING, "Failed to create schema '%s' for unknown reason, skipping sync", postgres_schema_name);
 			RollbackAndReleaseCurrentSubTransaction();
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


src/pgduckdb_background_worker.cpp
```diff
@@ -803,15 +811,18 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
...
-
 	if (!is_default_db) {
 		/*
 		 * For ddb$ schemas we need to record a dependency between the schema
 		 * and the extension, so that DROP EXTENSION also drops these schemas.
 		 */
-		schema_oid = get_namespace_oid(postgres_schema_name, false);
+		schema_oid = get_namespace_oid(postgres_schema_name, true);
+		if (schema_oid == InvalidOid) {
+			elog(WARNING, "Failed to create schema %s for unknown reason, skipping sync", postgres_schema_name);
```
Y--, line 821:
```suggestion
			elog(WARNING, "Failed to create schema '%s' for unknown reason, skipping sync", postgres_schema_name);
```

Y-- (approved):

src/pgduckdb_background_worker.cpp
```diff
@@ -734,6 +734,14 @@ GrantAccessToSchema(const char *postgres_schema_name) {
 
 static bool
 CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
+	/* -1 is for the NULL terminator */
+	if (strlen(postgres_schema_name) > NAMEDATALEN - 1) {
+		ereport(WARNING,
+		        (errmsg("Skipping sync of MotherDuck schema %s because its name is too long", postgres_schema_name),
```
JelteF, line 740:
```suggestion
		        (errmsg("Skipping sync of MotherDuck schema '%s' because its name is too long", postgres_schema_name),
```

Y-- (approved):

PR #688, opened 2025-03-21
Support `INSERT ... INTO ... SELECT ...` for Postgres table

### LINKED ISSUE #30: Writing into Postgres tables
We should enable COPY command to be able to write to remote S3 storage. 

Copying TO remote storage should be possible  by passing query directly to duckdb execution while in other direction (FROM) more complex logic will be needed.

- [x] COPY .. TO
- [ ] COPY .. FROM

YuweiXiao:
Partially fixes #30


src/pgduckdb_planner.cpp
```diff
@@ -33,7 +33,23 @@ extern "C" {
 duckdb::unique_ptr<duckdb::PreparedStatement>
 DuckdbPrepare(const Query *query, const char *explain_prefix) {
 	Query *copied_query = (Query *)copyObjectImpl(query);
-	const char *query_string = pgduckdb_get_querydef(copied_query);
+	const char *query_string;
+	if (copied_query->commandType == CMD_INSERT) {
+		ListCell *l;
+		RangeTblEntry *select_rte = NULL;
+		foreach(l, copied_query->rtable) {
+			RangeTblEntry *rte = (RangeTblEntry *)lfirst(l);
```
JelteF, line 41:
```suggestion
		RangeTblEntry *select_rte = NULL;
		foreach_node(RangeTblEntry, rte, copied_query->rtable) {
```

src/pgduckdb_hooks.cpp
```diff
@@ -169,10 +182,16 @@ IsAllowedStatement(Query *query, bool throw_error) {
 	/* We don't support modifying statements on Postgres tables yet */
 	if (query->commandType != CMD_SELECT) {
 		RangeTblEntry *resultRte = list_nth_node(RangeTblEntry, query->rtable, query->resultRelation - 1);
-		if (!::IsDuckdbTable(resultRte->relid)) {
+		if (::IsHeapTable(resultRte->relid) && query->commandType == CMD_INSERT) {
+			if (query->onConflict || query->returningList) {
+				elog(elevel, "DuckDB does not support INSERT ON CONFLICT or RETURNING on Postgres tables");
+				return false;
+			}
```
JelteF, line 189: Why not allow these? Do they not work? I would expect that if we insert using postgres its functionality that thees things would automatically work.

JelteF: To be clear, not saying that this should be implemented in the first version of this feature. But I'd at least like a code comment explaining why we do not support it.

src/pgduckdb_hooks.cpp
```diff
@@ -169,10 +182,16 @@ IsAllowedStatement(Query *query, bool throw_error) {
 	/* We don't support modifying statements on Postgres tables yet */
 	if (query->commandType != CMD_SELECT) {
 		RangeTblEntry *resultRte = list_nth_node(RangeTblEntry, query->rtable, query->resultRelation - 1);
-		if (!::IsDuckdbTable(resultRte->relid)) {
+		if (::IsHeapTable(resultRte->relid) && query->commandType == CMD_INSERT) {
```
JelteF, line 185: Why only heap tables? I'd expect this should logic would work fine for most table access method, because they implement the table access method interface.

JelteF (changes requested):
CI is failing in many places, because `duckdb.force_execution` is `true` in our tests. Apparently something breaks when selecting from `generate_series()` in duckdb for a PG INSERT.

src/pgduckdb_hooks.cpp
YuweiXiao: Ah, block them since I haven't tested them yet. I've done some initial testing, and RETURNING works out of box while onConflict doesn't. Will dig deeper into the WHY and add comments to clarify.

src/pgduckdb_hooks.cpp
YuweiXiao: YES, you are right! This is becoming much more exciting if we get the whole thing functional.

YuweiXiao:
> I think that would be worked around by #583, but then the opposite would hold (i.e. inserting into a varchar column woudl fail). So I think we need some logic to make sure that the types match against eachother, and maybe add some code to convert the type that we get back from DuckDB to the type that PG wants for the INSERT. I think that would also fix the `generate_series()` issue that all of the tests are failing on currently.

YES, the type check and conversion are substantial. I manually added an `Expr` to handle the coerce and it seems working so far (e.g., `varchar` and `generate_series` issue). I'll add more test cases to ensure if I haven't missed (or broken) anything, and code re-org (& cleanup) are also in progress. Thanks for the feedback, @JelteF !

YuweiXiao:
The MR is ready for review~ @JelteF Please take a look when you have a chance.

Currently, the regression is failing for `hugeint_numeric`. The issue seems related to the following behavior, which is reproducible in the `main` branch. Do you have any insights on this?

```
select * from duckdb.query($$ SELECT a FROM (VALUES (32942348563242.111222333444555666777888::numeric(38,24)), (NULL::numeric), (123456789.000000000000000000000001::numeric(38,24))) t(a) $$);

                   a
----------------------------------------
 32942348563242.11122233344455566677789  (correct one should be `32942348563242.111222333444555666777888`

      123456789.00000000000000000000000 (correct one should be 123456789.000000000000000000000001)
(3 rows)
```

src/pgduckdb_hooks.cpp
YuweiXiao: ON CONFLICT is also working

include/pgduckdb/pgduckdb_planner.hpp
```diff
@@ -14,3 +14,4 @@ extern bool duckdb_explain_ctas;
 PlannedStmt *DuckdbPlanNode(Query *parse, const char *query_string, int cursor_options, ParamListInfo bound_params,
                             bool throw_error);
 duckdb::unique_ptr<duckdb::PreparedStatement> DuckdbPrepare(const Query *query, const char *explain_prefix = NULL);
+bool IsAllowedPostgresInsert(Query *query, bool throw_error = false);
```
JelteF, line 17: Seems like you don't have https://editorconfig.org/ installed in your editor. Could you do that and re-save this file to add a trailing newline.

src/pgduckdb_planner.cpp
```diff
@@ -6,34 +6,107 @@
...
+			select_rte = rte;
+		}
+	}
+
+	if (!select_rte) {
+		elog(elevel, "DuckDB does not support INSERT without a subquery");
+		return false;
+	}
+	
+	/* TODO: Checking supported target column types */
```
JelteF, line 69: You wanna implement this TODO still?

YuweiXiao: ah, the TODO comment can be removed. The follow-up code handles the checking through blacklist

include/pgduckdb/pgduckdb_planner.hpp
YuweiXiao: sure

JelteF:
> Currently, the regression is failing for hugeint_numeric. The issue seems related to the following behavior, which is reproducible in the main branch. Do you have any insights on this?

I think you can probably fix the test by changing the plain `NULL` to `NULL::NUMERIC(38, 24)`. It seems at least that that solves the issue when running your reproducing example in the duckdb CLI.

src/pgduckdb_planner.cpp
```diff
@@ -6,34 +6,107 @@
...
+	Relation rel = RelationIdGetRelation(target_rel->relid);
+	TupleDesc target_desc = RelationGetDescr(rel);
+	bool ret = true;
+	for (int i = 0; i < target_desc->natts; i++) {
+		Form_pg_attribute attr = TupleDescAttr(target_desc, i);
+		if (attr->attisdropped) {
+			continue;
+		}
+
+		if (attr->atttypid == BYTEAOID || attr->atttypid == XMLOID || pgduckdb::pg::IsDomainType(attr->atttypid) ||
```
JelteF, line 79: Instead of hardcoding some unsupported types, I think we should instead determine the allowed types. Also, why are these types unsupported? Because we do support BYTEA as a type for regular duckdb execution, so that seems surprising that it doesn't work here.

test/regression/expected/postgres_insert_select.out
```diff
@@ -0,0 +1,103 @@
...
+-- case: INSERT INTO PARTITION TABLE
+CREATE TABLE tbl (a int PRIMARY KEY, b text) PARTITION BY RANGE (a);
+CREATE TABLE tbl_p1 PARTITION OF tbl FOR VALUES FROM (1) TO (10);
+CREATE TABLE tbl_p2 PARTITION OF tbl FOR VALUES FROM (10) TO (20);
+INSERT INTO tbl select r['a']::int, r['b'] from duckdb.query($$ SELECT 1 a, 'abc' b $$) r;
+INSERT INTO tbl select r['a']::int, r['b'] from duckdb.query($$ SELECT 11 a, 'def' b $$) r;
+INSERT INTO tbl select r['a']::int, r['b'] from duckdb.query($$ SELECT 21 a, 'ghi' b $$) r;
+ERROR:  no partition of relation "tbl" found for row
+DETAIL:  Partition key of the failing row contains (a) = (21).
+SELECT * FROM tbl ORDER BY a LIMIT 5;
```
JelteF, line 42: Would be good to also check the contents of each partition, to double check that values go to the correct partition.

src/pgduckdb_planner.cpp
```diff
@@ -6,34 +6,107 @@
...
+	/* Checking supported INSERT types */
+	RangeTblEntry *select_rte = NULL;
+	foreach_node(RangeTblEntry, rte, query->rtable) {
+		if (rte->rtekind == RTE_SUBQUERY) {
+			select_rte = rte;
+		}
+	}
+
+	if (!select_rte) {
+		elog(elevel, "DuckDB does not support INSERT without a subquery");
```
JelteF, line 65: There are no tests for this error case afaict

src/pgduckdb_planner.cpp
```diff
@@ -6,34 +6,107 @@
...
 #include "pgduckdb/pgduckdb_node.hpp"
 #include "pgduckdb/vendor/pg_list.hpp"
 #include "pgduckdb/utility/cpp_wrapper.hpp"
 #include "pgduckdb/pgduckdb_types.hpp"
 
+bool
+IsAllowedPostgresInsert(Query *query, bool throw_error) {
+	int elevel = throw_error ? ERROR : DEBUG4;
+	if (query->commandType != CMD_INSERT) {
+		elog(elevel, "DuckDB only supports INSERT/SELECT on Postgres tables");
```
JelteF, line 46: There are no tests for this error case afaict

test/regression/expected/postgres_insert_select.out
```diff
@@ -0,0 +1,103 @@
...
+ a |  b  
+---+-----
+ 1 | qux
+ 2 | qux
+ 3 | qux
+(3 rows)
+
+DROP TABLE tbl;
+-- case: ON CONFLICT DO NOTHING
+CREATE TABLE tbl (a int PRIMARY KEY, b text);
```
JelteF, line 91: What happens with a `SERIAL` column or a column with `GENERATED ALWAYS AS IDENTITY`.

JelteF:
To set expectations: I think this is a very cool feature and your implementation looks like the right direction. But it is relatively complex and impactful, so I would like to give it an appropriate round of testing and "baking" (aka testing by merging it and having it on the `main` branch for a while). Since we want to release 1.0 somewhat soon, I don't think it's realistic to still do that before the 1.0 release. So I added this to the new 1.1 milestone instead. I also won't spend time reviewing this PR in depth until then.

YuweiXiao:
> To set expectations: I think this is a very cool feature and your implementation looks like the right direction. But it is relatively complex and impactful, so I would like to give it an appropriate round of testing and "baking" (aka testing by merging it and having it on the `main` branch for a while). Since we want to release 1.0 somewhat soon, I don't think it's realistic to still do that before the 1.0 release. So I added this to the new 1.1 milestone instead. I also won't spend time reviewing this PR in depth until then.

Totally agree, 1.1 makes sense. btw, any target date for 1.0 release?

JelteF:
> any target date for 1.0 release?

The current plan is to do it quickly after the DuckDB 1.3.0 release, so early May. But dates may obviously slip.

src/pgduckdb_planner.cpp
YuweiXiao: Changed the behavior to whitelist. As long as the Postgres type can be converted to a DuckDB type using `ConvertPostgresToDuckColumnType`, it will be allowed.

test/regression/expected/postgres_insert_select.out
YuweiXiao: Fixed

src/pgduckdb_planner.cpp
YuweiXiao: Added

src/pgduckdb_planner.cpp
YuweiXiao: Added

JelteF (approved):

JelteF:
@YuweiXiao thanks for all your work on this. It took a while, but it's merged now!

PR #705, opened 2025-04-16
Remove unnecessary variables

JelteF:
Suggested in final review comments on #628, which were ignored because
of GitHubs auto-merge feature.


Y-- (approved):
Thanks!

PR #706, opened 2025-04-16
Fix type mismatch issues with CTAS and materialized views

JelteF:
DuckDB and Postgres can return different types for the same queries.
This can confuse Postgres in various ways when these queries are
involved in CTAS or materialized views. This fixes a bunch of issues
related to that and adds appropriate regression tests.

This includes a fix for the materialized view issue that was found in #583

### DIFF
--- a/src/pgduckdb_ddl.cpp
+++ b/src/pgduckdb_ddl.cpp
@@ -4,6 +4,7 @@
 #include "pgduckdb/pgduckdb_guc.h"
 #include "pgduckdb/pgduckdb_ddl.hpp"
 #include "pgduckdb/pgduckdb_hooks.hpp"
+#include "pgduckdb/pgduckdb_planner.hpp"
 
 extern "C" {
 #include "postgres.h"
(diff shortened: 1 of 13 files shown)


src/pgduckdb_ddl.cpp
```diff
@@ -246,20 +248,41 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
...
 		Query *original_query = castNode(Query, stmt->query);
 
-		// For cases where Postgres does not usually plan the query, we need
-		// to do hacky things if the targetlist contains duckdb.unresolved_type
-		// or duckdb.row columns. In those cases we want to run the query
-		// through duckdb to get the actual result types for these queries,
-		// which won't happen automatically if the query is not planned by
-		// Postgres. So we will manually do it here.
+		// For cases where Postgres does not usually plan the query, sometimes
+		// still need to. Specifically for these cases:
```
Y--, line 252:
```suggestion
		// we still need to. Specifically for these cases:
```

src/pgduckdb_ddl.cpp
```diff
@@ -282,7 +305,18 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
...
+		/*
+		 * We temporarily set the force_execution GUC to true, so that we can
+		 * plan the query in DuckDB. This is needed for a CTAS into a DuckDB
+		 * table, where the query itself does not require DuckDB execution.
+		 * If we don't set force_execution to true, the query will be planned
+		 * using Postgres, but then later executed using DuckDB when we
+		 * actually write into the DuckDB table.
+		 */
+		auto save_nestlevel = NewGUCNestLevel();
+		SetConfigOption("duckdb.force_execution", "true", PGC_USERSET, PGC_S_SESSION);
```
Y--, line 317: Do we have to go through the GUC? Why not using a different "force_execution" that would override it? Is it to limit the changes in this PR?

JelteF: Fixed now by calling `DuckdbPlanNode` directly.

Y-- (approved):
Nice one!

PR #707, opened 2025-04-16
Improve MD test user creation

Y--:
In order to correctly test https://github.com/duckdb/pg_duckdb/pull/545 we will need to have multiple test users.

This PR paves the way to generate as many MotherDuck test user as we want.

I am considering removing the support for `MOTHERDUCK_TEST_TOKEN`, but I have kept it for now.

The main change is now we can run the test suite with a token that can generate test users and it will generates one and use it through the test suite.

### DIFF
--- a/test/pycheck/conftest.py
+++ b/test/pycheck/conftest.py
@@ -4,6 +4,7 @@ import pytest
 import duckdb
 
 from .utils import Postgres, Duckdb
+from .motherduck_token_helper import create_test_user
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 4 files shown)


JelteF (changes requested):
So how can I now run the motherduck tests locally? Please update the "Testing" section in `CONTRIBUTING.md` accordingly, because setting MOTHERDUCK_TEST_TOKEN doesn't work anymore.

Y--:
> So how can I now run the motherduck tests locally? Please update the "Testing" section in `CONTRIBUTING.md` accordingly, because setting MOTHERDUCK_TEST_TOKEN doesn't work anymore.

As mentioned in the description, setting `MOTHERDUCK_TEST_TOKEN` is still a valid way, but I had broken the skip test, this is now fixed.

To be clear, this PR mostly brings a new mechanism to generate users.

JelteF (approved):

PR #724, opened 2025-04-23
Check overflows when appending values

Y--:
Address two small FIXMEs

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -21,6 +21,7 @@ extern "C" {
 #include "miscadmin.h"
 #include "access/tupdesc_details.h"
 #include "catalog/pg_type.h"
+#include "common/int.h"
 #include "executor/tuptable.h"
 #include "utils/array.h"
 #include "utils/builtins.h"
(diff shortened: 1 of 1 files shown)


src/pgduckdb_types.cpp
```diff
@@ -962,7 +971,7 @@ struct PostgresArrayAppendState {
 	idx_t count = 0;
 
 public:
-	idx_t expected_values = 1;
+	uint64 expected_values = 1;
```
JelteF, line 974:
```suggestion
	uint64_t expected_values = 1;
```

Y--: IIRC uint64 != uint64_t (at least on Mac) so it creates compiler errors. I'll check asap

JelteF: Weird

JelteF: Ugh, let's leave it as is then...

JelteF: This is changed in PG18 btw, where the Postgres types will simply be aliases for the `_t` variants: https://github.com/postgres/postgres/commit/962da900ac8f0927f1af2fd811ca67fa163c873a

Y--: > Ugh, let's leave it as is then...

You mean keep it an `idx_t`? That won't work either since `idx_t` is also an `uint64_t` :-)

JelteF: So `uint64`

PR #725, opened 2025-04-23
Fully build connection string in C++

Y--:


src/pgduckdb_duckdb.cpp
```diff
@@ -38,25 +38,25 @@ extern "C" {
...
 
 	while (*str) {
 		char c = *str++;
 		if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '-' || c == '_' ||
 		    c == '.' || c == '~') {
-			appendStringInfoChar(&buf, c);
+			oss << c;
 		} else {
-			appendStringInfo(&buf, "%%%02X", (unsigned char)c);
+			oss << std::setfill('0') << std::setw(2) << std::hex << static_cast<unsigned char>(c);
```
JelteF, line 54: Doesn't seem to do the same. i.e. it doesn't put a `%` in front. Sounds like we should probably add a simple test for this behaviour.

JelteF (approved):

PR #727, opened 2025-04-23
Small Python test improvements

Y--:
* add helper to silence output
* re-use MD user across all tests
* add a `wait_until` helper

### DIFF
--- a/test/pycheck/utils.py
+++ b/test/pycheck/utils.py
@@ -1,11 +1,8 @@
 import subprocess
-from contextlib import closing, contextmanager, redirect_stdout
+from contextlib import closing, contextmanager, asynccontextmanager, suppress
 from pathlib import Path
 
-from contextlib import asynccontextmanager
-
 import asyncio
-import io
 import os
 import platform
 import re
@@ -140,6 +137,18 @@ def create_duckdb(db_name, token):
     return con
 
 
+def loop_until(timeout=5, error_message="Did not complete in time"):
+    """
+    Loop until the timeout is reached. If the timeout is reached, raise an
+    exception with the given error message.
+    """
+    end = time.time() + timeout
+    while time.time() < end:
+        yield
+        time.sleep(0.1)
+    raise TimeoutError(error_message)
+
+
 PG_MAJOR_VERSION = get_pg_major_version()
 
 PG_BIN_DIR = get_bin_dir()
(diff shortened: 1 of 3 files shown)


test/pycheck/utils.py
```diff
@@ -258,7 +270,18 @@ def __getattr__(self, name):
...
-class Cursor:
+class OutputSilencer:
+    @contextmanager
+    def silent_logs(self):
+        """Set the log level to FATAL, so that we don't get any logs"""
+        buffer = io.StringIO()
+        with redirect_stdout(buffer):
+            self.sql("SET log_min_messages TO FATAL")
+            yield
+            self.sql("RESET log_min_messages")
```
JelteF, line 281: I think it would be good to clean up using a `finally`.
(don't apply suggestion verbatim it mixes tabs and spaces it seems)
```suggestion
			try:
            	yield
            finally:
            	self.sql("RESET log_min_messages")
```

test/pycheck/utils.py
```diff
@@ -258,7 +270,18 @@ def __getattr__(self, name):
 AsyncConnection = psycopg.AsyncConnection
 
 
-class Cursor:
+class OutputSilencer:
+    @contextmanager
+    def silent_logs(self):
+        """Set the log level to FATAL, so that we don't get any logs"""
+        buffer = io.StringIO()
+        with redirect_stdout(buffer):
```
JelteF, line 278: Is this still needed?

test/pycheck/utils.py
```diff
@@ -281,31 +304,48 @@ def dsql(self, query, **kwargs):
...
+                **kwargs,
+            ),
+            f"Table {table_name} did not appear in time",
+            timeout=timeout,
+        )
 
-    def wait_until_schema_exists(self, schema_name, timeout=5):
+    def wait_until_schema_exists(self, schema_name, timeout=5, **kwargs):
+        self.wait_until(
+            lambda: self.try_sql(
```
JelteF, line 321: nit: Instead of waiting for a predicate, how about we wait until no exception occurs. Then we can make this a context manager, instead of having to use lambdas. (in case you haven't noticed. I much prefer context managers over lambdas in python)

test/pycheck/utils.py
```diff
@@ -552,13 +592,14 @@ def cleanup_users(self):
...
 
     def cleanup_schemas(self):
-        for dbname, schema in self.schemas:
-            self.sql(
-                sql.SQL("DROP SCHEMA IF EXISTS {} CASCADE").format(
-                    sql.Identifier(schema)
-                ),
-                dbname=dbname,
-            )
+        with self.silent_logs():
```
JelteF, line 595: What errors/warnings is this silencing? Would be good to add a comment, because I wouldn't expect the below code to throw an error/warning.

JelteF (approved):

JelteF (approved):

Issue #716, opened 2025-04-21
0.4.0 create_simple_secret doesn't handle new url_style parameter


adeel-ansari:
### What happens?

I was trying v0.4.0 to use the url_style parameter that has been added to support min.io.  However I noticed that the create_simple_secret function has not yet incorporated some of the newer parameters (url_style, endpoint).  I think in v0.4.0 we have to use the create_simple_secret function and are no longer inserting to the secrets table.  Could we please incorporate the newer parameters?

### To Reproduce

I was trying something like this;

```
SELECT duckdb.create_simple_secret(
    type 'S3',
    endpoint 'myendpoint.org.ca',
    key_id 'mykeyid',
    secret 'mysecret',
    region 'us-east-1',
    url_style true
);
```

### OS:

Linux

### pg_duckdb Version (if built from source use commit hash):

0.4.0

### Postgres Version (if built from source use commit hash):

17

### Full Name:

Adeel Ansari

### Affiliation:

CAMH

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have not tested with any build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

adeel-ansari:
Thanks, that worked @Y--

Y--:
@adeel-ansari Thanks again for the great feedback. We've improved the `create_simple_secret` helper and you should now be able to write:

```sql
SELECT duckdb.create_simple_secret(
    type := 'S3',
    endpoint := 'myendpoint.org.ca',
    key_id := 'mykeyid',
    secret := 'mysecret',
    region := 'us-east-1',
    url_style := 'path'
);
```

PR #728, opened 2025-04-23
Improve `create_simple_secret` helper

Y--:
As reported in https://github.com/duckdb/pg_duckdb/issues/716, some parameters were not easily available with the `create_simple_secret` helper function.

This PR makes the following important changes:
* moves the `region` parameter *after* the `session_token`
* names the arguments of `create_simple_secret`
* adds `url_style`, `provider` and `endpoint` named parameters

Fixes https://github.com/duckdb/pg_duckdb/issues/716

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,11 +148,14 @@ Querying data stored in Parquet, CSV, JSON, Iceberg and Delta format can be done
 
 	```sql
 	SELECT duckdb.create_simple_secret(
-		'S3',                -- Type (S3, GCS, R2)
-		'access_key_id',     -- Key Id
-		'secret_access_key', -- Secret
-		'session_token',     -- Session Token (optional)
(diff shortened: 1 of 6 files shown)


JelteF (commented):
Code looks okay, but CI is failing in a way that's not immediately obvious to me why. Would be good to double check that it's not indicating a problem, before updating the output.

JelteF (approved):
Approved. Feel free to merge merge if you're happy with the test fix.

Y--:
> Code looks okay, but CI is failing in a way that's not immediately obvious to me why. Would be good to double check that it's not indicating a problem, before updating the output.

Yeah somehow on my Mac when you don't provide the `session_token`, DuckDB still does display "redacted". I've verified that it is indeed not set to some weird/garbage value manually. It sounds very minor to me so I think it's fine like that.

PR #729, opened 2025-04-23
Add support for changing secrets public options

Y--:
Follow-up of https://github.com/duckdb/pg_duckdb/pull/697: add support for `ALTER SERVER` syntax

### DIFF
--- a/src/pgduckdb_ddl.cpp
+++ b/src/pgduckdb_ddl.cpp
@@ -151,6 +151,7 @@ DropTablePreCheck(DropStmt *stmt) {
 }
 
 static void DuckdbHandleCreateForeignServerStmt(Node *parsetree);
+static void DuckdbHandleAlterForeignServerStmt(Node *parsetree);
 static void DuckdbHandleCreateUserMappingStmt(Node *parsetree);
 
 static void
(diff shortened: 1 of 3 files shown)


src/pgduckdb_ddl.cpp
```diff
@@ -392,12 +393,34 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
...
 	}
 }
 
+static void DuckdbHandleAlterForeignServerStmt(Node *parsetree) {
+	pgduckdb::CurrentServerType = nullptr;
+	// Propagate the "TYPE" to the server options, don't validate it here though
+	auto stmt = castNode(AlterForeignServerStmt, parsetree);
+	auto server = GetForeignServerByName(stmt->servername, false);
+
+	// XXX - should we cache this?
```
Y--, line 411: @JelteF would love your opinion on this one.

JelteF: Nah, `get_foreign_data_wrapper_oid` uses `GetSysCacheOid1` internally. So it's already using a cache. The main reason we have our own cache for some things is so that we can access it in a thread-safe way, which we don't need here.

Even without all that, this isn't a hot code path. So caching in general seems like a waste.

PR #738, opened 2025-04-24
Don't initialize DuckDB just to invalidate secrets

Y--:


include/pgduckdb/pgduckdb_duckdb.hpp
```diff
@@ -24,6 +24,11 @@ class DuckDBManager {
 		return manager_instance;
 	}
 
+	static inline DuckDBManager*
+	Find() {
```
JelteF, line 28: nit: I don't think it's clear from name exactly what this does. It seems quite rare that we'd need to do this, so I think a longer a bit more descriptive name would be nice. Maybe `GetIfInitialized()`.

JelteF (approved):
Code looks reasonable, only some decision around method naming.

PR #740, opened 2025-04-24
[MD] Add read scaling test

Y--:


src/pgduckdb_background_worker.cpp
```diff
@@ -475,9 +481,11 @@ TriggerActivity(void) {
 	}
 
 	SpinLockAcquire(&BgwShmemStruct->lock);
-	auto state = GetState();
-	state->activity_count++;
-	SetLatch(state->latch);
+	auto state = FindState();
+	if (state) {
```
Y--, line 485: Despite what I said (thought) in last PR, this can still happen. I can't reproduce manually, but the tests clearly hits it.

src/pgduckdb_utils.cpp
```diff
@@ -36,7 +36,7 @@ AppendEscapedUri(std::ostringstream &oss, const char *str) {
 		    c == '.' || c == '~') {
 			oss << c;
 		} else {
-			oss << '%' << std::setfill('0') << std::setw(2) << std::hex << std::uppercase << static_cast<int>(c);
+			oss << '%' << std::setfill('0') << std::setw(2) << std::hex << std::uppercase << static_cast<int>(static_cast<unsigned char>(c));
```
Y--, line 39: Fix to handle utf8 chars

src/pgduckdb_background_worker.cpp
JelteF: Yeah I hit this yesterday when running manually. It happens when we trigger activity without the background worker having been started. The worst part is that was that it would then hold the spinlock forever. So let's also unlock the spinlock before we throw the error in GetState.

test/pycheck/motherduck_test.py
```diff
@@ -56,6 +56,37 @@ def test_md_create_table(md_cur: Cursor, ddb):
         md_cur.sql("CREATE TABLE t4(a int) USING heap")
 
 
+def test_md_read_scaling(pg: Cursor, ddb, default_db_name, md_test_user):
+    weird_db_name = "some 19 really $  - @ weird name 😀 84"
+    try:
+        ddb.sql(f'DROP DATABASE "{weird_db_name}";')
```
JelteF, line 62:
```suggestion
        ddb.sql(f'DROP DATABASE "{weird_db_name}" IF EXISTS;')
```

JelteF (approved):

test/pycheck/motherduck_test.py
Y--: We very sadly don't support `IF EXISTS`

JelteF: It seems I put it in the wrong place:
```suggestion
        ddb.sql(f'DROP DATABASE IF EXISTS "{weird_db_name}";')
```

We use it in the `create_duckdb` function. So it at least seems like it works.

Y--: Oh wow! I've always assumed we didn't... TIL!

test/pycheck/motherduck_test.py
```diff
@@ -56,6 +56,37 @@ def test_md_create_table(md_cur: Cursor, ddb):
...
+    pg.search_path = f"ddb${default_db_name}, public"
+    with pg.cur() as cur:
+        hint = "abc123"
+        cur.sql(f"SET duckdb.motherduck_session_hint = '{hint}';")
+
+        # Sanity check
+        assert cur.sql("SHOW duckdb.motherduck_session_hint;") == hint
+
+        cur.execute(
+            f"SELECT duckdb.enable_motherduck('{md_test_user['token']}', '{weird_db_name}')"
```
JelteF, line 76: Does this actually do something? MotherDuck is already enabled right? We should at least check what the default database is.

test/pycheck/motherduck_test.py
```diff
@@ -56,6 +56,37 @@ def test_md_create_table(md_cur: Cursor, ddb):
         md_cur.sql("CREATE TABLE t4(a int) USING heap")
 
 
+def test_md_read_scaling(pg: Cursor, ddb, default_db_name, md_test_user):
+    weird_db_name = "some 19 really $  - @ weird name 😀 84"
```
JelteF, line 60: I feel like this test tests two very separate things at the same time. Which makes it hard to follow what's happening. Let's split it up in a test that tests that the escaping of the default database. And one that tests the session hint behavior.

I also think we need to add a bunch of other read scaling tests. But this seems like a nice start at least. And the PR fixes an important bug, so let's merge it quickly.

test/pycheck/motherduck_test.py
Y--: Though agree a test on the default DB would be even better. Should be trivial with `SHOW` right?

test/regression/sql/simple_functions.sql
```diff
@@ -10,3 +10,11 @@ SELECT test_escape_uri('https://duckdb.org');
 SELECT test_escape_uri('https://duckdb.org/with space');
 
 SELECT test_escape_uri('foo $ bar # baz / qux');
+
+SELECT test_escape_uri('foo 😀 bar # baz / qux');
+
+SELECT test_escape_uri('Hannes Mühleisen');
+
+SELECT test_escape_uri('Hannes M□hleisen');
```
JelteF, line 18: What character is this square? I only see a square. Maybe add a comment about what it is.

Y--: This is a tribute to the man, he'll know why :-) 
cc @hannes

test/pycheck/motherduck_test.py
```diff
@@ -56,6 +56,60 @@ def test_md_create_table(md_cur: Cursor, ddb):
         md_cur.sql("CREATE TABLE t4(a int) USING heap")
 
 
+def test_md_default_db_escape(pg: Cursor, ddb, default_db_name, md_test_user):
+    weird_db_name = "some 19 really $  - @ weird name 😀 84"
+    try:
+        ddb.sql(f'DROP DATABASE IF EXISTS "{weird_db_name}";')
+    except Exception:
```
JelteF, line 63: no need for this try except now anymore right?

test/pycheck/motherduck_test.py
```diff
@@ -56,6 +56,60 @@ def test_md_create_table(md_cur: Cursor, ddb):
...
+            LEFT JOIN pg_user_mapping um ON um.umserver = fs.oid
+            WHERE fdw.fdwname = 'duckdb' AND fs.srvtype = 'motherduck';
+        """) == ("motherduck", "motherduck", [f"default_database={weird_db_name}"])
+
+
+def test_md_read_scaling(pg: Cursor, ddb, default_db_name, md_test_user):
+    pg.search_path = f"ddb${default_db_name}, public"
+
+    # Make sure MD is not enabled
+    pg.sql("DROP SERVER IF EXISTS motherduck CASCADE;")
```
JelteF, line 92: instead of doing this, can we not simply use `md_cur` instead of `cur`, like we do in the other tests.

JelteF (approved):

PR #743, opened 2025-04-25
Throw error when changing a duckdb GUC after DuckDB is initialized

### LINKED ISSUE #736: Throw error when changing a duckdb GUC after DuckDB is initialized
### Description

We have a bunch of GUCs that control duckdb settings, currently these settings are only synced at creation of the duckdb connection. We should throw an error, not to confuse people and so we can improve this behaviour later by actually syncing the settings after connection setup.

Y--:
Fixes #736

### DIFF
--- a/src/pgduckdb.cpp
+++ b/src/pgduckdb.cpp
@@ -1,6 +1,7 @@
 #include "pgduckdb/pgduckdb.h"
 
 #include <type_traits>
+#include "pgduckdb/pgduckdb_duckdb.hpp"
 
 extern "C" {
 #include "postgres.h"
@@ -79,6 +80,18 @@ DefineCustomVariable(const char *name, const char *short_desc, char **var, GucCo
 }
 
 template <typename T>
+static bool
+GucCheckDuckDBNotInitdHook(T *, void **, GucSource) {
+	if (pgduckdb::DuckDBManager::IsInitialized()) {
+		GUC_check_errmsg(
+		    "Cannot set this variable after DuckDB has been initialized. Use `duckdb.recycle_ddb()` to reset "
+		    "the DuckDB instance.");
+		return false;
+	}
+	return true;
+}
+
+template <typename T>
 static void
 DefineCustomVariable(const char *name, const char *short_desc, T *var, T min, T max, GucContext context = PGC_USERSET,
                      int flags = 0, GucIntCheckHook check_hook = NULL, GucIntAssignHook assign_hook = NULL,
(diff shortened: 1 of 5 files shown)


src/pgduckdb.cpp
```diff
@@ -78,6 +79,18 @@ DefineCustomVariable(const char *name, const char *short_desc, char **var, GucCo
 	                           show_hook);
 }
 
+template <typename T>
+static bool
+GucCheckDuckDBNotInitdHook(T *, void **, GucSource) {
+	if (pgduckdb::DuckDBManager::IsInitialized()) {
+		GUC_check_errmsg(
+		    "Cannot set this variable after DuckDB has been initialized. Use `duckdb.recycle_ddb()` to reset "
```
JelteF, line 87: Reconnect or use...

src/pgduckdb.cpp
```diff
@@ -122,53 +135,54 @@ DuckdbInitGUC() {
 	                     &duckdb_log_pg_explain);
 
 	DefineCustomVariable("duckdb.enable_external_access", "Allow the DuckDB to access external state.",
-	                     &duckdb_enable_external_access, PGC_SUSET);
+	                     &duckdb_enable_external_access, PGC_SUSET, 0, GucCheckDuckDBNotInitdHook);
```
JelteF, line 138: nit: does it make sense to add another template without the `flags` argument, so we don't need to specify `0` soo many times.

src/pgduckdb.cpp
JelteF, line 87:
```suggestion
		    "Cannot set this variable after DuckDB has been initialized. Reconnect to Postgres or use `duckdb.recycle_ddb()` to reset "
```

src/pgduckdb.cpp
Y--: Ah I went back and forth on that. I thought I remembered that you created those with all the possible args, so didn't want to add toe many helpers.

But maybe we do want a `DefineCustomDuckDBVariable` that would do exactly that. I'll give it a shot so we can compare

src/pgduckdb.cpp
```diff
@@ -107,8 +115,8 @@ DefineCustomVariable(const char *name, const char *short_desc, T *var, T min, T
 			T maxValue,
 			GucContext context,
 			int flags,
-			GucIntCheckHook check_hook,
-			GucIntAssignHook assign_hook,
+			GucTypeCheckHook<T> check_hook,
+			GucTypeAssignHook<T> assign_hook,
```
Y--, line 119: This was a "bug" - using a float type would have failed to compile

src/pgduckdb.cpp
```diff
@@ -79,9 +94,15 @@ DefineCustomVariable(const char *name, const char *short_desc, char **var, GucCo
...
 
 template <typename T>
-static void
+using GucTypeCheckHook = bool (*)(T *, void **, GucSource);
+
+template <typename T>
+using GucTypeAssignHook = void (*)(T, void *);
+
+template <typename T>
+void
```
JelteF, line 103: Let's put static back so it doesn't conflict with symbols from possibly other libraries.

```suggestion
static void
```

JelteF (approved):

src/pgduckdb.cpp
Y--: So these are templates, so it won't conflict with anything, and I've put all the concrete implementation in the anonymous ns: https://github.com/duckdb/pg_duckdb/blob/f1fc8f4747e37daf0f9a23f26a50e46b287726fa/src/pgduckdb.cpp#L66

JelteF (approved):

Issue #751, opened 2025-04-30
DuckdbExplainOneQueryHook is not called for "explain execute" queries


ggnmstr:
### What happens?

With `duckdb.force_execution` set to true, this sequence of queries crashes server:
```sql
CREATE TABLE INT8_TBL(q1 int8, q2 int8);
explain (memory, analyze, format json) select * from int8_tbl i8;
prepare int8_query as select * from int8_tbl i8;
explain (memory) execute int8_query; 
```

This happens because first query sets `duckdb_explain_format`    to    `duckdb::ExplainFormat::JSON` in DuckdbExplainOneQueryHook, meanwhile `EXPLAIN EXECUTE` queries are considered Utility statements, so `ExplainOneQuery_hook` is not called and `duckdb_explain_format` is still set to `duckdb::ExplainFormat::JSON`, which it should not be. 

For instance, this sequence of queries works just well:
```sql
explain (memory, analyze, format json) select * from int8_tbl i8;
prepare int8_query as select * from int8_tbl i8;
explain (memory, format json) execute int8_query; 
```
 
I suggest moving code that sets value of  `duckdb_explain_format` from DuckdbExplainOneQueryHook to Duckdb_ExplainCustomScan_Cpp, it seems its the only place where that value is actually needed. 
With that change, server no longer crashes.

### To Reproduce

1. Create fresh DB with pg_duckdb and duckdb.force_execution set to true
2. Run this sequence of queries:
```sql
CREATE TABLE INT8_TBL(q1 int8, q2 int8);
explain (memory, analyze, format json) select * from int8_tbl i8;
prepare int8_query as select * from int8_tbl i8;
explain (memory) execute int8_query; 
```
3. Get a server crash

### OS:

Linux

### pg_duckdb Version (if built from source use commit hash):

ea75539aa6bf4520fce537cffbe75fb12e4bcc65

### Postgres Version (if built from source use commit hash):

17

### Full Name:

Jora Babayan

### Affiliation:

Postgres Professional

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #752, opened 2025-04-30
fix: resolve pg_duckdb #751

ggnmstr:
My suggestion to fix issue #751 
For now it's copied, so in cases where hook is called work is done twice, but we can remove that code from hook

### DIFF
--- a/test/pycheck/explain_test.py
+++ b/test/pycheck/explain_test.py
@@ -4,6 +4,8 @@ These tests are using Python mainly because the output of EXPLAIN ANALYZE
 contains timings, so the output is not deterministic.
 """
 
+import json
+
 from .utils import Cursor
 
 import pytest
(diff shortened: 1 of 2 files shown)


JelteF:
Could you add a regression test, probably in the test_explain.py

test/pycheck/explain_test.py
```diff
@@ -83,6 +83,18 @@ def test_explain(cur: Cursor):
     assert "Planning Time" in (result[0]).keys()
     assert "Execution Time" in (result[0]).keys()
 
+    cur.sql(
+        "EXPLAIN (MEMORY, ANALYZE, FORMAT JSON) SELECT * FROM test_table t"
```
JelteF, line 87: MEMORY is new in PG17, se let's use BUFFERS instead

JelteF:
This doesn't crash the server (which is obviously much better), but it still results in incorrect results. It will still show the DuckDB part of the plan as JSON on the first run. i.e. when running the following queries, only the last of the two queries (which are the same) returns the correct result.
```
set duckdb.force_execution = true;
CREATE TABLE INT8_TBL(q1 int8, q2 int8);
explain (memory, analyze, format json) select * from int8_tbl i8;
prepare int8_query as select * from int8_tbl i8;
explain (memory) execute int8_query;
explain (memory) execute int8_query;
```

JelteF:
I think the solution is instead to use a different hook to get this information. e.g. to use `explain_per_plan_hook`

ggnmstr:
`explain_per_plan_hook` will only be added in PostgreSQL 18. Should search for some other possible way then.

JelteF:
I think the best course of action is to merge this fix as is. Even if the resulting explain plans are not exactly what we want, it's still much better than crashing. I don't think there's an easy way to get the explain plans correct on pre-PG18. 

Could you update the tests to show that the current plans are not what we want? i.e. by using some code comments and running the `explain execute int8_query;` query two times in a row (showing that the results are different per run).

Finally EXPLAIN EXECUTE is a rather edge case for people to run. Especially since prepared statements have no perf benefit with pg_duckdb currently. So I don't feel too bad about it not working as expected in all cases.

ggnmstr:
Sure, now the test runs that query 2 times and asserts their output format is different

JelteF (approved):

test/pycheck/explain_test.py
```diff
@@ -84,6 +86,38 @@ def test_explain(cur: Cursor):
...
+    """
+    cur.sql("CREATE TABLE test_table (id int, name text)")
+    cur.sql("EXPLAIN (ANALYZE, FORMAT JSON) SELECT * FROM test_table t")
+    cur.sql("PREPARE test_query AS SELECT * FROM test_table t")
+    result = cur.sql("EXPLAIN EXECUTE test_query")
+    plan = "\n".join(result)
+    print(plan)
+    result2 = cur.sql("EXPLAIN EXECUTE test_query")
+    plan2 = "\n".join(result2)
+    print(plan2)
```
Y--, line 105: Can we make some assertions to the plans rather than printing them out? (or only print them when a debug mode is turned on?)

Y-- (approved):

Issue #748, opened 2025-04-28
Server crashes on executing "CREATE TABLE"


saygoodbyye:
### What happens?

```
CFLAGS='-Og' ./configure --enable-tap-tests --enable-debug --with-openssl --enable-cassert --quiet --prefix="$PGPREFIX"
```

### To Reproduce

postgresql.conf:
```
shared_preload_libraries='pg_duckdb'
duckdb.force_execution = true
duckdb.postgres_role = 'duckdb_group'
log_temp_files = -1
```
test.sql
```sql
create extension pg_duckdb;

DROP TABLE sitmp1;
CREATE SCHEMA selinto_schema;
  CREATE TABLE selinto_schema.tbl_nodata4 (a) AS
  EXECUTE data_sel WITH NO DATA;
RESET SESSION AUTHORIZATION;
```
backtrace:
```c
#0  list_nth (n=-1, list=0x0) at /tmp/pgsql/include/server/nodes/pg_list.h:301
#1  pgduckdb::IsAllowedStatement (query=query@entry=0x55e877d7f220, throw_error=throw_error@entry=false) at src/pgduckdb_hooks.cpp:189
#2  0x00007fd7e099a2cb in pgduckdb::ShouldTryToUseDuckdbExecution (query=query@entry=0x55e877d7f220) at src/pgduckdb_hooks.cpp:164
#3  0x00007fd7e0994288 in DuckdbHandleDDL (params=0x0, query_string=0x55e877d7e3d0 "CREATE TABLE selinto_schema.tbl_nodata4 (a) AS\n  EXECUTE data_sel WITH NO DATA;", pstmt=0x55e877d7f110) at src/pgduckdb_ddl.cpp:285
#4  DuckdbUtilityHook_Cpp (qc=0x7fffe2baa400, dest=0x55e877d7f6e0, query_env=0x0, params=0x0, context=PROCESS_UTILITY_TOPLEVEL, read_only_tree=false, 
    query_string=0x55e877d7e3d0 "CREATE TABLE selinto_schema.tbl_nodata4 (a) AS\n  EXECUTE data_sel WITH NO DATA;", pstmt=0x55e877d7f110) at src/pgduckdb_ddl.cpp:522
[...]

PR #757, opened 2025-05-05
Don't crash when query doesn't contain FROM

Y--:
Fixes https://github.com/duckdb/pg_duckdb/issues/748

### DIFF
--- a/src/pgduckdb_hooks.cpp
+++ b/src/pgduckdb_hooks.cpp
@@ -185,12 +185,15 @@ IsAllowedStatement(Query *query, bool throw_error) {
 	}
 
 	/* We don't support modifying statements on Postgres tables yet */
-	if (ContainsFromClause(query) && query->commandType != CMD_SELECT) {
-		RangeTblEntry *resultRte = list_nth_node(RangeTblEntry, query->rtable, query->resultRelation - 1);
-		if (!::IsDuckdbTable(resultRte->relid)) {
-			elog(elevel, "DuckDB does not support modififying Postgres tables");
-			return false;
+	if (query->commandType != CMD_SELECT) {
+		if (ContainsFromClause(query)) {
+			RangeTblEntry *resultRte = list_nth_node(RangeTblEntry, query->rtable, query->resultRelation - 1);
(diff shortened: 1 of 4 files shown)


test/regression/sql/issue_748.sql
```diff
@@ -0,0 +1,6 @@
+SET duckdb.force_execution = TRUE;
+
+CREATE SCHEMA selinto_schema;
+
+CREATE TABLE selinto_schema.tbl_nodata4 (a) AS
+EXECUTE data_sel WITH NO DATA;
```
JelteF, line 6: Should drop the schema here.

src/pgduckdb_hooks.cpp
```diff
@@ -185,7 +185,7 @@ IsAllowedStatement(Query *query, bool throw_error) {
 	}
 
 	/* We don't support modifying statements on Postgres tables yet */
-	if (query->commandType != CMD_SELECT) {
+	if (ContainsFromClause(query) && query->commandType != CMD_SELECT) {
```
JelteF, line 188: This seems a a weird place to add this check. It seem to have anything to do with checking whether the postgres table is a modifying statement or not.

Y--: I agree, though the line below uses `list_nth_node` with `query->rtable` which could be `NULL` without this check, thus failing the `Assert(list != NIL)` in `list_nth_cell`

Y--: Though re-reading now with fresh eyes it was indeed pretty weird. Reworked here: https://github.com/duckdb/pg_duckdb/pull/757/commits/2ed8ee441ddd4a0e74bb184f5872dc6235baca00

JelteF (approved):

JelteF (commented):
Actually, checking for rtable == NULL instead of using `ContainsFromClause` would be be a bit clearer too imo. Because `rtable` only corresponds to FROM for SELECT queries, and in this block we're explicitely not talking about SELECT queries due to the `!= CMD_SELECT` check.

e.g. `INSERT INTO abc VALUES (1)` will have `abc` in the rtable.

JelteF (approved):

PR #758, opened 2025-05-05
Add source location

Y--:
Fixes https://github.com/duckdb/pg_duckdb/issues/523

Reproduce `elog(ERROR, ...)` implementation to override the source function/file/line.

### DIFF
--- a/include/pgduckdb/utility/cpp_wrapper.hpp
+++ b/include/pgduckdb/utility/cpp_wrapper.hpp
@@ -10,7 +10,7 @@ namespace pgduckdb {
 
 template <typename Func, Func func, typename... FuncArgs>
 typename std::invoke_result<Func, FuncArgs &...>::type
-__CPPFunctionGuard__(const char *func_name, FuncArgs &...args) {
+__CPPFunctionGuard__(const char *func_name, const char *file_name, int line, FuncArgs &...args) {
 	const char *error_message = nullptr;
 	auto pg_es_start = PG_exception_stack;
 	try {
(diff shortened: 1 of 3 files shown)


JelteF (changes requested):
A few things:
1. I think having the line number output in tests is going to cause a ton of diff noise when changing files.
2. I think it's worth doing the more difficult thing to integrate into the normal postgres errors: 
   > The alternative would be to reproduce what elog(ERROR, ...) does and expose some of the elog.c internals.

   Afaict all that's needed is to inline the ereport/elog implementation, similarly to what we already do in `include/pgduckdb/logger.hpp`. And then change the call to `errfinish` to take the correct file, line number & function name.
3. Changing InfokeCPPFunc like is a great start, but it's only half of the issue. I think we'd also want to to change `PostgresFunctionGuard` to store the original errdata in the exception, and propagate it upwards. Otherwise you'll always get the line info from the final wrapper location, not from the actual source. This could be done in a future PR though.

Y--:
>    Afaict all that's needed is to inline the ereport/elog implementation, similarly to what we already do in `include/pgduckdb/logger.hpp`. And then change the call to `errfinish` to take the correct file, line number & function name.

I think I tried that but it wasn't working just quite. I'll revisit it.

Y--:
> > Afaict all that's needed is to inline the ereport/elog implementation, similarly to what we already do in `include/pgduckdb/logger.hpp`. And then change the call to `errfinish` to take the correct file, line number & function name.
> 
> I think I tried that but it wasn't working just quite. I'll revisit it.

Not sure what I missed yesterday but this wasn't actually that bad...

I'll also work on a PR for the PG wrapper.

JelteF (approved):

PR #765, opened 2025-05-07
Disallow ALTER TABLE ... RENAME in a mixed transaction

JelteF:
To protect users against themselves we disallow DDL on DuckDB tables and
changes to Postgres tables in the same transaction. We were accidentally
not doing this for `ALTER TABLE ... RENAME` statements though. This
fixes that and refactors code to catch issues like that better in the
future.

### DIFF
--- a/src/pgduckdb_ddl.cpp
+++ b/src/pgduckdb_ddl.cpp
@@ -371,16 +371,30 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
 		return;
 	} else if (IsA(parsetree, RenameStmt)) {
 		auto stmt = castNode(RenameStmt, parsetree);
-		if (stmt->renameType != OBJECT_SCHEMA) {
-			/* We only care about schema renames for now */
-			return;
+		if (stmt->renameType == OBJECT_SCHEMA) {
+			if (strncmp("ddb$", stmt->subname, 4) == 0) {
+				elog(ERROR, "Changing the name of a ddb$ schema is currently not supported");
+			}
+			if (strncmp("ddb$", stmt->newname, 4) == 0) {
+				elog(ERROR, "Changing a schema to a ddb$ schema is currently not supported");
+			}
 		}
-		if (strncmp("ddb$", stmt->subname, 4) == 0) {
-			elog(ERROR, "Changing the name of a ddb$ schema is currently not supported");
+		if (stmt->renameType == OBJECT_TABLE || stmt->renameType == OBJECT_COLUMN) {
+			Oid relation_oid = RangeVarGetRelid(stmt->relation, AccessShareLock, false);
+			Relation relation = RelationIdGetRelation(relation_oid);
+
+			if (pgduckdb::IsDuckdbTable(relation)) {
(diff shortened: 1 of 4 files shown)


src/pgduckdb_ddl.cpp
```diff
@@ -371,16 +371,30 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
...
+			if (strncmp("ddb$", stmt->subname, 4) == 0) {
+				elog(ERROR, "Changing the name of a ddb$ schema is currently not supported");
+			}
+			if (strncmp("ddb$", stmt->newname, 4) == 0) {
+				elog(ERROR, "Changing a schema to a ddb$ schema is currently not supported");
+			}
 		}
-		if (strncmp("ddb$", stmt->newname, 4) == 0) {
-			elog(ERROR, "Changing a schema to a ddb$ schema is currently not supported");
+		if (stmt->renameType == OBJECT_TABLE || stmt->renameType == OBJECT_COLUMN) {
```
Y--, line 382: nit - `else if` (no need to run the check again if we entered the one above)

src/pgduckdb_ddl.cpp
```diff
@@ -371,16 +371,30 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
...
+			Oid relation_oid = RangeVarGetRelid(stmt->relation, AccessShareLock, false);
+			Relation relation = RelationIdGetRelation(relation_oid);
```
Y--, line 384: Interestingly `relation` is colored as a keyword by GH... Not sure if that's a big deal, but we could name it `rel` instead too without loosing readability?

src/pgduckdb_ddl.cpp
```diff
@@ -371,16 +371,30 @@ DuckdbHandleDDL(PlannedStmt *pstmt, const char *query_string, ParamListInfo para
...
 	} else if (IsA(parsetree, RenameStmt)) {
 		auto stmt = castNode(RenameStmt, parsetree);
-		if (stmt->renameType != OBJECT_SCHEMA) {
-			/* We only care about schema renames for now */
-			return;
-		}
-		if (strncmp("ddb$", stmt->subname, 4) == 0) {
-			elog(ERROR, "Changing the name of a ddb$ schema is currently not supported");
+		if (stmt->renameType == OBJECT_SCHEMA) {
+			if (strncmp("ddb$", stmt->subname, 4) == 0) {
```
Y--, line 375: Can we add an `IsDuckDBSchema` in `string_utils`?

```c++
inline bool
IsDuckDBSchema(const char* s) {
  return strncmp("ddb$", s, 4) == 0;
}
```

```suggestion
			if (IsDuckDBSchema(stmt->subname)) {
```
(and replace all ~10 instances in our codebase?)

Y-- (approved):
LGTM, some minor nits

src/pgduckdb_ddl.cpp
JelteF: Doing this one in a follow up PR

Y-- (approved):

PR #768, opened 2025-05-07
Test two MD users connections

Y--:


test/pycheck/conftest.py
```diff
@@ -27,6 +25,16 @@ def default_db_name(request):
...
 
 
+def init_pg(pgobj, **kwargs):
+    pgobj.sql("GRANT CREATE ON SCHEMA public TO duckdb_group", **kwargs)
+    pgobj.sql("CREATE EXTENSION pg_duckdb", **kwargs)
+
+
+def teardown_pg(pgobj, **kwargs):
+    pgobj.sql("REVOKE ALL PRIVILEGES ON SCHEMA public FROM duckdb_group", **kwargs)
+    pgobj.sql("DROP EXTENSION pg_duckdb CASCADE", **kwargs)
```
JelteF, line 35: Why is this needed for every test now?

Y--: Ugh I didn't mean to keep it for every test- only for the multi-db ones. Thanks for catching.

PR #771, opened 2025-05-07
Drop MD tables when MD support is disabled

Y--:
Currently PG catalog entries mirroring MD tables are dropped only when the extension is.

This PR changes this behavior to have them depend on the `motherduck` SERVER so that when it is removed, the entries are removed as well.

### DIFF
--- a/src/pgduckdb_background_worker.cpp
+++ b/src/pgduckdb_background_worker.cpp
@@ -42,6 +42,7 @@ extern "C" {
 #include "catalog/namespace.h"
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_extension.h"
+#include "catalog/pg_foreign_server.h"
 #include "utils/acl.h"
 #include "utils/guc.h"
 #include "utils/memutils.h"
@@ -925,7 +926,7 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
 	if (!is_default_db) {
 		/*
 		 * For ddb$ schemas we need to record a dependency between the schema
-		 * and the extension, so that DROP EXTENSION also drops these schemas.
+		 * and the FDW, so that DROP SERVER motherduck also drops these schemas.
 		 */
 		schema_oid = get_namespace_oid(postgres_schema_name, true);
 		if (schema_oid == InvalidOid) {
@@ -939,12 +940,12 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
 		    .objectId = schema_oid,
 		    .objectSubId = 0,
 		};
-		ObjectAddress extension_address = {
-		    .classId = ExtensionRelationId,
-		    .objectId = pgduckdb::ExtensionOid(),
+		ObjectAddress server_address = {
+		    .classId = ForeignServerRelationId,
+		    .objectId = GetMotherduckForeignServerOid(), // pgduckdb::ExtensionOid(),
 		    .objectSubId = 0,
 		};
-		recordDependencyOn(&schema_address, &extension_address, DEPENDENCY_NORMAL);
+		recordDependencyOn(&schema_address, &server_address, DEPENDENCY_NORMAL);
 	}
 
 	/* Success, so we commit the subtransaction */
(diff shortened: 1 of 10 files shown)


src/pgduckdb_background_worker.cpp
```diff
@@ -932,12 +933,12 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
 		    .objectId = schema_oid,
 		    .objectSubId = 0,
 		};
-		ObjectAddress extension_address = {
-		    .classId = ExtensionRelationId,
-		    .objectId = pgduckdb::ExtensionOid(),
+		ObjectAddress server_address = {
+		    .classId = ForeignServerRelationId,
+		    .objectId = GetMotherduckForeignServerOid(), // pgduckdb::ExtensionOid(),
```
JelteF, line 938:
```suggestion
		    .objectId = GetMotherduckForeignServerOid(),
```

src/pgduckdb_background_worker.cpp
```diff
@@ -932,12 +933,12 @@ CreateSchemaIfNotExists(const char *postgres_schema_name, bool is_default_db) {
...
 		    .objectSubId = 0,
 		};
-		recordDependencyOn(&schema_address, &extension_address, DEPENDENCY_NORMAL);
+		recordDependencyOn(&schema_address, &server_address, DEPENDENCY_NORMAL);
```
JelteF, line 941: This works for tables in a `ddb$` schema, because those will then depend on the SERVER indirectly through their schema. But this doesn't yet work for tables in the default database, for those we need to create a direct dependency between the table and the schema. (and probably we just want to create the direct dependency for any `duckdb` table, no matter what schema they are in, just for consistency).

JelteF (changes requested):

src/pgduckdb_background_worker.cpp
```diff
@@ -773,6 +784,27 @@ CreateTable(const char *postgres_schema_name, const char *table_name, const char
...
+		if (HeapTupleIsValid(new_table_tuple)) {
+			Form_pg_class postgres_relation = (Form_pg_class)GETSTRUCT(new_table_tuple);
+			auto oid = postgres_relation->oid;
+			ReleaseSysCache(new_table_tuple);
+			ObjectAddress object_address = {
+			    .classId = RelationRelationId,
+			    .objectId = oid,
+			    .objectSubId = 0,
+			};
+			recordDependencyOnMDServer(&object_address);
```
JelteF, line 800: This should be done in our DDL hook instead. That way the dependency is also recorded when someone creates a motherduck table from Postgres, not just when a table is synced to by the background worker.

src/pgduckdb_background_worker.cpp
```diff
@@ -773,6 +786,26 @@ CreateTable(const char *postgres_schema_name, const char *table_name, const char
...
+			ObjectAddress object_address = {
+			    .classId = RelationRelationId,
+			    .objectId = GetOid(postgres_relation),
+			    .objectSubId = 0,
+			};
+			recordDependencyOnMDServer(&object_address);
+		} else {
+			// Something went really wrong (since `create_table_succeeded`)
+			elog(WARNING, "Failed to find table '%s.%s' so won't record dependency to 'motherduck' SERVER", table_name,
+			     postgres_schema_name);
```
JelteF, line 805:
```suggestion
			     postgres_schema_name);
			     RollbackAndReleaseCurrentSubTransaction();
				 return false;
```

And as a nit: then we can change this to "early return" style if we change
```
if (HeapTupleIsValid(new_table_tuple)) {
```
to the inverse
```
if (!HeapTupleIsValid(new_table_tuple)) {

File: src/pgduckdb_background_worker.cpp [hunk 5]
```diff
@@ -230,6 +233,16 @@ BgwMainLoop() {
 	elog(LOG, "pg_duckdb background worker for database '%s' (%u) has now terminated.", db_name, MyDatabaseId);
 }
 
+void
+RecordDependencyOnMDServer(ObjectAddress *object_address) {
```
JelteF, line 237: ultra-nit: maybe move this to `pgduckdb_fdw.cpp`

JelteF (dismissed):
Feel free to ignore the nit

JelteF (approved):

PR #779, opened 2025-05-09
CI: Auto-expand the regression.diffs output

JelteF:
You basically always want to look at the diffs when a test failed. This
makes sure it gets auto-expanded.


Y-- (approved):
Smart!

PR #783, opened 2025-05-09
Simplify how CREATE SECRET command is built

JelteF:
By adding TYPE as the first argument to CREATE SECRET, we don't need to
track if we still need to add commas. The final command also reads much
more natural, because the TYPE is at the start.

This also drop all SERVERs at the end of the duckdb_secrets.sql test,
keeping them around was significantly slowing down the tests that were
later in the schedule. The `duckdb_secrets.sql` test also takes very
long itself to run on my machine (170 seconds!). That's something we
should probably fix too, but this is at least a good start.

### DIFF
--- a/src/pgduckdb_secrets_helper.cpp
+++ b/src/pgduckdb_secrets_helper.cpp
@@ -49,21 +49,12 @@ GetOption(List *options_list, const char *name) {
 
 namespace {
 
-bool
+void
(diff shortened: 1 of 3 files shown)


Y--:
Ah yeah much nicer indeed :-) I wonder how I missed it in the initial PR! thanks!

Y-- (approved):

PR #784, opened 2025-05-10
Fix memory leak for JSON/LIST type during Postgres table scan

### LINKED ISSUE #750: Memory leak when scanning postgres table with jsonb & list type
### What happens?

When converting Postgres tuples to DuckDB chunk data, we reuse some PG functions (e.g., `JsonbToCString` and `deconstruct_array`). These functions internally allocate memory via `palloc` that isn't explicitly freed. This memory leaks until the end of query (when it is released with the ExecutorState).

Possible fixes:
1) Implement custom version of `JsonbToCString` and `deconstruct_array` in pgduckdb. However, this approach might be boverse and introduce compatibility issue.
2) Switch to a dedicated MemoryContext before calling these functions, then periodically reset memory context to reclaim memory (e.g., based on the MemoryContext's allocated memory)

### To Reproduce

```
set duckdb.force_execution=on;

create table j1(c jsonb);
insert into j1 select '{"aasdkjhzxcuyaisdyasuidqwe": 1, "b12314jhkjahsdjkahsdjk": 2}'::jsonb from generate_series(1, 10000000);
-- memory keeps growing
select * from j1 order by 1 limit 1; 

create table a1(c text[]);
insert into a1 select array['aasdkjhzxcuyaisdyasuidqwe', 'b12314jhkjahsdjkahsdjk'] from generate_series(1, 10000000);
-- memory keeps growing
select * from a1 order by 1 limit 1; 
```

### OS:

centos

main branch

17

### Hardware:

_No response_

### Full Name:

Yuwei XIAO

### Affiliation:

zbyte-inc

Yes

- [x] Yes, I have

- [x] Yes, I have

YuweiXiao:
Fixes #750

### DIFF
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -1715,7 +1715,7 @@ ConvertPostgresToDuckValue(Oid attr_type, Datum value, duckdb::Vector &result, i
 	case duckdb::LogicalTypeId::VARCHAR: {
 		// NOTE: This also handles JSON
 		if (attr_type == JSONBOID) {
-			AppendJsonb(result, value, offset);
+			PostgresFunctionGuard(AppendJsonb, result, value, offset);
 			break;
 		} else {
 			AppendString(result, value, offset, attr_type == BPCHAROID);
@@ -1820,13 +1820,13 @@ ConvertPostgresToDuckValue(Oid attr_type, Datum value, duckdb::Vector &result, i
 		int16 typlen;
 		bool typbyval;
 		char typalign;
-		get_typlenbyvalalign(elem_type, &typlen, &typbyval, &typalign);
+		PostgresFunctionGuard(get_typlenbyvalalign, elem_type, &typlen, &typbyval, &typalign);
 
 		int nelems;
 		Datum *elems;
 		bool *nulls;
 		// Deconstruct the array into Datum elements
-		deconstruct_array(array, elem_type, typlen, typbyval, typalign, &elems, &nulls, &nelems);
+		PostgresFunctionGuard(deconstruct_array, array, elem_type, typlen, typbyval, typalign, &elems, &nulls, &nelems);
 
 		if (ndims == -1) {
 			throw duckdb::InternalException("Array type has an ndims of -1, so it's actually not an array??");
--- a/src/pgduckdb_types.cpp
+++ b/src/pgduckdb_types.cpp
@@ -40,6 +40,8 @@ extern "C" {
 
 namespace pgduckdb {
 
+extern MemoryContext duckdb_pg_scan_memory_ctx;
+
 NumericVar FromNumeric(Numeric num);
 
 struct NumericAsDouble : public duckdb::ExtraTypeInfo {
@@ -1505,11 +1507,16 @@ AppendString(duckdb::Vector &result, Datum value, idx_t offset, bool is_bpchar) 
 
 static void
 AppendJsonb(duckdb::Vector &result, Datum value, idx_t offset) {
+	D_ASSERT(duckdb_pg_scan_memory_ctx != nullptr);
+	MemoryContext old_context = MemoryContextSwitchTo(duckdb_pg_scan_memory_ctx);
 	auto jsonb = DatumGetJsonbP(value);
-	auto jsonb_str = JsonbToCString(NULL, &jsonb->root, VARSIZE(jsonb));
-	duckdb::string_t str(jsonb_str);
+	StringInfo str = makeStringInfo();
+	auto json_str = JsonbToCString(str, &jsonb->root, VARSIZE(jsonb));
 	auto data = duckdb::FlatVector::GetData<duckdb::string_t>(result);
-	data[offset] = duckdb::StringVector::AddString(result, str);
+	data[offset] = duckdb::StringVector::AddString(result, json_str, str->len);
+	pfree(str->data);
+	pfree(str);
+	MemoryContextSwitchTo(old_context);
 }
 
 static void
@@ -1810,6 +1817,8 @@ ConvertPostgresToDuckValue(Oid attr_type, Datum value, duckdb::Vector &result, i
 		break;
 	}
 	case duckdb::LogicalTypeId::LIST: {
+		D_ASSERT(duckdb_pg_scan_memory_ctx != nullptr);
+		MemoryContext old_context = MemoryContextSwitchTo(duckdb_pg_scan_memory_ctx);
 		// Convert Datum to ArrayType
 		auto array = DatumGetArrayTypeP(value);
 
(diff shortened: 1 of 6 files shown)


YuweiXiao:
Hey @JelteF, to establish a clear functional boundary for the memory context, I keep it scoped within PostgresScan and initialize & reset it alongside the scan's global state.

YuweiXiao:
It seems the pycheck regression also failed on my local build with the main branch :(

src/scan/postgres_scan.cpp
```diff
@@ -267,8 +284,13 @@ PostgresScanTableFunction::PostgresScanFunction(duckdb::ClientContext &, duckdb:
 
 		SlotGetAllAttrs(slot);
 		InsertTupleIntoChunk(output, local_state, slot);
+
+		if (MemoryContextMemAllocated(duckdb_pg_scan_memory_ctx, false) > 8 * 1024 * 1024L) {
+			MemoryContextReset(duckdb_pg_scan_memory_ctx);
+		}
```
JelteF, line 290: Is this one really needed? I'd rather keep it simple and only do the reset after we are done with the vector. Doing stuff with duckdb is always at the vector level. Then we also don't have to find a decent magic value, like the 8MB you use now.

src/scan/postgres_scan.cpp
```diff
@@ -170,10 +182,15 @@ PostgresScanGlobalState::PostgresScanGlobalState(Snapshot _snapshot, Relation _r
 	ConstructTableScanQuery(input);
 	table_reader_global_state =
 	    duckdb::make_shared_ptr<PostgresTableReader>(scan_query.str().c_str(), count_tuples_only);
+	duckdb_pg_scan_memory_ctx =
+	    AllocSetContextCreate(CurrentMemoryContext, "DuckDBPerTupleContext", ALLOCSET_DEFAULT_MINSIZE,
+	                          ALLOCSET_DEFAULT_INITSIZE, ALLOCSET_DEFAULT_MAXSIZE);
```
JelteF, line 187: This is incorrect. Every postgres scan will now create a separate memory context, which they all assign to the same global variable. The cleanup to set to nullptr is also dangerous, if one scan finishes before another. I think what would be better is to do the following:
1. Storing this context actually in the "global state" of the scan, instead of storing it into a process global variable.
2. Switch to this context right after we take the `GlobalProcessLock` in `PostgresScanFunction`, and switching back right before we release it.

Also the name of the context could be improved imo
```suggestion
	duckdb_pg_scan_memory_ctx =
	    AllocSetContextCreate(CurrentMemoryContext, "DuckdbScanContext", ALLOCSET_DEFAULT_MINSIZE,
	                          ALLOCSET_DEFAULT_INITSIZE, ALLOCSET_DEFAULT_MAXSIZE);
```

src/pgduckdb_types.cpp
```diff
@@ -1505,11 +1507,16 @@ AppendString(duckdb::Vector &result, Datum value, idx_t offset, bool is_bpchar)
...
 	auto jsonb = DatumGetJsonbP(value);
-	auto jsonb_str = JsonbToCString(NULL, &jsonb->root, VARSIZE(jsonb));
-	duckdb::string_t str(jsonb_str);
+	StringInfo str = makeStringInfo();
+	auto json_str = JsonbToCString(str, &jsonb->root, VARSIZE(jsonb));
 	auto data = duckdb::FlatVector::GetData<duckdb::string_t>(result);
-	data[offset] = duckdb::StringVector::AddString(result, str);
+	data[offset] = duckdb::StringVector::AddString(result, json_str, str->len);
+	pfree(str->data);
+	pfree(str);
```
JelteF, line 1518: Is this still needed? If not, let's keep this simple and just rely on the memory context resets.

JelteF (changes requested):

src/scan/postgres_scan.cpp
YuweiXiao: ah yes, you're right.

src/scan/postgres_scan.cpp
YuweiXiao: Yes, let's simplify by removing it. I was concerned about the wide table or large JSON columns. For example, a single JSON row can be several MB in size, and 2048 such rows would quickly consume a lot of memory. However, in this case, the DuckDB data chunk itself also consumes a significant amount of memory.

src/scan/postgres_scan.cpp
```diff
@@ -11,6 +11,12 @@
 
 #include <numeric> // std::accumulate
 
+extern "C" {
+#include "postgres.h"
+
+#include "utils/memutils.h"
+}
+
```
JelteF, line 19: Instead of inlcuding postgres headers here (in a file that we managed to remove all postgres includes from). Let's create some basic wrapper functions in a new file in `src/pg/`, maybe called `memory.cpp`. Also `AllocSetContextCreate` and `MemoryContextReset` need to be wrapped in a `PostgresFunctionGuard` when doing so.

src/pgduckdb_types.cpp
```diff
@@ -1506,10 +1506,10 @@ AppendString(duckdb::Vector &result, Datum value, idx_t offset, bool is_bpchar)
 static void
 AppendJsonb(duckdb::Vector &result, Datum value, idx_t offset) {
 	auto jsonb = DatumGetJsonbP(value);
-	auto jsonb_str = JsonbToCString(NULL, &jsonb->root, VARSIZE(jsonb));
-	duckdb::string_t str(jsonb_str);
+	StringInfo str = makeStringInfo();
+	auto json_str = JsonbToCString(str, &jsonb->root, VARSIZE(jsonb));
```
JelteF, line 1510: nice change to remove the additional copy. This change does make me realize that we need to wrap AppendJsonb in a PostgresFunctionGuard. And probably we need the same for the list one, since that uses also allocates right?

JelteF: Actually it doesn't seem to save an additional copy. AddString internally will still create a duckdb::string_t out of the `json_str` + `str->len`. But it does save an additional `strlen`, so this still seems like a good change.

JelteF (changes requested):
CI is failing (and see rest of my comments).

src/pgduckdb_types.cpp
YuweiXiao: I believe we saved one copy of the string. The conversion steps are as follows:

1. `JsonBToString` creates the string buffer.
2. `duckdb::string_t str(jsonb_str);` duplicates the string buffer into a C++ string.
3. `AppendString` performs the string -> string copy internally.

With this PR, step 2 is eliminated. For LIST conversion, I wrap the PostgreSQL function with the guard. Do you think we need a separate function, perhaps `AppendList`?

src/scan/postgres_scan.cpp
YuweiXiao: good point!

YuweiXiao:
The CI passed in my local run with the latest change. Could you please trigger the pipeline, @JelteF ? Thank you!

JelteF (approved):

PR #785, opened 2025-05-12
tests: Handle DDL command in cur.sql() differently

JelteF:
For ease of programming we have a `cur.sql()` method that can be used to execute any sql command. Fetching rows for DDL statements doesn't work though, so we were catching that error and ignoring it. [In psycopg 3.2.8 the exact error message was changed][1].

Instead of detecting the new error message, this change makes sure that we don't trigger this error at all. This is done by detecting whether our result has rows, before calling `fetchall()`.

Originally reported @alphaxxxxx in https://github.com/duckdb/pg_duckdb/pull/784#issuecomment-2871060527

[1]: https://github.com/psycopg/psycopg/commit/1eb7e5aa85b5350c25abd5f35ad9323aa531ff56

### DIFF
--- a/test/pycheck/utils.py
+++ b/test/pycheck/utils.py
@@ -312,13 +312,12 @@ class Cursor(OutputSilencer):
 
     def sql(self, query, params=None, **kwargs) -> Any:
         self.execute(query, params, **kwargs)
-        try:
-            return simplify_query_results(self.fetchall())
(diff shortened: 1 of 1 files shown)


Y--:
Ah I was just working on this :-) Thnaks!

Y-- (approved):

PR #788, opened 2025-05-12
Support PG18 beta1

JelteF:
Did some of the initial to support PG18 beta1. It definitely needs another self-review, but seems good to at least put it in a draft PR.


Z-Xiao-M:
Perhaps it needs to be modified like this

https://github.com/duckdb/pg_duckdb/blob/49cd67618e683a15404f9edd78e2b4e79b401547/src/pgduckdb.cpp#L16
```c
#ifdef PG_MODULE_MAGIC_EXT
PG_MODULE_MAGIC_EXT(.name = "pg_duckdb", .version = "1.0.0");
#else
PG_MODULE_MAGIC;
#endif
```

PR #790, opened 2025-05-13
Fix `loop-until`

Y--:
In a recent refactoring we changed the `loop_until` implementation using a generator with `while`. This, in effect defeat the purpose of the function since it will loop indefinitely, the generator being always truthy.

This PR replace `loop_until` with a `wait_until` function everywhere which checks if the given function becomes truthy eventually.

### DIFF
--- a/test/pycheck/motherduck_test.py
+++ b/test/pycheck/motherduck_test.py
@@ -7,6 +7,7 @@ library for that purpose.
 """
 
 import datetime
+import duckdb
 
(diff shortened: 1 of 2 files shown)


JelteF:
Instead, how about we replace the `while loop_until():` with `for _ in loop_until():`. Then we don't need to define functions and can use plain `break`, `continue` and `return` to control the flow of the loop.

JelteF:
No super strong opinion against this change though. Feel free to merge your new function.

JelteF (approved):

JelteF (approved):

PR #795, opened 2025-05-19
Don't convert unsupported numerics to double by default

JelteF:
This stops converting unsupported NUMERIC sizes to doubles. This behaviour is still supported, but it is now opt-in. This means pg_duckdb won't silently reduce precision. It also starts requiring users to specify precision when using NUMERIC types in a table definition, because the default precision is different between PostgreSQL and DuckDB.

Closes #471

### DIFF
--- a/src/pgduckdb_guc.cpp
+++ b/src/pgduckdb_guc.cpp
@@ -111,6 +111,7 @@ DefineCustomDuckDBVariable(const char *name, const char *short_desc, T *var, T m
 
 bool duckdb_force_execution = false;
 bool duckdb_unsafe_allow_mixed_transactions = false;
+bool duckdb_convert_unsupported_numeric_to_double = false;
 bool duckdb_log_pg_explain = false;
 int duckdb_max_workers_per_postgres_scan = 2;
 char *duckdb_motherduck_session_hint = strdup("");
@@ -136,6 +137,10 @@ InitGUC() {
 	DefineCustomVariable("duckdb.unsafe_allow_mixed_transactions",
 	                     "Allow mixed transactions between DuckDB and Postgres",
 	                     &duckdb_unsafe_allow_mixed_transactions);
+
+	DefineCustomVariable("duckdb.convert_unsupported_numeric_to_double",
+	                     "Convert NUMERIC types of unsupported precision to DOUBLE",
+	                     &duckdb_convert_unsupported_numeric_to_double);
 
 	DefineCustomVariable("duckdb.log_pg_explain", "Logs the EXPLAIN plan of a Postgres scan at the NOTICE log level",
 	                     &duckdb_log_pg_explain);
--- a/src/pgduckdb_ruleutils.cpp
+++ b/src/pgduckdb_ruleutils.cpp
@@ -1,5 +1,6 @@
 #include "duckdb.hpp"
 #include "pgduckdb/pg/string_utils.hpp"
+#include "pgduckdb/pgduckdb_types.hpp"
 
 extern "C" {
 #include "postgres.h"
@@ -634,6 +635,15 @@ pgduckdb_get_tabledef(Oid relation_oid) {
 		}
 
 		const char *column_name = NameStr(column->attname);
+
+		/*
+		 * Check that this type is known by DuckDB, and throw the appropriate
+		 * error otherwise. This is particularly important for NUMERIC without
+		 * precision specified. Because that means something very different in
+		 * Postgres
+		 */
+		auto duck_type = pgduckdb::ConvertPostgresToDuckColumnType(column);
+		pgduckdb::GetPostgresDuckDBType(duck_type, true);
 
 		const char *column_type_name = format_type_with_typemod(column->atttypid, column->atttypmod);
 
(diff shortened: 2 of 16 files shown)


src/pgduckdb_guc.cpp
```diff
@@ -137,6 +138,10 @@ InitGUC() {
 	                     "Allow mixed transactions between DuckDB and Postgres",
 	                     &duckdb_unsafe_allow_mixed_transactions);
 
+	DefineCustomVariable("duckdb.convert_unsupported_numeric_to_double",
```
Y--, line 141: nit
```suggestion
	DefineCustomVariable("duckdb.allow_unsupported_numeric_to_double_conversion",
```

src/pgduckdb_ruleutils.cpp
```diff
@@ -635,6 +636,15 @@ pgduckdb_get_tabledef(Oid relation_oid) {
...
 		const char *column_name = NameStr(column->attname);
 
+		/*
+		 * Check that this type is known by DuckDB, and throw the appropriate
+		 * error otherwise. This is particularly important for NUMERIC without
+		 * precision specified. Because that means something very different in
+		 * Postgres
+		 */
+		auto duck_type = pgduckdb::ConvertPostgresToDuckColumnType(column);
+		pgduckdb::GetPostgresDuckDBType(duck_type, true);
```
Y--, line 646: This is a bit too bad but it looks like `GetPostgresDuckDBType` can throw cpp exceptions.. Here it's fine because `pgduckdb_get_tabledef` is only called in a method that is robust to it. But we should finish cleaning up our code everywhere to make sure this is less likely to happen ;-)

src/pgduckdb_types.cpp
```diff
@@ -1265,12 +1267,33 @@ ConvertPostgresToBaseDuckColumnType(Form_pg_attribute &attribute) {
...
 	case NUMERICARRAYOID: {
-		auto &typmod = attribute->atttypmod;
-		auto precision = numeric_typmod_precision(typmod);
-		auto scale = numeric_typmod_scale(typmod);
-		if (typmod == -1 || precision < 0 || scale < 0 || precision > 38) {
-			auto extra_type_info = duckdb::make_shared_ptr<NumericAsDouble>();
-			return duckdb::LogicalType(duckdb::LogicalTypeId::DOUBLE, std::move(extra_type_info));
+		auto precision = numeric_typmod_precision(type_modifier);
+		auto scale = numeric_typmod_scale(type_modifier);
+		if (type_modifier == -1 || precision < 1 || precision > 38 || scale < 0 || scale > 38 || scale > precision) {
```
Y--, line 1272: Would be great to replace 38 with:
```
std::numeric_limits<float>::max_exponent10
```

(also is `duckdb::LogicalTypeId::DOUBLE` backed by `float` not a `double`? that sounds surprising?)

src/pgduckdb_types.cpp
```diff
@@ -1265,12 +1267,33 @@ ConvertPostgresToBaseDuckColumnType(Form_pg_attribute &attribute) {
...
+			/* We don't allow conversion then! */
+			if (type_modifier == -1) {
+				return duckdb::LogicalType::USER(
+				    "DuckDB requires the precision of a NUMERIC to be set. You can choose to convert these NUMERICs to "
+				    "a DOUBLE by using 'SET duckdb.duckdb.convert_unsupported_numeric_to_double = true'");
+			} else if (precision < 1 || precision > 38) {
+				return duckdb::LogicalType::USER(
+				    "DuckDB only supports NUMERIC with a precision of 1-38. You can choose to convert these NUMERICs "
+				    "to a DOUBLE by using 'SET duckdb.duckdb.convert_unsupported_numeric_to_double = true'");
+			} else if (scale < 0 || scale > 38) {
```
Y--, line 1287: same here - let's use C++ [numeric limits](https://en.cppreference.com/w/cpp/types/numeric_limits)

test/regression/expected/type_support.out
```diff
@@ -326,6 +326,16 @@ INSERT INTO numeric_as_double SELECT a FROM (VALUES
     (NULL),
     (458234502034234234234.000012)
 ) t(a);
+SELECT * FROM numeric_as_double;
+WARNING:  Unsupported Postgres type: "DuckDB requires the precision of a NUMERIC to be set. You can choose to convert these NUMERICs to a DOUBLE by using 'SET duckdb.duckdb.convert_unsupported_numeric_to_double = true'"
```
Y--, line 330: This looks off no? shouldn't we remove the `"`?

src/pgduckdb_types.cpp
JelteF: the 38 has nothing to do with floats, it's because the maximum number of decimal digits that fit into an `int128` is 38: https://duckdb.org/docs/stable/sql/data_types/numeric.html#fixed-point-decimals

Y--: Ah yeah that makes more sense :) Is there a DuckDB constant we can use?

Y--: Maybe here https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/limits.hpp#L56

JelteF: Given that the number there is 39 and the docs use 38, I decided to simply refer to the docs in a code comment. Also if numerics ever get more precision using the constant you linked wouldn't help in updating it automatically, since that is linked specifically to `hugeint_t`

src/pgduckdb_ruleutils.cpp
JelteF: Changed the new code to throw cpp exceptions for these errors, at least then we're consistent.

Y-- (approved):

dentiny:
Thank you @JelteF for the help!

PR #801, opened 2025-05-23
Redesign extension installation logic

JelteF:
This PR starts auto-installing missing extensions if both of the following are true:
1. `duckdb.autoinstall_known_extensions` is set to `true`
2. The extension is listed in the `duckdb.extensions` table.

### DIFF
--- a/sql/pg_duckdb--0.3.0--1.0.0.sql
+++ b/sql/pg_duckdb--0.3.0--1.0.0.sql
@@ -64,7 +64,9 @@ LANGUAGE C;
 
 DROP FUNCTION duckdb.install_extension(TEXT);
 CREATE FUNCTION duckdb.install_extension(extension_name TEXT, source TEXT DEFAULT 'core') RETURNS void
+    SECURITY DEFINER
     SET search_path = pg_catalog, pg_temp
+    SET duckdb.force_execution = false
     LANGUAGE C AS 'MODULE_PATHNAME', 'install_extension';
 REVOKE ALL ON FUNCTION duckdb.install_extension(TEXT, TEXT) FROM PUBLIC;
 
@@ -725,4 +727,6 @@ RETURNS TEXT
 SET search_path = pg_catalog, pg_temp
 LANGUAGE C AS 'MODULE_PATHNAME', 'pgduckdb_create_azure_secret';
 
-
+ALTER TABLE duckdb.extensions ADD COLUMN repository TEXT DEFAULT 'core';
+-- TODO: Maybe rename/replace "enabled" column of extensions with autoload or
+-- something similar.
--- a/src/pgduckdb_guc.cpp
+++ b/src/pgduckdb_guc.cpp
@@ -119,7 +119,7 @@ char *duckdb_postgres_role = strdup("");
 
 int duckdb_maximum_threads = -1;
 char *duckdb_maximum_memory = strdup("4GB");
-char *duckdb_disabled_filesystems = strdup("LocalFileSystem");
+char *duckdb_disabled_filesystems = strdup("");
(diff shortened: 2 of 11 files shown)


sql/pg_duckdb--0.3.0--1.0.0.sql
```diff
@@ -64,7 +64,9 @@ LANGUAGE C;
 
 DROP FUNCTION duckdb.install_extension(TEXT);
 CREATE FUNCTION duckdb.install_extension(extension_name TEXT, source TEXT DEFAULT 'core') RETURNS void
+    SECURITY DEFINER
     SET search_path = pg_catalog, pg_temp
+    SET duckdb.force_execution = false
```
Y--, line 69: I assume it is the case - but will that be automatically reset after execution of the function?

sql/pg_duckdb--0.3.0--1.0.0.sql
```diff
@@ -725,4 +727,6 @@ RETURNS TEXT
 SET search_path = pg_catalog, pg_temp
 LANGUAGE C AS 'MODULE_PATHNAME', 'pgduckdb_create_azure_secret';
 
-
+ALTER TABLE duckdb.extensions ADD COLUMN repository TEXT DEFAULT 'core';
+-- TODO: Maybe rename/replace "enabled" column of extensions with autoload or
+-- something similar.
```
Y--, line 732: I feel we're never gonna see this TODO after 1.0 ships :-) Maybe we should move it to where the column is used?

src/pgduckdb_guc.cpp
```diff
@@ -119,7 +119,7 @@ char *duckdb_postgres_role = strdup("");
 
 int duckdb_maximum_threads = -1;
 char *duckdb_maximum_memory = strdup("4GB");
-char *duckdb_disabled_filesystems = strdup("LocalFileSystem");
```
Y--, line 122: Why don't we want that by default anymore?

src/pgduckdb_options.cpp
```diff
@@ -55,62 +55,31 @@ ReadDuckdbExtensions() {
 		bool is_null_array[Natts_duckdb_extension];
 
 		heap_deform_tuple(tuple, RelationGetDescr(duckdb_extension_relation), datum_array, is_null_array);
-		DuckdbExtension secret;
```
Y--, line 58: Copy-paste detected 😆

src/pgduckdb_options.cpp
```diff
@@ -185,7 +154,26 @@ extern "C" {
...
+
+	SPI_connect();
+	auto ret = SPI_execute_with_args(R"(
+		INSERT INTO duckdb.extensions (name, enabled, repository)
+		VALUES ($1, true, $2)
+		ON CONFLICT (name) DO UPDATE SET enabled = true
+		)",
+	                                 lengthof(arg_types), arg_types, values, NULL, false, 0);
+	if (ret != SPI_OK_INSERT)
+		elog(ERROR, "SPI_exec failed: error code %s", SPI_result_code_string(ret));
```
Y--, line 174:
```suggestion
	if (ret != SPI_OK_INSERT) {
		elog(ERROR, "SPI_exec failed: error code %s", SPI_result_code_string(ret));
	}
```
also - shouldn't we uninstall the extension? (to maintain a consistent state?

Y-- (commented):
LGTM, added minor nits and looks like the test are sad ;-)

src/pgduckdb_guc.cpp
JelteF: Moved this change to a follow up PR

PR #806, opened 2025-05-26
Enable tests

Y--:
* order tests by name in `schedule`
* enable a few tests missing from `schedule`

### DIFF
--- a/src/pg/relations.cpp
+++ b/src/pg/relations.cpp
@@ -177,4 +177,19 @@ GetOid(Form_pg_class rel) {
 	return rel->oid;
 }
 
+namespace pg {
+
+Form_pg_attribute
+GetAttributeByName(TupleDesc tupdesc, const char *colname) {
+	for (int i = 0; i < tupdesc->natts; i++) {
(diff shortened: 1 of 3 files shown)


src/pgduckdb_ruleutils.cpp
```diff
@@ -992,7 +983,6 @@ pgduckdb_get_alter_tabledef(Oid relation_oid, AlterTableStmt *alter_stmt) {
 		}
 
 		case AT_AddConstraint: {
-			pprint(cmd);
```
JelteF, line 995: Forgotten test print, which was causing issues in the tests for PG14

src/pg/relations.cpp
```diff
@@ -177,4 +177,19 @@ GetOid(Form_pg_class rel) {
 	return rel->oid;
 }
 
+namespace pg {
+
+Form_pg_attribute
+GetAttributeByName(TupleDesc tupdesc, const char *colname) {
```
JelteF, line 183: `GetAttributeByName` also existed as a function in PG14/15 (and that function has different behaviour). We now put it in a namespace to avoid this naming collision.

PR #811, opened 2025-05-28
Only show hints about duckdb.row access for queries including duckdb functions

JelteF:
We were adding hints to queries to which they did not apply. This is
unhelpful and confusing for users. So this makes some changes to our
logic for adding these hints to only do so when the query string
includes the pg_duckdb functions that return a duckdb.row.

### DIFF
--- a/src/pgduckdb_hooks.cpp
+++ b/src/pgduckdb_hooks.cpp
@@ -411,6 +411,13 @@ IsOutdatedMotherduckCatalogErrcode(int error_code) {
 	}
 }
 
+static bool
+ContainsDuckdbRowReturningFunction(const char *query_string) {
+	return strstr(query_string, "read_parquet") || strstr(query_string, "read_csv") ||
+	       strstr(query_string, "read_json") || strstr(query_string, "delta_scan") ||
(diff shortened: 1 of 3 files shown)


src/pgduckdb_hooks.cpp
```diff
@@ -411,37 +411,39 @@ IsOutdatedMotherduckCatalogErrcode(int error_code) {
...
-		    "If you use DuckDB functions like read_parquet, you need to use the r['colname'] syntax introduced "
-		    "in pg_duckdb 0.3.0. It seems like you might be using the outdated \"AS (colname coltype, ...)\" syntax");
+		if (ContainsDuckdbRowReturningFunction(debug_query_string)) {
+			/*
+			 * NOTE: We can probably remove this hint after a few releases once
+			 * we've updated all known blogposts that still used the old syntax.
+			 */
+			edata->hint = pstrdup(
+			    "If you use DuckDB functions like read_parquet, you need to use the r['colname'] syntax introduced "
+			    "in pg_duckdb 0.3.0. It seems like you might be using the outdated \"AS (colname coltype, ...)\" "
```
Y--, line 444: Not for this PR but we'll probably want to drop old version refs before going 1.0

Y-- (approved):

PR #812, opened 2025-05-31
Add sqlsmith fuzzy test

### LINKED ISSUE #222: Fuzzing using SQLsmith
Once we're unable to crash postgres easily ourselves we should use [SQLsmith](https://github.com/anse1/sqlsmith) to run a bunch of randomly generated queries. If these cause errors or don't work, that's "fine" but they shouldn't be causing crashes.

0xgouda:
This adds a test that runs sqlsmith on a sqlsmith_test database that has pg_duckdb extension enabled and any caused crashes will get detected by the logs in the teardown of pg fixture

```
...
assert (
    "was terminated by signal 6" not in logs
), "Postgres crashed! Check the logs above."
assert (
    "was terminated by signal 11" not in logs
), "Postgres crashed! Check the logs above."
```
closes: #222


0xgouda:
@JelteF 
It's been about 3 months since I opened this. Can I have a review when convenient?

Actually, the failing test after i rebased is because SQLSmith detected a crash:
```
_________________ ERROR at teardown of test_fuzzy_pg_crashing __________________
...
E   AssertionError: Postgres crashed! Check the logs above.
E   assert 'was terminated by signal 6' not in '2025-08-26 ...\tlimit 78\n'
```

JelteF:
Sorry for not responding. I've been trying to focus on getting 1.0 out (which has been going less well than intended). I'll get back to this after I've actually done that.

PR #835, opened 2025-06-11
Remove unnecessary workaround for iceberg_scan/query function

JelteF:
The issue that this code was working around has been fixed for a while
now in DuckDB: https://github.com/duckdb/duckdb-iceberg/issues/44

So this removes the now unnecessary workaround.


Y-- (approved):
Woot!

Issue #852, opened 2025-07-03
timestamp = timestamptz yields inconsistent results across SET TIME ZONE


chestnutsj:
### What happens?

In PostgreSQL, comparing a timestamp without time zone column to a timestamp with time zone column using an IN subquery returns inconsistent results depending on the SET TIME ZONE setting. This inconsistency can lead to unexpected query behavior and result set differences without any data change

```
postgres=# SET TIME ZONE 'Asia/Shanghai';
SET
postgres=# select * from t1;
 id |             d              |              tz               
----+----------------------------+-------------------------------
  1 | 2025-07-02 08:57:54.094103 | 2025-07-02 16:57:54.094103+08
  2 | 2025-07-02 22:30:15.5      | 2025-07-02 22:30:15.5+08
  3 | 2025-07-02 22:30:15.5      | 2025-07-02 22:30:15.5+08
  4 | 2025-07-02 22:30:15.5      | 2025-07-02 22:30:15.5+08
  5 | 2025-07-02 22:30:15.5      | 2025-07-02 22:30:15.5+08
(5 rows)

postgres=# select * from t1 where d in (select tz from t1);
 id |           d           |            tz            
----+-----------------------+--------------------------
  2 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
  3 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
  4 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
  5 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
(4 rows)

postgres=# select * from t1 where d in (select tz from t1);
 id |           d           |            tz            
----+-----------------------+--------------------------
  2 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
  3 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
  4 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
  5 | 2025-07-02 22:30:15.5 | 2025-07-02 22:30:15.5+08
(4 rows)

postgres=# SET TIME ZONE 'UTC';
SET
postgres=# select * from t1 where d in (select tz from t1);
 id |             d              |              tz               
----+----------------------------+-------------------------------
  1 | 2025-07-02 08:57:54.094103 | 2025-07-02 08:57:54.094103+00
(1 row)

[...]

PR #853, opened 2025-07-03
Dynamically set the time zone of duckdb

askyx:
Inconsistent time zone  between Postgres and DuckDB can lead to   
unexpected results when performing some queries which use   
timestamptz data type. To avoid this, we should keep time zone  
same during execution.

### DIFF
--- a/src/pgduckdb_guc.cpp
+++ b/src/pgduckdb_guc.cpp
@@ -2,14 +2,20 @@
 
 #include "pgduckdb/pgduckdb_duckdb.hpp"
 #include "pgduckdb/pgduckdb_guc.hpp"
+#include "pgduckdb/pgduckdb_metadata_cache.hpp"
+#include "pgduckdb/pgduckdb_utils.hpp"
+#include "pgduckdb/utility/cpp_wrapper.hpp"
 
 extern "C" {
 #include "postgres.h"
 #include "utils/guc.h"
+#include "utils/guc_tables.h"
 #include "miscadmin.h" // DataDir
 #include "lib/stringinfo.h"
 #include "postmaster/bgworker_internals.h"
 }
+
+static GucStringAssignHook prev_tz_assign_hook = nullptr;
 
 namespace pgduckdb {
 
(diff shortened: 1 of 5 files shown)


askyx:
this is fix for #852

src/pgduckdb_guc.cpp
```diff
@@ -225,4 +234,72 @@ InitGUC() {
...
+
+	for (int i = 0; i < var_count; i++) {
+		if (pg_strcasecmp(name, guc_variables[i]->name) == 0) {
+			gen = (struct config_generic *)guc_variables[i];
+			break;
+		}
+	}
+
+	if (gen && ((gen->flags & GUC_CUSTOM_PLACEHOLDER) == 0))
+		return gen;
```
JelteF, line 260: This seems unnecessary for our usecase.

src/pgduckdb_guc.cpp
```diff
@@ -225,4 +234,72 @@ InitGUC() {
...
+extern "C" {
+
+#if PG_VERSION_NUM < 170000
+
+struct config_generic *
+get_config_handle(const char *name) {
+	struct config_generic *gen = nullptr;
+#if PG_VERSION_NUM >= 160000
+	int var_count = 0;
+	auto *guc_variables = get_guc_variables(&var_count);
```
JelteF, line 246: Can you use this for loop only for versions below 16? And use `find_option` for 16 and up?

src/pgduckdb_guc.cpp
```diff
@@ -225,4 +234,72 @@ InitGUC() {
...
+		// update duckdb tz
+		auto connection = pgduckdb::DuckDBManager::GetConnection(false);
+		pgduckdb::DuckDBQueryOrThrow(*connection, "SET TimeZone =" + duckdb::KeywordHelper::WriteQuoted(tz));
+		elog(DEBUG2, "[PGDuckDB] Set DuckDB option: 'TimeZone'=%s", tz);
+	}
+}
+
+static void
+DuckAssignTimezone(const char *newval, void *extra) {
+	assign_timezone(newval, extra);
```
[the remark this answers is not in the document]
askyx: you are right, done

Issue #861, opened 2025-07-15
Error using UUID type in stored procedure


dpxcc:
### What happens?

pg_duckdb throws `Could not convert Postgres parameter of type: 2950 to DuckDB type` error when using UUID type in stored procedure or prepared statement

### To Reproduce

```
SET duckdb.force_execution TO true;

CREATE TABLE t (a int, b uuid);

INSERT INTO t VALUES (1, '82e2bd72-d25c-4562-bc61-36aa0ad15fe1'), (2, '82e2bd72-d25c-4562-bc61-36aa0ad15fe2');

SELECT * FROM t WHERE b = '82e2bd72-d25c-4562-bc61-36aa0ad15fe1';

CREATE OR REPLACE PROCEDURE get_rows_by_uuid(x UUID)
LANGUAGE plpgsql
AS $$
BEGIN
    PERFORM * FROM t WHERE b = x;
END;
$$;

CALL get_rows_by_uuid('82e2bd72-d25c-4562-bc61-36aa0ad15fe1');
```

```sql
postgres=# SET duckdb.force_execution TO true;
SET
postgres=# CREATE TABLE t (a int, b uuid);
CREATE TABLE
postgres=# INSERT INTO t VALUES (1, '82e2bd72-d25c-4562-bc61-36aa0ad15fe1'), (2, '82e2bd72-d25c-4562-bc61-36aa0ad15fe2');
INSERT 0 2
postgres=# SELECT * FROM t WHERE b = '82e2bd72-d25c-4562-bc61-36aa0ad15fe1';
 a |                  b                   
---+--------------------------------------
 1 | 82e2bd72-d25c-4562-bc61-36aa0ad15fe1
(1 row)

postgres=# CREATE OR REPLACE PROCEDURE get_rows_by_uuid(x UUID)
LANGUAGE plpgsql
AS $$
BEGIN
    PERFORM * FROM t WHERE b = x;
END;
$$;
CREATE PROCEDURE
postgres=# CALL get_rows_by_uuid('82e2bd72-d25c-4562-bc61-36aa0ad15fe1');
ERROR:  Could not convert Postgres parameter of type: 2950 to DuckDB type
CONTEXT:  SQL statement "SELECT * FROM t WHERE b = x"
PL/pgSQL function get_rows_by_uuid(uuid) line 3 at PERFORM
```

### OS:

Linux

### pg_duckdb Version (if built from source use commit hash):

pgduckdb/pgduckdb:17-main

[...]

PR #863, opened 2025-07-16
Support UUID in prepared statement

dpxcc:


JelteF (commented):
I think this looks fine, but it's failing the linting/formmatting checks. Could you update it and update the tests here to include a test for UUIDs:

https://github.com/duckdb/pg_duckdb/blob/main/test/pycheck/prepared_test.py

dpxcc:
Thanks, lint/format is fixed, and test is also added

The PR was in draft state because while working on it, I noticed that many more types, not just UUID, are not supported in prepared statement: https://github.com/duckdb/pg_duckdb/issues/861#issuecomment-3080696460
But yea, let's add support for UUID first, and I will open an issue for other types in the future

dpxcc:
https://github.com/duckdb/pg_duckdb/issues/892

JelteF (approved):

PR #871, opened 2025-07-29
Avoid extension registration check in `TimeZone` guc assign hook

YuweiXiao:


JelteF:
I think we also want to harden `IsExtensionRegistered` by returning false if `IsTransactionState()` returns false.

YuweiXiao:
> I think we also want to harden `IsExtensionRegistered` by returning false if `IsTransactionState()` returns false.

Yeah. And it is weird that the DuckDB context is still held after dropping the extension.

JelteF:
> And it is weird that the DuckDB context is still held after dropping the extension.

Yeah I think it makes sense to fix that too.

PR #875, opened 2025-07-31
Fix infinite loop when cancelling query

YuweiXiao:


JelteF:
Looks reasonable. Did you actually hit this issue in practice? Or did you simply noticed the problem in the code?

JelteF (approved):

YuweiXiao:
YES, actually hit on this and gdb shows the result is `EXECUTION_FINISHED`

PR #877, opened 2025-08-01
Disable `allow_stream_result` to force a materialized DuckDB execution results

YuweiXiao:
We don't correctly take the global Postgres lock in the main Postgres thread when fetching rows one by one from the chunks in `Duckdb_ExecCustomScan_Cpp`. Background threads might be calling postgres functions (while holding the lock), but the main thread never takes the lock so race conditions are bound to happen when that's the case. Fixing this requires some refactoring and careful testing. So for now we work around the issue by disallowing streaming results, which makes sure that the only the background threads will call Postgres functions, and the main thread will only continue running when all background threads are finished. 

Discussion: https://github.com/duckdb/pg_duckdb/discussions/866

### DIFF
--- a/src/pgduckdb_hooks.cpp
+++ b/src/pgduckdb_hooks.cpp
@@ -143,6 +143,20 @@ namespace pgduckdb {
 int64_t executor_nest_level = 0;
 
 bool
+ContainsPostgresTable(const Query *query) {
+	List *rtable = query->rtable;
+	foreach_node(RangeTblEntry, rte, rtable) {
+		if (rte->relid == InvalidOid) {
+			continue;
+		}
+		if (!::IsDuckdbTable(rte->relid)) {
+			return true;
+		}
+	}
+	return false;
+}
+
+bool
 ShouldTryToUseDuckdbExecution(Query *query) {
 	if (top_level_duckdb_ddl_type == DDLType::REFRESH_MATERIALIZED_VIEW) {
 		/* When refreshing materialized views, we only want to use DuckDB
--- a/src/pgduckdb_hooks.cpp
+++ b/src/pgduckdb_hooks.cpp
@@ -143,17 +143,34 @@ namespace pgduckdb {
 int64_t executor_nest_level = 0;
 
 bool
-ContainsPostgresTable(const Query *query) {
-	List *rtable = query->rtable;
-	foreach_node(RangeTblEntry, rte, rtable) {
-		if (rte->relid == InvalidOid) {
-			continue;
+ContainsPostgresTable(Node *node, void *context) {
+	if (node == NULL)
+		return false;
+
+	if (IsA(node, Query)) {
+		Query *query = (Query *)node;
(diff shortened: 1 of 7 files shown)


src/pgduckdb_node.cpp
```diff
@@ -193,7 +193,11 @@ ExecuteQuery(DuckdbScanState *state) {
 		named_values[duckdb::to_string(i + 1)] = duckdb::BoundParameterData(duckdb_param);
 	}
 
-	auto pending = prepared.PendingQuery(named_values, true);
+	// Always set `allow_stream_result` to false to force a fully materialized DuckDB result.
+	// This is required for cases like CTAS from a Postgres table, where allowing streaming results
+	// could lead to race conditions on Postgres resources.
+	// Checkout discussion: https://github.com/duckdb/pg_duckdb/discussions/866
+	auto pending = prepared.PendingQuery(named_values, false);
```
JelteF, line 200: I think it would be good to only do this if postgres tables are actually involved in the query that we send to duckdb. Otherwise dataloading steps (e.g. loading a few GB parquet file) will use much more memory for no real reason, because now the result cannot be streamed anymore. The only way to really know whether a postgres query is involved in the query (given that `duckdb.query` exists) is to do this detection while preparing the statement, i.e. if it the plan involves a postgres scan, then we should not allow streaming results.

src/scan/postgres_scan.cpp
```diff
@@ -422,6 +422,8 @@ PostgresScanGlobalState::PostgresScanGlobalState(Snapshot _snapshot, Relation _r
     : snapshot(_snapshot), rel(_rel), table_tuple_desc(RelationGetDescr(rel)), count_tuples_only(false),
       output_columns(), total_row_count(0), registered_local_states(0), scan_query(),
       table_reader_global_state(nullptr), duckdb_scan_memory_ctx(nullptr), max_threads(1) {
+	PostgresScopedStackReset scoped_stack_reset;
+	std::lock_guard<std::recursive_mutex> lock(GlobalProcessLock::GetLock());
```
JelteF, line 426: Why are these changes suddenly needed? (as well as the one below in `PostgresScanFunction`) It seems like these additions could very well be the cause for the test failures in CI.

src/pgduckdb_node.cpp
YuweiXiao: I added a check on the rtables of the Query instance. And as long as a Postgres table is referenced, the stream is disabled.

src/pgduckdb_hooks.cpp
```diff
@@ -149,6 +149,20 @@ namespace pgduckdb {
 
 int64_t executor_nest_level = 0;
 
+bool
+ContainsPostgresTable(const Query *query) {
+	List *rtable = query->rtable;
```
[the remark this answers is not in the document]
JelteF: @YuweiXiao are you still fixing this?

JelteF (approved):

PR #764, opened 2025-05-07
Disallow DuckDB execution inside functions

JelteF:
pg_duckdb changes result types of queries at the planning stage. This is needed for pg_duckdb to function correctly. However, Postgres code is sometimes not written with this in mind. One such case is sql functions, as reported in #749. There it causes a crash, because Postgres expects a different type to be returned by the function than what is actually returned.

As far as I can tell there is no easy way of fixing this, apart from changing Postgres to learn how to deal with this. I'll open issue/patch to Postgres for this at some point in the near future.

For now, we simply start disallowing all DuckDB execution inside functions. That is quite a shame because using DuckDB execution in functions works fine in many other cases. Sadly, I cannot think of a way to only disallow the crashing cases. Since a very important part of pg_duckdb 1.0 is stability, it seems better to err on the side of caution and disable this feature for now and stabilize it in a future release.

Fixes #749

PR #884, opened 2025-08-09
Add GUC to allow DuckDB execution inside functions

dpxcc:
DuckDB execution inside functions was disallowed in https://github.com/duckdb/pg_duckdb/pull/764 for good reason
However, some users depend on running analytics with pg_duckdb inside functions

Given that (1) the problematic cases are rare, and (2) pg_duckdb is mostly used for read queries, failures are limited to crashing the current session and can be recovered by restarting the session without data loss. This change introduces an unsafe GUC to restore this capability for users who need it

### DIFF
--- a/src/pgduckdb_guc.cpp
+++ b/src/pgduckdb_guc.cpp
@@ -116,6 +116,7 @@ DefineCustomDuckDBVariable(const char *name, const char *short_desc, T *var, T m
 } // namespace
 
 bool duckdb_force_execution = false;
+bool duckdb_unsafe_allow_execution_inside_functions = false;
 bool duckdb_unsafe_allow_mixed_transactions = false;
 bool duckdb_convert_unsupported_numeric_to_double = false;
 bool duckdb_log_pg_explain = false;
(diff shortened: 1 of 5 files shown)


src/pgduckdb_guc.cpp
```diff
@@ -143,6 +144,9 @@ InitGUC() {
 	/* pg_duckdb specific GUCs */
 	DefineCustomVariable("duckdb.force_execution", "Force queries to use DuckDB execution", &duckdb_force_execution);
 
+	DefineCustomVariable("duckdb.unsafe_allow_execution_inside_functions", "Allow DuckDB execution inside functions",
+	                     &duckdb_unsafe_allow_execution_inside_functions);
```
JelteF, line 148: Given that this can introduce crashes I think a superuser should be configuring this.  

```suggestion
	                     &duckdb_unsafe_allow_execution_inside_functions, 0, GUC_SUPERUSER_ONLY);
```

test/regression/expected/transactions.out
```diff
@@ -117,9 +118,8 @@ $$;
 -- After executing the function the table should not contain the value 8,
 -- because that change was rolled back
 SELECT * FROM f(true);
-ERROR:  DuckDB execution is not supported inside functions
```
JelteF, line 120: Do we still have some tests for the default behaviour?

dpxcc: Yes, `test/regression/expected/issue_730.out` still uses the default behavior
https://github.com/duckdb/pg_duckdb/pull/764/files#diff-6e283896067d2e581c1917f5595977aa405d8f337cc77303a72e49dd9463eb26

JelteF (approved):

PR #887, opened 2025-08-17
Fix multiple duckdb instances clash on the same temporary directory.

dursegov:
Multiple in-memory duckdb instances may clash on the same temporary directory. This PR fixes the issue on pg_duckdb side until https://github.com/duckdb/duckdb/issues/15173 is resolved.


fpapadamis:
Looking forward to that!
If this could be added to `v1.2.0` it would be great.

PR #893, opened 2025-08-20
Add support for subextensions

dpxcc:
This PR adds support for subextensions in pg_duckdb. Subextensions can now register their own TAMs to trigger DuckDB execution, similar to how DuckDB temp tables are handled:
1. Subextensions can call `RegisterDuckdbTableAm()` to register their DuckDB-specific TAMs
2. During query planning, pg_duckdb calls `IsDuckdbTable()` in `src/pgduckdb_hooks.cpp` to check if it's a DuckDB-specific TAM. A query needs DuckDB execution when there are any DuckDB-specific TAMs
3. When translating Postgres queries to DuckDB queries, a Postgres table `<schema>.<table>` with DuckDB-specific TAM `<tam>` will be translated to a DuckDB table `<tam>.<schema>.<table>` (special handling for DuckDB temp table is preserved)
4. Subextensions can create a DuckDB storage extension that attaches to the `<tam>` database, enabling custom scans against their DuckDB-specific TAMs

Even better, the DuckDB extension can detect at runtime whether it is running inside Postgres with pg_duckdb or in standalone DuckDB, and adjust its logic accordingly. This allows the same extension to be used seamlessly in both environments


dpxcc:
Here is a full working example of basing pg_mooncake on top of pg_duckdb:
pg_mooncake: https://github.com/Mooncake-Labs/pg_mooncake/tree/9a9cf080f58c8ac9539effbac85102cb5396ddd8
duckdb_mooncake: https://github.com/Mooncake-Labs/duckdb_mooncake/tree/3896a13bdb54f4f36a494d39893aecdd1d80e027

src/pgduckdb_table_am.cpp
```diff
@@ -492,9 +495,35 @@ duckdb_am_handler(FunctionCallInfo /*funcinfo*/) {
 }
 }
 
+static duckdb::unordered_map<const TableAmRoutine * /*am*/, duckdb::string /*name*/> duckdb_table_ams = {
+    {&duckdb_methods, "duckdb"}};
+
+extern "C" __attribute__((visibility("default"))) void
+RegisterDuckdbTableAm(const char *name, const TableAmRoutine *am) {
+	duckdb_table_ams[am] = name;
```
JelteF, line 503: Let's check first if it's already in there, so extensions cannot accidentally override already existing ones.

JelteF (approved):

PR #895, opened 2025-08-21
Grant foreign data wrapper duckdb to PUBLIC

YuweiXiao:
To enable normal user to user motherduck and manage secret


YuweiXiao:
agreed

Issue #848, opened 2025-06-19
Build fails on GCC 8 due to missing -lstdc++fs when using std::filesystem


liuchuang001:
### What happens?

When building pg_duckdb on systems using GCC 8 (e.g. CentOS 8), the PostgreSQL extension fails to load due to unresolved std::filesystem symbols.

GCC 8 supports std::filesystem but requires explicitly linking against libstdc++fs. This is no longer necessary in GCC 9+, where filesystem was merged into the standard C++ library.

### To Reproduce

On a system with GCC 8.x (e.g., CentOS 8):

Build pg_duckdb using the current Makefile

Load the extension in PostgreSQL:

sql
```
LOAD 'pg_duckdb';
```
You will encounter an error like:
```
undefined symbol: _ZNSt10filesystem7__cxx114path14_M_split_cmptsEv
```

Diagnosis:

Inspecting pg_duckdb.so with nm -D shows unresolved symbols related to std::filesystem:
```
nm -D pg_duckdb.so | grep filesystem
```
```
# Output:
# U _ZNSt10filesystem7__cxx114path14_M_split_cmptsEv
# U _ZNSt10filesystem18create_directoriesERKNS_7__cxx114pathE
```

These symbols are undefined unless -lstdc++fs is explicitly passed during linking on GCC 8.

Suggested Fix:

Modify the Makefile to detect if GCC version is 8 and add -lstdc++fs accordingly:

```
GCC_VERSION_MAJOR := $(shell g++ -dumpversion | cut -f1 -d.)
ifeq ($(shell [ $(GCC_VERSION_MAJOR) -eq 8 ] && echo true),true)
  PG_DUCKDB_LINK_FLAGS += -lstdc++fs
endif
```
This keeps the build compatible across all compiler versions:

GCC < 8: no change (but std::filesystem isn't usable anyway)

GCC 8: adds required -lstdc++fs

GCC ≥ 9: safely ignored

Let me know what you think!
Thanks for all the amazing work on pg_duckdb! 🙌

### OS:

linux centos8 x86

### pg_duckdb Version (if built from source use commit hash):

12dee2e43520144059f007d37ce4863b9c879d31

### Postgres Version (if built from source use commit hash):

16.9

### Full Name:

liu chuang

### Affiliation:

liuchuang66888@gmail.com

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

[...]

PR #920, opened 2025-09-05
Fix undefined symbol error on GCC 8 build

clayoster:
This resolves the following undefined symbol error seen when building the plugin with GCC 8 on Rocky 8 (and other RHEL 8 variants)

```
undefined symbol: _ZNSt10filesystem7__cxx114path14_M_split_cmptsEv
```

This fix was pulled from #848.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,12 @@ endif
 
 PG_DUCKDB_LINK_FLAGS += -Wl,-rpath,$(PG_LIB)/ -L$(DUCKDB_BUILD_DIR)/src -L$(PG_LIB) -lstdc++ -llz4
 
+# Ensure -lstdcc++fs is included for GCC 8 builds
+GCC_VERSION_MAJOR := $(shell g++ -dumpversion | cut -f1 -d.)
+ifeq ($(shell [ $(GCC_VERSION_MAJOR) -eq 8 ] && echo true),true)
+  PG_DUCKDB_LINK_FLAGS += -lstdc++fs
+endif
+
 ERROR_ON_WARNING ?=
 ifeq ($(ERROR_ON_WARNING), 1)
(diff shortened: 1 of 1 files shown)


JelteF (approved):

Makefile
```diff
@@ -51,6 +51,12 @@ endif
 
 PG_DUCKDB_LINK_FLAGS += -Wl,-rpath,$(PG_LIB)/ -L$(DUCKDB_BUILD_DIR)/src -L$(PG_LIB) -lstdc++ -llz4
 
+# Ensure -lstdcc++fs is included for GCC 8 builds
+GCC_VERSION_MAJOR := $(shell g++ -dumpversion | cut -f1 -d.)
+ifeq ($(shell [ $(GCC_VERSION_MAJOR) -eq 8 ] && echo true),true)
+  PG_DUCKDB_LINK_FLAGS += -lstdc++fs
+endif
```
JelteF, line 58: This should be using CXX instead of `g++`. Otherwise if you override the default compiler to a newer one you'd still get this flag added. I think something like this should work:
```suggestion
# Ensure -lstdcc++fs is included for GCC 8 builds
CXX_IS_GCC8 := $(shell $(CXX) --version 2>/dev/null | head -n1 | grep -q "g++.*\s8\." && echo true || echo false)
ifeq ($(CXX_IS_GCC8),true)
  PG_DUCKDB_LINK_FLAGS += -lstdc++fs
endif
```

JelteF (changes requested):
Can you update and try with the suggested snippet to see if it works for you?

clayoster:
I have tested and replied to your suggestion above.

clayoster:
The last suggestion tested successfully! I have updated the pull request with that change.

PR #935, opened 2025-09-22
Align the semantic of nulls order across PostgreSQL and DuckDB

YuweiXiao:
In Postgres, the ordering behavior of NULL values is as follows:
- `SELECT * FROM t1 ORDER BY t1` is equivalent to `SELECT * FROM t1 ORDER BY t1 ASC NULLS LAST`.
- `SELECT * FROM t1 ORDER BY t1 DESC` is equivalent to `SELECT * FROM t1 ORDER BY t1 DESC NULLS FIRST`.

Conversely, in DuckDB (by default):
- `SELECT * FROM t1 ORDER BY t1` is equivalent to `SELECT * FROM t1 ORDER BY t1 ASC NULLS LAST`. (Correct)
- `SELECT * FROM t1 ORDER BY t1 DESC` is equivalent to `SELECT * FROM t1 ORDER BY t1 DESC NULLS LAST`. (Inconsistent with PG)

This PR has DuckDB to be consistent with Postgres by explicitly setting `default_null_order=postgres` when initialize DuckDB session.

### DIFF
--- a/src/pgduckdb_duckdb.cpp
+++ b/src/pgduckdb_duckdb.cpp
@@ -98,6 +98,7 @@ DuckDBManager::Initialize() {
 
 	duckdb::DBConfig config;
 	config.SetOptionByName("custom_user_agent", "pg_duckdb");
+	config.SetOptionByName("default_null_order", "postgres");
 
 	SET_DUCKDB_OPTION(allow_unsigned_extensions);
 	SET_DUCKDB_OPTION(enable_external_access);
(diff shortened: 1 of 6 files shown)


JelteF:
Nice catch. Let's instead configure `default_null_order` though

YuweiXiao:
> Nice catch. Let's instead configure `default_null_order` though

Ah, I overlooked the config. But it appears that the configuration solely pertains to nulls_order and is not affected by the ASC/DESC order, i.e., regardless of the value assigned to `default_null_order`, there will be inconsistency.

JelteF:
There's `default_null_order=postgres`

PR #945, opened 2025-10-03
Add docs and clarifications around changing the default database

leonardovida:
Add docs around:
- Changing the default database: disconnect and reconnect
- Clarifying that all MD tables are accessible even if the default motherduck database is specified


leonardovida:
Bumping this one @JelteF

PR #949, opened 2025-10-09
feat(planner): check postgres user rights for view in duckdb planner

### LINKED ISSUE #948: Accessing views from pg_duckdb is inconsistent with PostgreSQL
### What happens?

In PostgreSQL, when accessing views, rights of user are checked in planner. 
It is possible to set different rights for different users.
Here's the example:
```sql
postgres=# create user user1;
CREATE ROLE
postgres=# create table test(a int);
CREATE TABLE
postgres=# insert into test (a) values (1);
INSERT 0 1
postgres=# create view pgv as select * from test ;
CREATE VIEW

postgres=# set role user1 ;
SET
postgres=> select * from pgv;
ERROR:  permission denied for view pgv

postgres=> set role jorq;
SET
postgres=# grant select on pgv to user1 ;
GRANT

postgres=# set role user1;
SET
postgres=> select * from pgv;
 a
---
 1
(1 row)
```

However, in pg_duckdb planner they are not checked. 

Due to that any view that calls duckdb can be accessed by any user, and only postgres_role matters here. 
GRANT / REVOKE on these views are useless. This is inconsistent to recognized PostgreSQL behavior

### To Reproduce

(see above)

### OS:

Linux

main

17

### Hardware:

_No response_

### Full Name:

Jora Babayan

### Affiliation:

Postgres Professional

I have not tested with any build

No - Other reason (please specify in the issue body)

- [ ] Yes, I have

- [ ] Yes, I have

ggnmstr:
resolves #948

### DIFF
--- a/test/regression/expected/duckdb_views.out
+++ b/test/regression/expected/duckdb_views.out
@@ -1,12 +1,41 @@
 CREATE USER duckdb_view_admin IN ROLE duckdb_group;
+GRANT ALL ON SCHEMA public TO duckdb_view_admin;
+CREATE USER duckdb_view_user1 IN ROLE duckdb_group;
 SET ROLE duckdb_view_admin;
-CREATE USER duckdb_view_user1;
-CREATE VIEW duckdb_view AS SELECT * FROM read_parquet('/nonexisting.parquet');
+CREATE VIEW duckdb_view AS SELECT r['a']::int as a FROM duckdb.query($$ SELECT 1 a $$) r;
+CREATE VIEW postgres_view AS SELECT 2 from generate_series(1,2);
 SET ROLE duckdb_view_user1;
-SELECT * from duckdb_view LIMIT 1;
+SELECT * from duckdb_view;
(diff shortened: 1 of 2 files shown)


ggnmstr:
@JelteF, sure. I updated patch to fit different PostgreSQL versions and created a simple regression test.

src/pgduckdb_planner.cpp
```diff
@@ -135,6 +143,35 @@ DuckdbRangeTableEntry(CustomScan *custom_scan) {
 
 PlannedStmt *
 DuckdbPlanNode(Query *parse, int cursor_options, bool throw_error) {
+
+	/* If we're accessing from view, check
+	 * that Postgres user is allowed to access it */
+
+	ListCell *l;
+	foreach (l, parse->rtable) {
```
JelteF: I expect this not to error for views in subqueries? I think you'll need to walk the query tree like is done for `ContainsDuckdbItems`.

PR #979, opened 2025-11-21
Convert DuckDB VARCHAR[] to PG TEXT[] to Prevent Type Mismatch

YuweiXiao:
DuckDB’s VARCHAR type is mapped to PostgreSQL TEXT, but its VARCHAR[] arrays were previously mapped to PostgreSQL VARCHAR[]. This inconsistency causes a type mismatch on the Postgres side during statistics analysis, triggering an assertion in `array_typeanalyze.c::compute_array_stats`:
```
Assert(ARR_ELEMTYPE(array) == extra_data->type_id)
```

This PR standardizes the behavior by converting DuckDB VARCHAR[] to PostgreSQL TEXT[].

A sample dataset is included to reproduce the original failure.


JelteF:
Thanks for the report and the fix. I updated and extended the tests a bit.

PR #991, opened 2025-12-15
chore: upgrade to debian 13

reneleonhardt:
### Chores
- Upgrade Debian 12 to 13
- Let Dependabot update docker-compose, github-actions and gitsubmodule

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: docker-compose
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: github-actions
+    directory: /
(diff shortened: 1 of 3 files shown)


reneleonhardt:
### Notes
- I removed `docker` and `pip` from dependabot.yaml as they contain no versioned dependencies at the moment.
- docker-bake.hcl has to be updated manually to Debian `Forky` in 2027.
- Could you tag releases too like your floating `18-main` tag, for example `18-v1` to follow the latest release instead of the `main` branch?
- If possible, your CI shouldn't require credentials to run; otherwise they should be declared as secrets.

https://github.com/duckdb/pg_duckdb/actions/runs/20245314917/job/58123710291?pr=991#step:2:9

PR #1021, opened 2026-03-01
fix: prevent redefine FirstCommandId macro with PostgreSQL

mrdrivingduck:
Fix:

```
In file included from ../../src/include/postgres.h:45,
                 from src/pgduckdb_hooks.cpp:11:
../../src/include/c.h:668: warning: "FirstCommandId" redefined
  668 | #define FirstCommandId ((CommandId) 0)
      |
```


JelteF:
When do you run into this?

JelteF:
It seems this is already guarded against this on main

PR #1023, opened 2026-03-08
feat: allow installation in a non-public schema

TylerHillery:
## Overview

Removes `schema = public` from `pg_duckdb.control` to allow installing the extension in a non-public schema.

Default behavior is unchanged, `CREATE EXTENSION pg_duckdb` without a `SCHEMA` clause still installs into `public`.

In general I agree with the original sentiment:
```
It would be sad for usability if people would have to prefix those with duckdb.read_csv
```

but it would be nice if users can configure this if they want

### DIFF
--- a/pg_duckdb.control
+++ b/pg_duckdb.control
@@ -2,4 +2,3 @@ comment = 'DuckDB Embedded in Postgres'
 default_version = '1.1.0'
 module_pathname = '$libdir/pg_duckdb'
 relocatable = false
-schema = public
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
Thanks for addressing this @TylerHillery.

For what it's worth, I'm in the "schema-qualify-everything" camp, so I don't think it "...would be sad for usability if people would have to prefix those with duckdb...".  I think that's ideal.  ;)

Regarding your note over on the discussion (https://github.com/duckdb/pg_duckdb/discussions/1013#discussioncomment-16083100), I don't think it's a problem that it's not relocatable.  This should give us the ability to let users place it where they need it up-front without also needing to worry about what the may have already done with their timsescale extension.

Much appreciated!

JelteF:
I thought you'd have to always specify a schema in this case. But that's not the case. So this seems fine (i.e. still nice usability by default, but freedom for people to use other schemas)

PR #1025, opened 2026-03-13
Bump up duckdb to v1.5.4

qsliu2017:
Bump DuckDB from v1.4.3 to v1.5.4.

## Key Changes

1. DuckDB v1.5.0 removed direct field access on `DBConfigOptions` for several settings. Migrated all usages to `config.SetOptionByName()`.

2. Storage and optimizer extensions can no longer be registered by directly pushing into `dbconfig.{storage_extensions,optimizer_extensions}`. Migrated to the static methods
  `duckdb::{StorageExtension,OptimizerExtension}::Register`

3. DuckDB v1.5.0 enforces signature verification on `INSTALL ... FROM community`, not just on LOAD. Updated non_superuser_test.py to enable `allow_community_extensions` before installing.

4. Test output updates
- Cosmetic changes in v1.5 error messages
- Added alternative expected output (json_functions_duckdb_1.out) for PG 17 where function names are quoted in error contexts due to JSON_EXISTS being a keyword

## New Features

1. GEOMETRY type conversion
  Map DuckDB's GEOMETRY type to PostgreSQL TEXT so values are returned as human-readable Well-Known Text (e.g. `POINT EMPTY`, `GEOMETRYCOLLECTION (...)`) instead of raw WKB bytes.

## Bug Fixes

1. `PostgresScanGlobalState::ExtractQueryFilters` OR with unsupported children
  Dropping a child of an OR makes the filter more restrictive, which silently narrows results. Previously this was guarded by `is_inside_optional_filter`, but the logic is wrong regardless of context. Now the entire OR is dropped whenever any child can't be pushed down. This bug existed before v1.5 but was never triggered until the `TOP_N` optimizer started pushing `optional: id IS NULL OR Dynamic Filter(id)` down to scans.

## Temporary Workarounds

1. Skips `quack` community extension tests (not yet published for v1.5.3).


test/regression/expected/extensions.out
```diff
@@ -171,13 +171,13 @@ SELECT * FROM duckdb.query($$ SELECT extension_name, loaded, installed, installe
...
  httpfs         | t      | t         | 
- icu            | t      | t         | 
+ icu            | t      | t         | core
  json           | t      | t         | 
  parquet        | t      | t         | 
  pgduckdb       | t      | f         | 
 (6 rows)
 
-SELECT duckdb.install_extension('prql', 'community');
+SELECT duckdb.install_extension('quack', 'community');
```
qsliu2017: prql have no 1.5.0 version yet. use quack instead.

test/regression/expected/json_functions_duckdb.out
```diff
@@ -37,6 +37,9 @@ SELECT json_exists('{"a": {"b": 1}}', '$.a.c') FROM duckdb.query($$ SELECT 1 $$)
 SELECT public.json_exists('{"a": {"b": 1}}'::jsonb, '$.a.c');
 ERROR:  (PGDuckDB/CreatePlan) Prepared query returned an error: Catalog Error: Type with name jsonb does not exist!
 Did you mean "blob"?
+
+LINE 1: SELECT system.main.json_exists(('{"a": {"b": 1}}'::jsonb), '$.a.c'::character varying) AS json_exists
```
qsliu2017: PG>=17 would be

```
LINE 1: SELECT system.main."json_exists"(('{"a": {"b": 1}}'::jsonb), '$.a.c'::character varying) AS "json_exists"
```

i'm not sure if we can ignore this line. but I add `test/regression/expected/json_functions_duckdb_1.out` to be an alternative match.

qsliu2017:
seems like we don't need `PostgresScanSerialize` / `PostgresScanDeserialize` in next duckdb minor version https://github.com/duckdb/duckdb/pull/21461

zhaoxinyu:
@qsliu2017 ,Is this PR ready to be merged?

zhaoxinyu:
@JelteF, Hi, Jelte, Could you please review this MR? duckdb 1.5 contains a lot of exciting improvements.

qsliu2017:
> @qsliu2017 ,Is this PR ready to be merged?

yes. I've run this version in my dev for a while, and so far so good. if you're rush to duckdb 1.5, think you can just use my branch. feedback is welcome.

qsliu2017:
the only failed CI is an issue from duckdb-httpfs (see https://github.com/duckdb/duckdb-httpfs/pull/308).

test/regression/expected/extensions.out
qsliu2017: oh woah, quack is not an example extension at all. DuckDB guys occupy this name for this secret project.

https://duckdb.org/2026/05/12/quack-remote-protocol

kinghuang:
Who can review this PR? @qsliu2017 has been diligently updating the PR. But, there doesn't seem to be any notice from eligible reviewers.

JelteF:
Thank you for this! And sorry for the slow response. Merged as part of: [02899db](https://github.com/duckdb/pg_duckdb/commit/02899dbbb2ac9ab7b956b5c399a073eb831ee9b8)

Issue #892, opened 2025-08-20
Support more types in prepared statement


dpxcc:
### What happens?

While working on https://github.com/duckdb/pg_duckdb/pull/863, I noticed that several other data types, besides UUID, are currently unsupported in prepared statements

Specifically, those types exist in `ConvertPostgresToDuckValue`, but not in `ConvertPostgresParameterToDuckValue`: https://github.com/duckdb/pg_duckdb/blob/main/src/pgduckdb_types.cpp

### To Reproduce

n/a

### OS:

Linux

### pg_duckdb Version (if built from source use commit hash):

55874e3

### Postgres Version (if built from source use commit hash):

17.6

### Full Name:

Cheng Chen

### Affiliation:

Mooncake Labs

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have not tested with any build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

cburyta:
A colleague hit this limitation on a project, and I had been chipping away at adding the feature. Finally got to pushing a PR, details in the pull request description though.

PR #1032, opened 2026-04-09
fix: improve prepared statement parameter support

cburyta:
## Summary

Adds support for several PostgreSQL data types missing from `ConvertPostgresParameterToDuckValue`, fixing #892. Also addresses related issues discovered during integration testing with prepared statements over `read_parquet()` via the extended query protocol.

### Parameter conversion (`src/pgduckdb_types.cpp`)
- Add NUMERIC to DuckDB DECIMAL conversion with precision/scale inferred from the value
- Add array parameter support: INT2[], INT4[], INT8[], FLOAT4[], FLOAT8[], TEXT[], VARCHAR[], BPCHAR[]
- Fix uint8_t overflow in NUMERIC precision calculation where values with 256+ integral digits silently wrapped precision to 0, bypassing the 38-digit maximum clamp

### Result type mapping (`src/pgduckdb_types.cpp`)
- Map DuckDB `UNKNOWN` result type to PostgreSQL `TEXT` in CreatePlan, preventing `Could not convert DuckDB type: UNKNOWN to Postgres type` errors on queries with select-list parameters

### Prepared statement stability (`src/pgduckdb_node.cpp`, `src/pgduckdb_planner.cpp`)
- Stabilize select-list parameter schema across prepare and execute phases
- Add inline-parameter fallback when column count drifts between planning and execution

### Deparse fixes (`src/vendor/pg_ruleutils_{14,15,16,17,18}.c`)
- Handle `duckdb.unresolved_type` parameters in parquet predicate deparse
- Add `int4` assignment cast for `duckdb.unresolved_type` to enable integer literal binding in BETWEEN/IN predicates
- Expand `ScalarArrayOpExpr` as `IN (...)` instead of `= ANY(ARRAY[...])` when departing for DuckDB, preventing column-count mismatch in prepared statements

### Tests (`test/pycheck/prepared_test.py`)
- Regression tests for parquet untyped/typed bind parameters (BETWEEN, IN)
- Native PREPARE/EXECUTE protocol tests
- Unsupported parameter type failure test

## Type coverage

| Type | OID | Before | After |
|------|-----|--------|-------|
| NUMERIC | 1700 | `Could not convert` | PASS |
| SMALLINT[] | 1005 | `Could not convert` | PASS |
| INTEGER[] | 1007 | `Could not convert` | PASS |
| BIGINT[] | 1016 | `Could not convert` | PASS |
| FLOAT4[] | 1021 | `Could not convert` | PASS |
| FLOAT8[] | 1022 | `Could not convert` | PASS |
| TEXT[] | 1009 | `Could not convert` | PASS |
| VARCHAR[] | 1015 | `Could not convert` | PASS |

### Extension-level tests

All tests run via prepared statements with `duckdb.force_execution = true`:

| Test suite | Unpatched v1.1.1 | Patched |
|-----------|------------------|---------|
| Prepared statement types (31 types) | 15/31 pass | **31/31 pass** |
| Issue #892 gate (8 required types) | 0/8 | **8/8** |
| Typed bind params — parquet BETWEEN/IN (20 cases) | 14/20 | **20/20** |
| Select-list parameters (6 cases) | FAIL (crash) | **6/6 pass** |
| UNKNOWN result type mapping | FAIL | **PASS** |
| Column-count mismatch harness | 0 mismatch hits | 0 mismatch hits |
| TIMESTAMPTZ[] via native PREPARE | PASS | PASS |

### Application-level validation

Additionally validated against a real-world production application that executes complex `read_parquet()` queries with 20+ bind parameters via the extended query protocol (the app used a Node.js based Knex/pg driver).

### DIFF
--- a/sql/pg_duckdb--1.0.0--1.1.0.sql
+++ b/sql/pg_duckdb--1.0.0--1.1.0.sql
@@ -1,4 +1,8 @@
 -- Add MAP functions support
+-- Allow native PREPARE/EXECUTE integer literals to bind against unresolved parquet predicates.
+-- The unresolved type lives in the duckdb schema.
+CREATE CAST (integer AS duckdb.unresolved_type) WITH INOUT AS ASSIGNMENT;
+
 -- Extract value from map using key
 CREATE FUNCTION @extschema@.map_extract(map_col duckdb.map, key "any")
 RETURNS duckdb.unresolved_type AS 'MODULE_PATHNAME', 'duckdb_only_function' LANGUAGE C;
--- a/src/pgduckdb_node.cpp
+++ b/src/pgduckdb_node.cpp
@@ -7,7 +7,6 @@
 
 #include "pgduckdb/pgduckdb_hooks.hpp"
 #include "pgduckdb/pgduckdb_planner.hpp"
-#include "pgduckdb/pgduckdb_ruleutils.h"
 #include "pgduckdb/pgduckdb_types.hpp"
 #include "pgduckdb/vendor/pg_explain.hpp"
 #include "pgduckdb/pg/explain.hpp"
@@ -18,6 +17,8 @@ extern "C" {
 #include "tcop/pquery.h"
 #include "nodes/params.h"
 #include "utils/ruleutils.h"
+
+char *pgduckdb_get_querydef(Query *query);
 }
 
 #include "pgduckdb/pgduckdb_node.hpp"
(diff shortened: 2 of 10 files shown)


cburyta:
Heads up, I believe PR #1033 (`fix: suppress ::numeric cast for unqualified NUMERIC constants`) is tackling the same NUMERIC/DECIMAL mismatch from the deparse side, while this PR addresses it on the parameter binding side.

The two approaches are complementary: #1033 suppresses the ::numeric annotation when typmod is -1 so DuckDB can infer type from context, and this PR adds explicit NUMERIC to DECIMAL conversion with inferred precision/scale during parameter binding. 

They touch adjacent files (vendor `pg_ruleutils_*.c` here vs. `pgduckdb_ruleutils.cpp` in #1033), so merge conflicts should be minimal, but whoever lands second will want to verify the combined behavior... particularly around unqualified NUMERIC constants in prepared statement parameters.

Happy to coordinate on merge order if that's helpful, but I'm mostly hoping for input from maintainers.

cburyta:
Unsure if there's anything I should do to follow up, request a PR review or otherwise. Let me know if you have time please, thank you! cc @JelteF @Y-- 🙏

PR #1038, opened 2026-04-17
add postgres wrapper for `read_vortex` from vortex core extension

a10y:
Similar to `read_paruet` and `iceberg_scan`, this PR adds a new Postgres function wrapper for calling the `read_vortex()` DuckDB table function.

See https://duckdb.org/docs/current/core_extensions/vortex for documentation on the extension.


test/regression/data/unsigned_types.vortex
a10y, line 1: this is just conversion of the in-tree single-row `unsigned_types.parquet` file.

As you can see, Vortex is not optimized for small data 😅

sql/pg_duckdb--1.1.0--1.2.0.sql
```diff
@@ -0,0 +1,6 @@
+-- read_vortex function
```
a10y, line 1: let me know if this is the correct file to put this in. I do not have a ton of experience with postgres plugin dev

sql/pg_duckdb--1.1.0--1.2.0.sql
```diff
@@ -0,0 +1,6 @@
...
+CREATE FUNCTION @extschema@.read_vortex(path text)
```
a10y, line 2: also note that in the next duckdb release we should support taking an array of paths, similar to `read_parquet`, but that is not the case on any currently released version yet.

JelteF (approved):
Thank you!

PR #1045, opened 2026-04-28
revoke PUBLIC execute on raw_query

x4m:
raw_query(TEXT) executes arbitrary DuckDB SQL, including HTTP requests to any host and access to any S3 bucket the server can reach.  Leaving it callable by public might be dangerous.

I suppose the REVOKE on line 123 was inadvertently overridden by a later GRANT ALL TO PUBLIC near the aggregate definitions. Or do I misunderstand something?

I'm not sure, but maybe recycle_ddb() also deserves same treating.

### DIFF
--- a/sql/pg_duckdb--1.0.0.sql
+++ b/sql/pg_duckdb--1.0.0.sql
@@ -1120,7 +1120,10 @@ CREATE AGGREGATE @extschema@.json_group_structure(duckdb.json)
     initcond = 0
 );
 
-GRANT ALL ON FUNCTION duckdb.raw_query(TEXT) TO PUBLIC;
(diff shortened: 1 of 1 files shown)


JelteF:
Instead of blocking running of raw queries, I believe pg_duckdb should be configuring DuckDB in such a way that raw sql access to duckdb isn't a security issue. As I just responded to https://github.com/duckdb/pg_duckdb/pull/1046 http/s3 request filtering can now be done (on `main`) using `duckdb.allowed_directories`. If you find more issues with with raw sql allowing security issues, feel free to report them, so pg_duckdb can lock them down (either by default, or provide knobs to admins to do so).

PR #1024, opened 2026-03-11
Add allowed_directories to GUC

adam-lithus:
Similar to #884 and #699, this adds `allowed_directories` to the GUC.

Note that DuckDB rejects changes to `allowed_directories` after external access is disabled. I've therefore moved `enable_external_access` from `DBConfig` struct to a SQL SET in Initialize() so `allowed_directories` can be applied first. Let me know if this is not desirable.

Update: [CI is passing on our side](https://github.com/golithus/pg_duckdb/actions/runs/22954773102)

PR #1046, opened 2026-04-28
add duckdb.allowed_endpoint_suffixes GUC

x4m:
Hi!

I'm a bit afraid of SSRF from my clusters. 
I have a wild idea to make a GUC to control where DuckDB can go. Like

```
duckdb.allowed_endpoint_suffixes = 'storage.yandexcloud.net,s3.amazonaws.com'
```

WDYT?


JelteF:
DuckDB already provides a setting similar to this, which was added by https://github.com/duckdb/pg_duckdb/commit/d5b95c4b76fedd80de3656e34794abbb1a869a96 (based on #1024). I believe that should be enough to protect yourself against SSRF.

PR #1061, opened 2026-07-20
Remove default password from docker-compose.yml

arijitroy003:
## Summary
- Replace the hardcoded default password (`duckdb`) in `docker-compose.yml` with a required environment variable check using `${PASSWORD:?...}`
- Docker Compose now fails fast with a clear error message if `PASSWORD` is not set, preventing accidental deployments with a well-known credential
- Added a comment explaining how to set the variable

## Test plan
- [ ] Run `docker compose config` without `PASSWORD` set — should error with "Set PASSWORD env var before running"
- [ ] Run `PASSWORD=mysecret docker compose config` — should render the compose file with the password set


arijitroy003:
Friendly ping — is there anything needed from my side to move this forward? Happy to make any adjustments.

arijitroy003:
Another bump — this removes the default password from docker-compose.yml (security hygiene). CI passes on all versions except REL_19_STABLE, which appears to be an unrelated test failure. Would appreciate a review.

PR #1065, opened 2026-07-27
Add PostgreSQL 12.12 compatibility

smartyue076:


vyruss:
Your own LLM's comments state:

>  This fork contains a compatibility layer for PostgreSQL 12.12. The upstream project targets PostgreSQL 14 and newer, so the PostgreSQL 12.12 build uses a separate extension version, `1.0.0-pg12`, and intentionally disables the parts that depend on PostgreSQL APIs introduced after PostgreSQL 12.


 You must understand that this is not upstream mergeable (or desirable, since PG12 went EOL in 2024).

PR #1082, opened 2026-09-23
Use pg catalog schema

AndrewJackson2020:
This PR updates the control file with `schema = pg_catalog`. This option is documented here [0]. Per the docs

> This parameter can only be set for non-relocatable extensions. It forces the extension to be loaded into exactly the named schema and not any other. The schema parameter is consulted only when initially creating an extension, not during extension updates. See [Section 36.17.2](https://www.postgresql.org/docs/current/extend-extensions.html#EXTEND-EXTENSIONS-RELOCATION) for more information.

In the postgres source code it seems like all non relocatable extension except `xml2` are set up this way. Also `pg_cron` which is also non relocatable is set up in this way [1].

The advantages here are that the default schema  during `create extension` is public. Many platforms will flag having objects in public schema as a security warning, supabase for example and one internal platform that I have worked on.

This should also not have any backwards compatibility issues as the change only applies to new installs.

```
> rg --after-context=1 --glob='*.control' 'relocatable = false'
contrib/xml2/xml2.control
6:relocatable = false

src/pl/tcl/pltcl.control
5:relocatable = false
6-schema = pg_catalog

src/pl/tcl/pltclu.control
5:relocatable = false
6-schema = pg_catalog

src/pl/plperl/plperl.control
5:relocatable = false
6-schema = pg_catalog

src/pl/plperl/plperlu.control
5:relocatable = false
6-schema = pg_catalog

tmp_install_aj/share/extension/plpgsql.control
5:relocatable = false
6-schema = pg_catalog

src/pl/plpgsql/src/plpgsql.control
5:relocatable = false
6-schema = pg_catalog

src/pl/plpython/plpython3u.control
5:relocatable = false
6-schema = pg_catalog
```

[0]: https://www.postgresql.org/docs/current/extend-extensions.html#EXTEND-EXTENSIONS-FILES-SCHEMA
[1]: https://github.com/citusdata/pg_cron/blob/main/pg_cron.control


AndrewJackson2020:
Nevermind, misunderstood how pg_duckdb works. It looks like stuff gets installed in the extension schema as well as the duckdb schema. Also per [0] looks like the end goal is to make the extension relocatable which I think is better long term.

[0] https://github.com/duckdb/pg_duckdb/commit/38db514f6191cca307f3a5985caee80105efc130

